PR #10, opened 2017-04-11
Add and theme components according to styleguide

LucianBuzzo:
- Update Button to match styleguide
 - Add emphasized attribute to buttons
 - Update input fields to match styleguide
 - Update Buttons
 - Add textarea input
 - Add Alert component
 - Add Warning component
 - Buttons can be links
 - Added ability to add arbitrary attributes to components


.storybook/head.html
```diff
@@ -0,0 +1,3 @@
+<!-- Load Proxima nova font -->
+<script src="https://use.typekit.net/eic0uah.js"></script>
```
craig-mulligan: You're missing `<U+2028>` ;)

src/components/Alert/index.js
```diff
@@ -0,0 +1,114 @@
...
+    );
+
+    const alert = (
+      <div
+        {...attributes}
+        role="alert"
+        id={id}
+        className={classes}
+      >
+        { onDismiss || dismissable ?
```
craig-mulligan: You can simplify this to onDismiss || dismissable && <button..

https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator

Saves having to do the `: null`

src/components/Button/index.js
```diff
@@ -8,24 +8,31 @@ const Button = ({
...
 } = {}) => {
-  let classes = classNames('resin-control', className);
+  let classes = classNames('resin-control', 'resin-control-btn', className, { emphasized });
 
   if (btnStyle) {
-    classes += ` resin-control_btn-${btnStyle}`;
+    classes += ` resin-control-btn-${btnStyle}`;
   }
 
+  const Tag = href ? 'a' : 'button';
```
craig-mulligan: What benefits do native buttons actually give us?

craig-mulligan: This reminds me we should have a Link primitive. That basically checks if a link is for an external site or an internal one and then does html5 push state if it's internal and regular <a> if not. We can chat on it tomorrow.

src/components/Alert/index.js
```diff
@@ -0,0 +1,114 @@
...
+      dismissable,
+      emphasized,
+      id,
+      onDismiss,
+      transitionAppearTimeout,
+      transitionEnterTimeout,
+      transitionLeaveTimeout,
+      type,
+      title,
+      ...attributes
```
craig-mulligan: Can we just call this props? I think it makes it clearer that it's all the extra props.

src/components/Button/index.js
```diff
@@ -34,13 +41,15 @@ Button.propTypes = {
   children: React.PropTypes.node,
   className: PropTypes.string,
   disabled: PropTypes.bool,
+  emphasized: PropTypes.bool,
```
craig-mulligan: one thing I realised with `emphasized` is that we only have two options because it's boolean. Though imagine the style guide may want small medium big. It's not a big deal currently but worth checking with taahir.

src/styles/variables.less
```diff
@@ -20,7 +28,18 @@
 @resin-secondary-dark: darken(@resin-secondary, 10%);
 
 @resin-success: #76c950;
+@resin-success-light: #f9fcf7;
```
craig-mulligan: These don't work with a lighten(), darken() 😿 ?

.storybook/head.html
LucianBuzzo: :joy:

src/components/Button/index.js
LucianBuzzo: Nice idea! 
WRT buttons, they will submit forms.

src/styles/variables.less
LucianBuzzo: I struggled with this, but I think I will take another look today.

craig-mulligan:
Hey @LucianBuzzo Are you waiting on me for review, or do you still need to add things?

LucianBuzzo:
@craig-mulligan If you could re-review this it would be great. I'm not going to add more components in this PR.

.scripts/postinstall.js
```diff
@@ -7,7 +7,9 @@
 const fs = require('fs');
 const path = require('path');
 
-if (process.env.NODE_ENV === 'dev') {
+// Only run this script if the module is being installed into
+// a "node_modules" folder.
+if (process.cwd().indexOf('node_modules') === -1) {
```
craig-mulligan: We can remove this in favour of the es6 module approach + tree shaking. We can then have a single entry point in root that exports the components. (I'll add it in a PR soon)

src/components/Button/index.js
craig-mulligan: We can leave this for later but we definitely need a good approach to dealing with links. I've seen some libraries allow let you pass a component that acts as a wrapper around the link. But I'm not sure what the best approach is here. 

Let's bring it up in the FR call?

Issue #13, opened 2017-05-02
Have the components library use Bootstrap 4 and reactstrap as a base


LucianBuzzo:
https://github.com/reactstrap/reactstrap/
The idea is to use reactstrap as a base for our custom components.
The components library will include the BS4 library as well as the components
We'll also need to convert existing Less files to Sass.

PR #14, opened 2017-05-02
Add bootstrap 4 + reactstrap primitives

craig-mulligan:
Connects to #13

This is pretty massive (sorry). Have all basic bootstrap functionality, but now I'm trying to map our style guide to bootstrap variables. 

It may make sense to change the style guide slightly (interms of naming) to suite our needs. eg. Primary color in bootstrap is used for default link color so we should probably swap that with the secondary. There are some other considerations too, worth jumping on a call for that if you have opinions too. 

Change-Type: patch


craig-mulligan:
ping @LucianBuzzo. 

I'm waiting for the exact measurements to do the styling etc. But this adds a good base we can perfect the styling later.

craig-mulligan:
PingyPong @LucianBuzzo when you have a sec.

src/styles/variables.scss
```diff
@@ -0,0 +1,42 @@
...
+
+// BOOTSTRAP SEMANTICS
+$brand-primary: $blue;
+
+// CUSTOM SEMANTICS
+$brand-secondary: $yellow;
+$brand-tertiary: $gray-dark;
+
+// FONTS
+$font-family-sans-serif: "proxima-nova", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial;
```
LucianBuzzo: After recent discussion, we're going to be using Roboto (same as the dashboard) so this should probably be changed.

craig-mulligan:
Updated @LucianBuzzo

PR #16, opened 2017-06-05
Clean up scripts

craig-mulligan:
removed old scripts + add gh-pages to handle deploys

Change-Type: patch


craig-mulligan:
Oh I find gh-pages so useful, what issues did you hit @thgreasi ?

thgreasi:
I had trouble in cases where:
* the `gh-pages` branch was not already there
* the `.publish` directory was in bad state and had to manually delete it
* after a bad un/pass authorization + some caching, I had to logout to make it work again

Anyway it might be just me in my crazy side-project :)

PR #23, opened 2017-09-04
Recompose

craig-mulligan:
@LucianBuzzo 

Sorry this is a little big, I did most of it while on the plane yesterday so couldn't push, I've done a few things. 

1. update storybook to `v3`
2. removed rebass and just used underlying modules, rebass wraps up a few core libraries, I found it more flexible to just use those directly and instead of hacking around the rebass.
3. Tried to make the `styled.<component>` as simple as possible and apply customisations with higher-order components (see the recompose stuff) - this works quite nice only weird thing atm, is that our default botton is outlined so when you apply the `outline` hoc, it switches the props around so the `color` & `bg` props api is in reverse, which is weird. May need to look at it with fresh eyes.


src/theme.js
```diff
@@ -79,3 +64,53 @@ export const colors = {
...
+]
+
+export const weights = [
+  400,
+  700
+]
+
+export const radius = 4
+export const font = `-apple-system, BlinkMacSystemFont, sans-serif`
+export const monospace = '"SF Mono", "Roboto Mono", Menlo, monospace'
```
LucianBuzzo: font should be `Roboto, Arial, sans-serif` and monospace should be `'Ubuntu Mono Web', 'Courier New', monospace`

LucianBuzzo:
LGTM apart from the font

PR #29, opened 2017-09-04
Add Device Status Gauge component

LucianBuzzo:
Connects to #22

change-type: minor


src/components/DeviceStatusGauge.js
```diff
@@ -0,0 +1,241 @@
...
+    center: [0, 0],
+    data: data,
+    R: 46,
+    r: 25
+  })
+
+  const gauge = graph.curves
+  const totalDevices = allDevices.length
+  return (
+    <DeviceStatusGaugeWrapper className='device-status-gauge' {...props}>
```
craig-mulligan: What are the classNames used for?

src/components/DeviceStatusGauge.js
```diff
@@ -0,0 +1,241 @@
...
+import _ from 'lodash'
+import Pie from 'paths-js/pie'
+import styled, { withTheme } from 'styled-components'
+import { compose } from 'recompose'
+import ResinDeviceStatus from 'resin-device-status'
+
+import hoc from '../hoc'
+import theme from '../theme'
+
+const { colors, font } = theme
```
craig-mulligan: We should be getting the theme from `.props`.

src/components/DeviceStatusGauge.js
```diff
@@ -0,0 +1,241 @@
...
+
+  if (status === 'offline') {
+    return colors.statusOffline.main
+  }
+}
+
+const DeviceStatusGauge = ({ devices, ...props }) => {
+  const data = []
+  const allDevices = devices || []
+  const statuses = ResinDeviceStatus.statuses
```
craig-mulligan: I think we are to tightly coupling components to business logic here. I think it'd be better to think of this as vanilla gauge.

You pass in a data type like array of status and renders a graph. The reason I say this is that there are likely cases where you will need to customize the data the component consumes. You may use this gauge on the admin dashboard, where you want all devices and not just a single users devices, IMO it's better to keep the component lib as dumb and agnostic as possible data in -> component out.

src/components/DeviceStatusGauge.js
```diff
@@ -0,0 +1,241 @@
...
+
+  .device-status-gauge__graph {
+    display: block;
+    width: 100%;
+    max-width: 180px;
+    margin: 0 auto 10px;
+  }
+
+  .device-status-gauge__graph__total-devices-label {
+    font-family: ${font};
```
craig-mulligan: Why use `inherit`?

src/components/DeviceStatusGauge.js
```diff
@@ -0,0 +1,241 @@
...
+    font-size: 5px;
+    fill: #b8b8b8;
+    text-anchor: middle;
+  }
+
+  .device-status-gauge__graph__total-devices-number {
+    font-family: ${font};
+    font-size: 26px;
+    fill: #5e5e5e;
```
craig-mulligan: Should these colors should be derived from our theme?

src/components/DeviceStatusGauge.js
```diff
@@ -0,0 +1,241 @@
...
+      s => s.key === ResinDeviceStatus.status.OFFLINE
+    ),
+    'name',
+    'Offline'
+  ),
+  value: 1
+})
+
+const pathFillColor = status => {
+  if (status === 'idle') {
```
craig-mulligan: this could be `return colors[`status${camelCase(status)}`].main`

craig-mulligan: But in IMO this should just be called warning, info, ... then etcher and other libs could use the gauge with out it being device specific.

src/components/DeviceStatusGauge.js
LucianBuzzo: Do you mean *why not ?

src/components/DeviceStatusGauge.js
LucianBuzzo: Yep that would be better

src/components/DeviceStatusGauge.js
LucianBuzzo: agreed

src/components/DeviceStatusGauge.js
LucianBuzzo: I think we should probably offer both, a generic gauge and a device gauge that wraps it.

PR #37, opened 2017-09-07
Update README and add script for publishing to GH pages

LucianBuzzo:
Connects to #31

change-type: patch


README.md
```diff
@@ -40,22 +45,26 @@ npm install
...
+The interactive storybook can be published by running:
+
+```sh
+npm run publish-storybook
 ```
 
-### Publishing
+## Testing
 
 ```
```
thgreasi: How about adding an `sh` here as well?

src/stories/PineTypes.js
```diff
@@ -1,81 +0,0 @@
-import { h, Component } from 'preact'
```
thgreasi: I wouldn't expect to see such a deletion in this PR.

.scripts/publish_storybook.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
thgreasi: Could we use something like [gh-pages](https://github.com/tschaub/gh-pages) for this?

src/stories/PineTypes.js
LucianBuzzo: Yeah it got commited by mistake I think

.scripts/publish_storybook.sh
LucianBuzzo: This looks good, I'm going to create a separate issue for addressing this in the future.

LucianBuzzo: https://github.com/resin-io-modules/resin-components/issues/39

Issue #41, opened 2017-09-11
Auto publish to npm on master merge


LucianBuzzo:
This should be a straightforward job on travis/circle

PR #52, opened 2017-09-14
Auto deploy package to npm

LucianBuzzo:
Connects to #41

change-type: patch


.travis.yml
```diff
@@ -1,12 +1,20 @@
 language: node_js
-node_js:
-  - "7"
-
+matrix:
```
craig-mulligan: Why did you add a matrix with one build?

LucianBuzzo: I was copying the yml configuration from the CLI https://github.com/resin-io/resin-cli/blob/master/.travis.yml
I think if we drop the matrix value we can get rid of the "can deploy" env var as well.

LucianBuzzo: I'm tempted to keep this here in case we want to test against multiple node versions in the future

PR #56, opened 2017-09-14
Add dist file to repo and auto build on commit

LucianBuzzo:
Connects to #55

change-type: patch


craig-mulligan:
Why checkin in `dist`?

LucianBuzzo:
I'd like to make our build process transparent, and we now have the tooling to ensure the artefacts are built correctly. I think we can easily exclude the stories directory from the dist output.

craig-mulligan:
Hmm, I disagree quite strongly especially now that we have pre-commit hook, it'll make pr's twice as noisy with little benefit, but let's see how it feels and remove later if we like.

LucianBuzzo:
@craig-mulligan I'm happy to discuss it further. We could use the travis values `skip_cleanup` and `after_success` to build and publish the files on Travis.
Our `after_success`  entry would look something like this
```yml
after_success:
  if ([ "$TRAVIS_BRANCH" == "master" ] || [ ! -z "$TRAVIS_TAG" ]) && 
      [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
    npm run build
  fi
```

craig-mulligan:
@LucianBuzzo, Yep that'd be rad :) I'd put that script in the main travis `script` section because if the npm build fails it'll still run the other `after_success` scripts.

Issue #69, opened 2017-09-25
Gauge renders the placeholder color when only one item has value


thgreasi:
See: https://github.com/andreaferretti/paths-js/issues/47

thgreasi:
Resolved by #70

PR #70, opened 2017-09-25
Fix gauge color when only one item has value

thgreasi:
Connects-To: #69
Change-Type: patch
See: https://github.com/andreaferretti/paths-js/issues/47
---- Autogenerated Waffleboard Connection: Connects to #69


src/stories/Gauge.js
```diff
@@ -20,6 +20,28 @@ const data = [
...
+    name: 'Strawberry',
+    color: 'pink'
+  }
+]
+
+storiesOf('Gauge', module)
+  .addWithInfo('default', () => {
+    return <Gauge m={30} title='Ice Cream' data={data} />
+  })
+  .addWithInfo('single item', () => {
```
LucianBuzzo: Can we change this to `single segment`? 
`single item` makes it seem like there is only 1 item in the array.

PR #76, opened 2017-09-26
Add SchemaSieve support for arrays of rules in filter() method

LucianBuzzo:
Connects to #75

change-type: minor


test/filter.spec.js
```diff
@@ -776,6 +776,23 @@ describe('resin-filter', () => {
         const result = sieve.filter(this.collection, input)
         expect(result[0].test).to.equal(this.collection[0].test)
       })
+
+      it.only('should allow an array of rules', function () {
```
pimterry: Don't commit `.only`!

src/components/Filters/SchemaSieve.js
```diff
@@ -53,11 +53,23 @@ class SchemaSieve {
...
 
   filterObject (collection, input) {
-    return _.pickBy(collection, value => this.baseTest(value, input))
+    if (_.isArray(input)) {
+      return _.pickBy(collection, value =>
+        _.every(input, i => this.baseTest(value, i))
+      )
+    } else {
+      return _.pickBy(collection, value => this.baseTest(value, input))
+    }
```
pimterry: You could more simply refactor the common structure here (if array, `_.every(x => test(x))`, otherwise `test(x)`)  into `baseTest`, or into a new function which wraps `baseTest` (`test` and `testSingleInput`?).

That'll be much simpler, and means this code wouldn't have to change, and you wouldn't end up duplicating the same logic and structure you've used here.

test/filter.spec.js
LucianBuzzo: All the cool kids do it 😎

src/components/Filters/SchemaSieve.js
LucianBuzzo: Very good point, looking at the code I think it would be best to wrap `baseTest` in a function `test` that will perform the `isArray` check.

Issue #79, opened 2017-09-26
Add support for human friendly labels when using filters/schema


LucianBuzzo:
Currently, the filterable field uses the field name on the display.
If my model looks like this:
```js
const schema = {
  is_supported: {
    type: 'Boolean'
  }
};
```

Then the output looks like this:

![screen shot 2017-09-26 at 15 21 45](https://user-images.githubusercontent.com/15064535/30865413-6f60d01e-a2ce-11e7-9164-e8f7324e1a5b.png)

Ideally I want to provide human friendly labels to my schema.

PR #81, opened 2017-09-26
Add support for human friendly labels when using filters

LucianBuzzo:
Connects to #79

change-type: minor


LucianBuzzo:
@thgreasi Yeah absolutely, I'll update this PR as it should be bundled together ideally

PR #89, opened 2017-09-27
Add drop down button

craig-mulligan:
Connects to #88 

We should probably have a chat about how we handle multi-component components, in most cases so far we wrap it into a single component and allow a few controls threw the level component props.

But many other libs split this out into multiple components and then make the user handle the state. Both have merits but we should chat about it, maybe in the next frontend call?


src/components/DropDownButton.js
```diff
@@ -0,0 +1,97 @@
...
+
+const ButtonBase = styled(Button)`
+  border-top-right-radius: 0;
+  border-bottom-right-radius: 0;
+  border-right: 0;
+  margin: 0;
+  border-width: ${props => props.outline && '1px'};
+`
+
+const MenuBase = styled.div`
```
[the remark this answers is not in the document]
craig-mulligan: Yea very good point, we don't want layout changing on dropdown open.

src/components/DropDownButton.js
```diff
@@ -0,0 +1,104 @@
...
+      <Wrapper {...props}>
+        <ButtonBase {...props}>{label}</ButtonBase>
+        <Toggle
+          {...props}
+          handler={e => this.toggle(e)}
+          open={this.state.open}
+        />
+        {this.state.open && (
+          <MenuBase
+            onClick={e => this.toggle(e)}
```
thgreasi: I guess that in case that we need to want the dropdown open after clicking on an item, we could `preventDefault` or something as an item onClick?

craig-mulligan: Hmm yea, currently the listener is attached to the MenuBase container, which the user has no control over, so the menuBox would always currently always toggle visibility. We can definitely add that if needed though.

thgreasi: We could have an "auto-close" property, but I'm fine with adding this at a later point of time.

Issue #59, opened 2017-09-18
Add underline buttons


LucianBuzzo:
This is currently used for the 'View Pricing' button on the billing page

thgreasi:
This was fixed by #59 but `connects to` was used so the issue didn't auto-close.

PR #90, opened 2017-09-28
Refactor Button + add underline prop

craig-mulligan:
I was reading the styled-components docs, and they suggest using StyledComponent.extend`ccs` rather than styled(Component)`css` to extend styles. It has performance improvements and only renders the differing styles rather than creating new classes for each type of component.

Connects to #59
Change-Type: patch


src/components/Button.js
```diff
@@ -44,79 +71,39 @@ const Base = styled.button`
...
-    color: '#fff',
-    bg: get(theme.colors[type], 'main'),
-    active: {
-      color: '#fff',
-      bg: get(theme.colors[type], 'dark')
-    }
+  &:hover,
+  &:focus,
+  &:active {
+    background: none;
```
thgreasi: Just a note, in the billing page of the dashboard the underline buttons get filled on hover/focus/active, looking similar with the hovered outlined buttons. Never had a spec on this and neither have a strong opinion though.

src/components/Button.js
```diff
@@ -44,79 +71,39 @@ const Base = styled.button`
...
+    color: ${props =>
+    getColor(props, 'color', 'main') || props.theme.colors.text.main};
+    border-bottom: 2px solid;
```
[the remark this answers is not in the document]
craig-mulligan: Nice 👍

src/components/Button.js
craig-mulligan: Hmm, yea I'm not sure either? I was basing it off the designs for the marketing site pricing page. Great that we are catching these inconsistencies :) @konmouz any opinions? If not we can create an issue and change the button when we have a design for it.

konmouz: @craig-mulligan @thgreasi, at the moment @taahirisaacs is preparing a new styleguide version that will be shared soon. After that, we can review what elements are missing and populate the guide.

craig-mulligan: Okay thanks @konmouz, I'll create an issue so we can revisit it once we have designs.

craig-mulligan:
@thgreasi @LucianBuzzo could you have a sec. 👍

PR #100, opened 2017-10-02
Add Headings + refactor Text

craig-mulligan:
Connects to #97

Change-type: minor

Shamelessly borrowed from https://github.com/pricelinelabs/design-system. I've realized that the HoC is actually not a great pattern for styled components, for reasons I mentioned in the button refactor PR. I think we should instead pull in utilities from styled-system on a per component basis, it seems to be making things simpler.

### DIFF
--- a/src/components/Heading.js
+++ b/src/components/Heading.js
@@ -1,0 +1,48 @@
+import Text from './Text'
+
+const Heading = Text.withComponent('h3')
+
+Heading.displayName = 'Heading'
+
+Heading.defaultProps = {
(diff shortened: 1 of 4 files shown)


craig-mulligan:
Huh, Versionbot just merged without running the versions and adding the tags.

craig-mulligan:
[Manually fixed with aed5f5549](https://github.com/resin-io-modules/resin-components/commit/aed5f5549da8773f1f1191d3657fb80f41af5654)

PR #102, opened 2017-10-02
Add Image component

craig-mulligan:
Connects to #96

Change-Type: patch


LucianBuzzo:
@craig-mulligan Can you add a story for this component please?

PR #106, opened 2017-10-02
Default Date/Time related filters to the current time

LucianBuzzo:
Connects to #105

change-type: patch


src/components/Filters/index.js
```diff
@@ -187,10 +188,19 @@ class Filters extends Component {
...
-      update.label = inputModels[value].label
-      update.value = ''
+      update.operator = model.availableOperators[0]
+      update.label = model.label
+      if (model.type === 'Date Time') {
+        update.value = moment().format('YYYY-MM-DDTHH:mm')
+      } else if (model.type === 'Date') {
+        update.value = moment().format('YYYY-MM-DD')
+      } else if (model.type === 'Time') {
+        update.value = moment().format('HH:mm')
```
thgreasi: How about moving the format strings to properties with default values like you did for the respective pine types?

LucianBuzzo: Unfortunately, this doesn't work as the value sent to the component is defined in the `edit` object. If the `Display` component in the pine type has a default value, then the display will show a value, but when you hit save, the rule won't have that value in it

LucianBuzzo:
@thgreasi Only two of the pokemon have `first seen` values in their entries

PR #120, opened 2017-10-04
Change `storybook` npm script to `start`

LucianBuzzo:
change-type: patch


LucianBuzzo:
@thgreasi I think the package is young enough that there is no need to try to keep backwards compatibility in the npm scripts. I'm adverse to having duplicate commands or aliases in `package.json` anyway.

thgreasi:
:+1:

PR #122, opened 2017-10-04
Add Link component

craig-mulligan:
Connects to #121

Change-Type: minor


craig-mulligan:
Is there a way to have the button components work like links (I mean an easier way than doing that inside the onClick)?

Could you explain a bit more? Do you mean passing a url to a button and making it act like an Link in that case?

thgreasi:
> Do you mean passing a url to a button and making it act like an Link in that case?

Exactly.

craig-mulligan:
In that case you can just compose them like this: 

```
<Link to='/'>
   <Button mx={2} primary >
        Button
   </Button>
</Link>
```

Though Link also allows you to swap out the underlying component like:
```

<Link is={Button} primary mx={2} />
```
Though I'm not sure the `is` prop provide the correct styling to the Button component it is mainly an escape hatch so you can use a `react-router` `Route` component instead of a native `a` element.

PR #125, opened 2017-10-05
State.open is toggled on navbar children click

craig-mulligan:
Connects to #124
Also changed the Link prop for href from `to` to `href`

Change-type: patch


src/stories/Typography.js
```diff
@@ -40,10 +40,10 @@ storiesOf('Typography', module)
     return (
       <div>
         <div>
-          <Link to={`#`}>Internal Link</Link>
+          <Link href={`#`}>Internal Link</Link>
```
thgreasi: Should this also be changed in the Navbar story as well?

craig-mulligan: Yes definitely, thanks 🙏

PR #147, opened 2017-10-11
Add Enum type to filters

### LINKED ISSUE #146: Add Enum type to filters
Connects to https://github.com/resin-io/resin-ui/issues/1374

LucianBuzzo:
Fixes #146

change-type: patch


src/components/Filters/index.js
```diff
@@ -56,14 +56,18 @@ const SimpleSearchBox = styled.div`
   }
 `
 
-const FilterWrapper = styled(Box)`
-  position: relative;
-`
+const FilterWrapper = styled(Box)`position: relative;`
```
thgreasi: I like it when prettier just hasn't yet decided what to use and randomly changes stuff between versions.
It's like its taste evolves with time or maybe it's a kind of a joke messing with us on reviews.

src/components/PineTypes/Enum.js
```diff
@@ -0,0 +1,28 @@
+import { h } from 'preact'
+import Select from '../Select'
+import _ from 'lodash'
+
+export const Edit = ({ schema, value, onChange, ...props }) => (
+  <Select {...props} value={value} onChange={e => onChange(e.target.value)}>
+    {_.map(schema.values, item => {
+      if (_.isString(item)) {
+        return <option>{item}</option>
```
thgreasi: I guess that we don't need something like `value={item}`, do we?

src/components/PineTypes/Enum.js
```diff
@@ -0,0 +1,28 @@
...
+export const rules = {
+  is: (target, value) => target === value,
+  'is not': (target, value) => target !== value
+}
+
+export const validate = _.isBoolean
+
+export const Display = ({ data, ...props }) => (
+  <div {...props}>
+    <span>{data ? 'true' : 'false'}</span>
```
thgreasi: Is that correct?

src/stories/Filters.js
```diff
@@ -72,6 +74,8 @@ class FiltersDemo extends Component {
       ],
       schema
     }
+
+    window.comp = this
```
thgreasi: Is that a debug leftover?

src/components/PineTypes/Enum.js
LucianBuzzo: `<option>` tags default to using their inner text as the value. If you need to provide a different label to the value then you can pass in an object here.

src/components/PineTypes/Enum.js
LucianBuzzo: Nope, this needs to change :smile:

Issue #145, opened 2017-10-10
bundle size


hobochild:
I was building a site with resin components and see the bundle size is massive, and around half of if is lodash + react-icons. 

I think this is because we are importing lodash like `import _ from 'lodash'` and `import { icon } from 'react-icons'` we then transpile it to commonjs and webpack only support's tree shaking with es modules `export/import syntax`

@LucianBuzzo have you see similar things with the dashboard bundle? 

The one way to definitely it is do explicit imports like `import filter from 'lodash/filter'`, but open to other ideas?

LucianBuzzo:
@craig-mulligan I think we could probably do a lot just by using explicit imports and taking advantage of tree shaking. I've been hesitant of jumping into bundle size optimization whilst the project is still maturing (though it's always been on my mind).

PR #150, opened 2017-10-11
Improve bundle size with explicit imports

craig-mulligan:
This halved the bundle size on the etcher website.
Connects to #145

Change-Type: patch


craig-mulligan:
It went from 370kb to 200kb (thats minified + gzipped) :)

craig-mulligan:
Showdown and moment still account for around 15% of the bundle, we should figure out how to optionally include those, but that can come later when we optimize things.

PR #156, opened 2017-10-16
Switch to React instead of Preact

LucianBuzzo:
Now that React has an MIT license, switching to it will ease development and
third party integrations, as well as making a transition to TypeScript easier.

Connects to #155

change-type: major


thgreasi:
So now get our semicolons back but also use double quotes? :)

src/components/CodeWithCopy.js
```diff
@@ -31,15 +31,13 @@ const Wrapper = styled.span`
 const Base = ({ copy, text, ...props }) => {
   const getCopyText = () => (copy || text || '').trim()
   return (
-    <Wrapper class='code-with-copy'>
-      <code ng-copy='showTooltip(true)' title={copy}>
```
thgreasi: I guess that `ng-copy='showTooltip(true)'` was a left-over.

src/components/Filters/index.js
```diff
@@ -304,7 +308,7 @@ class Filters extends Component {
                   >
                     {map(
                       inputModels[this.state.edit.name].availableOperators,
-                      name => <option>{name}</option>
+                      name => <option key={name}>{name}</option>
```
thgreasi: I think that this should also be added to `PineTypes/Enum.js`.

LucianBuzzo: Well spotted! 🐄

src/stories/Filters.js
```diff
@@ -7,7 +7,9 @@ import PokeDex from './assets/pokedex'
 
 const sieve = SchemaSieve()
 
-const Container = styled.div`margin: 30px;`
+const Container = styled.div`
+  margin: 30px;
+`
 
 const StyledTable = styled.table`
```
craig-mulligan: We should add an issue to add a table component as well.

LucianBuzzo: Agreed! https://github.com/resin-io-modules/resin-components/issues/157

PR #158, opened 2017-10-19
Add TypeScript support and begin conversion to TypeScript

LucianBuzzo:
Connects to #44

change-type: patch


.storybook/config.js
```diff
@@ -1,29 +1,29 @@
-import { configure, setAddon, addDecorator } from "@storybook/react";
-import infoAddon from "@storybook/addon-info";
-import React from "react";
-import { Provider } from "../src/index";
-import theme from "../src/theme";
-import { injectGlobal } from "styled-components";
+import { configure, setAddon, addDecorator } from '@storybook/react'
+import infoAddon from '@storybook/addon-info'
+import * as React from 'react'
```
craig-mulligan: Can't we import .default with type script?

craig-mulligan: I don't mind the syntax either way really, just find it weird that you can't do the same as you can do in es6? I see there is a `--allowSyntheticDefaultImports` would that allow us to do `import React from 'react'`?

LucianBuzzo: Babel does some magic where it uses an interOp to normalize these kinds of imports. The typescript compiler doesn't do this AFAIK

.editorconfig
```diff
@@ -9,5 +9,19 @@ charset = utf-8
...
 insert_final_newline = true
 
+[*.ts]
+indent_style = tab
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.tsx]
```
pimterry: You can collapse these two into `[*.{ts,tsx}]`

.editorconfig
```diff
@@ -9,5 +9,19 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[*.ts]
+indent_style = tab
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
pimterry: All but `indent_style` here is the same as for `*` - I think that means you can just omit them entirely.

package.json
```diff
@@ -11,9 +11,9 @@
   "scripts": {
     "start": "start-storybook -p 6006",
     "watch": "babel src -w -d dist --ignore src/stories",
```
pimterry: Does this `watch` not also need updating?

package.json
```diff
@@ -11,9 +11,9 @@
...
-    "build": "babel src -d dist --ignore src/stories",
+    "build": "rimraf dist && tsc",
     "build:storybook": "build-storybook -o .out",
-    "test": "npm run lint && mocha --compilers js:babel-core/register test/*.spec.js",
+    "test": "npm run lint && mocha --require ts-node/register test/*.spec.js",
```
pimterry: If we're no longer using `babel` at all, can we drop it from the dev dependencies?

src/components/PineTypes/Enum.js
```diff
@@ -23,7 +24,7 @@ export const rules = {
   'is not': (target, value) => target !== value
 }
 
-export const validate = _.isBoolean
+export const validate = value => !!value
```
pimterry: These aren't the same! Is that intentional?

package.json
```diff
@@ -11,9 +11,9 @@
   "scripts": {
```
pimterry: Seems like this would also be an excellent time to pull the `.d.ts` from the UI into here, until this is fully converted, and add a `types` here, so we can start adding and changing types here and immediately pull the result through elsewhere.

src/utils.js
```diff
@@ -1,5 +1,5 @@
-import Color from 'color'
-import get from 'lodash/get'
+const Color = require('color')
+const get = require('lodash/get')
```
thgreasi: I guess that these can also be
```
import * as Color from 'color'
import * as get from 'lodash/get'
```

.storybook/webpack.config.js
```diff
@@ -1,21 +1,21 @@
 module.exports = {
-  name: "client",
-  target: "web",
+  name: 'client',
+  target: 'web',
   resolve: {
-    extensions: [".js", "jsx"]
+    extensions: ['.ts', '.tsx', '.js', 'jsx']
```
thgreasi: `jsx` is the only one missing a dot here. Is that correct? Nevertheless I would love to make it cnsistent.

package.json
LucianBuzzo: This watch can be removed entirely, as development happens using storybook.

package.json
LucianBuzzo: We need to keep our presets, as they are used by storybook, but I think everything else can be removed.

src/components/PineTypes/Enum.js
LucianBuzzo: I snuck in a little hotfix here 😉 The `isBoolean` implementation is completely wrong.

typings.d.ts
```diff
@@ -0,0 +1,139 @@
+declare module 'resin-components' {
+	import { Component } from 'react';
+	import * as ResinSdk from 'resin-sdk';
```
thgreasi: Is this a 'single space' indentation?

package.json
```diff
@@ -8,12 +8,12 @@
...
     "start": "start-storybook -p 6006",
-    "watch": "babel src -w -d dist --ignore src/stories",
-    "build": "babel src -d dist --ignore src/stories",
+    "build": "rimraf dist && tsc",
     "build:storybook": "build-storybook -o .out",
-    "test": "npm run lint && mocha --compilers js:babel-core/register test/*.spec.js",
+    "test": "npm run lint && mocha --require ts-node/register test/*.spec.js",
     "lint": "eslint src test",
     "precommit": "lint-staged",
     "prettify": "prettier-eslint --write \"src/**/*.js\" \"test/**/*.js\"",
```
thgreasi: I guess that we should now also target `ts`, `tsx` & `jsx`?

thgreasi: And probably (later?) also add tslint along with the configuration from the UI?

LucianBuzzo: Yeah eventually that would be good to include tslint.

LucianBuzzo: This won't target ts or tsx as the eslint config is JS specific

typings.d.ts
LucianBuzzo: No, it uses tabs

package.json
thgreasi: We can add the repsective TS prettier command when we start converting to ts(x) files :+1:

typings.d.ts
thgreasi: I swear that Chrome rendered it like a single space yesterday... Nevermind & sorry for the noise.

thgreasi: I swear that Chrome rendered it like a single space yesterday... Nevermind & sorry for the noise.

PR #160, opened 2017-10-27
Prevent CodeWithCopy component from propagating click events

### LINKED ISSUE #159: Prevent CodeWithCopy component from propagating click events
If the CodeWithCopy component is nested in a clickable element, clicking the copy Icon will trigger the parents click events.

LucianBuzzo:
Fixes #159

change-type: patch


src/components/CodeWithCopy.js
```diff
@@ -28,18 +29,23 @@ const Wrapper = styled.span`
...
+const copyIt = (copy, text) => {
+  const copyText = (copy || text || '').trim()
+
+  copyToClipboard(copyText)
+}
+
+const Base = ({ copy, text, color, ...props }) => {
   return (
-    <Wrapper className='code-with-copy'>
+    <Wrapper onClick={stopEvent} {...props} className='code-with-copy'>
```
thgreasi: Would it make sense to move the `stopEvent()` a bit deeper, so that clicking the text still works like a click?

PR #164, opened 2017-10-30
Improve bundled typings

LucianBuzzo:
change-type: minor


typings.d.ts
```diff
@@ -72,11 +71,35 @@ declare module 'resin-components' {
...
+
+	interface Coloring {
+		primary?: boolean;
+		secondary?: boolean;
+		tertiary?: boolean;
+		quartenary?: boolean;
+	}
+
+	interface BannerProps extends DefaultProps {
+		backgorundImage?: string;
```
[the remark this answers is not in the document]
LucianBuzzo: https://www.shutterstock.com/search?search_source=base_landing_page&language=en&searchterm=backgorund&image_type=all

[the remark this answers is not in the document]
LucianBuzzo: Typo 😒

typings.d.ts
```diff
@@ -110,29 +146,101 @@ declare module 'resin-components' {
...
 
+	class ProgressBar extends RenderableElementWithProps<ProgressBarProps, any> {}
+
 	class SchemaSieveClass {
 		filter<T>(items: T[], rule: FilterRule): T[];
 	}
 
 	function SchemaSieve(): SchemaSieveClass;
+
+	class Select extends RenderableElementWithProps<DefaultProps, any> {}
```
thgreasi: Should this also have `emphasized` as part of an inherited interface?

typings.d.ts
```diff
@@ -72,11 +71,35 @@ declare module 'resin-components' {
...
-		refs: any;
-		render(): JSX.Element | null;
+	> {}
+
+	interface Coloring {
+		primary?: boolean;
+		secondary?: boolean;
+		tertiary?: boolean;
+		quartenary?: boolean;
+	}
```
thgreasi: Should we also have a `Sizing` interface with props like `emphasized`?

thgreasi: Should we also have danger, warning, success, info?

typings.d.ts
```diff
@@ -110,29 +146,101 @@ declare module 'resin-components' {
...
+	class Text extends RenderableElementWithProps<DefaultProps, any> {}
+
+	interface TextareaProps extends DefaultProps {
+		value?: string;
+		rows?: number;
+		placeholder: string;
+	}
+
+	class Textarea extends RenderableElementWithProps<TextareaProps, any> {
+		onKeyUp: () => any;
```
thgreasi: Do we have to redefine this?
Since it seems that DOMAttributes also contains onKeyUp (but not with `any`), which seems to be inherited by the following chain: DefaultProps extends React.HTMLAttributes extends DOMAttributes

typings.d.ts
```diff
@@ -110,29 +146,101 @@ declare module 'resin-components' {
...
 		filter<T>(items: T[], rule: FilterRule): T[];
 	}
 
 	function SchemaSieve(): SchemaSieveClass;
+
+	class Select extends RenderableElementWithProps<DefaultProps, any> {}
+
+	class Text extends RenderableElementWithProps<DefaultProps, any> {}
+
+	interface TextareaProps extends DefaultProps {
```
thgreasi: Would it be possible to also extend `React.TextareaHTMLAttributes` to avoid redefining the following properties? (Even though `value is types as `string | string[] | number` in there.)

typings.d.ts
```diff
@@ -110,29 +146,101 @@ declare module 'resin-components' {
...
-	class Provider extends Component<any, any> {
-		refs: any;
-		render(): JSX.Element | null;
+	var PineTypes: { [K in PineDataType]: PineTypeModule }
+
+	class Provider extends Component<any, any> {}
+
+	interface ProgressBarProps extends DefaultProps {
+		value: number;
+		info?: boolean;
```
thgreasi: Would it make sense to have this along with `danger`, `warning` & `success` in a different interface?

typings.d.ts
```diff
@@ -110,29 +146,101 @@ declare module 'resin-components' {
...
+	class Navbar extends RenderableElementWithProps<
+		DefaultProps & {
+			brand: JSX.Element;
+		},
+		any
+	> {}
+
+	interface PineTypeModule {
+		rules: {
+			[key: string]: (target: any, value: any) => boolean
```
thgreasi: The Boolean type has an one argument method for this. Should this be reflected in here?

typings.d.ts
```diff
@@ -110,29 +146,101 @@ declare module 'resin-components' {
...
+	class Input extends RenderableElementWithProps<DefaultProps, any> {}
+
+	class Image extends RenderableElementWithProps<
+		DefaultProps & { src: string },
+		any
+	> {}
+
+	class Modal extends RenderableElementWithProps<
+		DefaultProps & {
+			title: string;
```
thgreasi: Heads-up, this is also part of the `HTMLAttributes`, but we probably don't have to do something for that.

typings.d.ts
```diff
@@ -110,29 +146,101 @@ declare module 'resin-components' {
...
 
-	class ProgressBar extends RenderableElementWithProps<ProgressBarProps, any> {
-		refs: any;
-		render(): JSX.Element | null;
+	class Link extends RenderableElementWithProps<LinkProps, any> {}
+
+	class Input extends RenderableElementWithProps<DefaultProps, any> {}
+
+	class Image extends RenderableElementWithProps<
+		DefaultProps & { src: string },
```
thgreasi: Would it make sense to extend `ImgHTMLAttributes` instead of manually defining `src`?

typings.d.ts
```diff
@@ -72,11 +71,35 @@ declare module 'resin-components' {
...
 	}
 
+	class Box extends RenderableElementWithProps<BoxProps, any> {}
+
+	interface ButtonProps extends DefaultProps, Coloring {
+		square?: boolean;
+		disabled?: boolean;
+	}
+
+	class Button extends RenderableElementWithProps<ButtonProps, any> {}
```
thgreasi: The story for this element uses the `onPress` property that I wasn't able to find anywhere (neither in the src nor in any typings). Perhaps the story is wrong.

typings.d.ts
```diff
@@ -72,11 +71,35 @@ declare module 'resin-components' {
...
+	interface BoxProps extends DefaultProps {
+		flex?: string | string[];
+		order?: number | string | Array<number | string>;
 	}
 
+	class Box extends RenderableElementWithProps<BoxProps, any> {}
+
+	interface ButtonProps extends DefaultProps, Coloring {
+		square?: boolean;
+		disabled?: boolean;
```
thgreasi: I think that this is missing the `emphasized`, `outline` & `underline` properties.

PR #166, opened 2017-10-30
Don't show an input field when filtering on a boolean value

### LINKED ISSUE #80: Don't show an input field when filtering on a boolean value
At the moment if you have a boolean value in your schema you will see an unnecessary input field for boolean values.

screen shot 2017-09-26 at 15 36 15

LucianBuzzo:
Fixes #80

change-type: patch


src/components/Filters/index.js
```diff
@@ -311,12 +311,14 @@ class Filters extends React.Component {
                       name => <option key={name}>{name}</option>
                     )}
                   </Select>
-                  <FilterInput
-                    schema={this.props.schema[this.state.edit.name]}
-                    value={this.state.edit.value}
-                    onChange={value => this.handleEditChange(value, 'value')}
-                    type={inputModels[this.state.edit.name].type}
-                  />
+                  {inputModels[this.state.edit.name].type !== 'Boolean' && (
```
thgreasi: I guess that having an `shouldShowInput()` method might be an overkill.

LucianBuzzo: AFAIK the boolean value is the only instance of this behaviour

PR #169, opened 2017-11-02
Improve typings to use responsive styled-system props

LucianBuzzo:
change-type: patch


typings.d.ts
```diff
@@ -179,11 +205,17 @@ declare module 'resin-components' {
 		class h4 extends RenderableElementWithProps<FlexProps, any> {}
 	}
 
-	interface LinkProps
-		extends DefaultProps,
-			React.AnchorHTMLAttributes<HTMLElement> {
+	interface LinkProps extends DefaultProps {
 		blank?: boolean;
 		disabled?: boolean;
+		download?: any;
```
pimterry: I think this should be `string`, shouldn't it?

typings.d.ts
```diff
@@ -245,7 +277,25 @@ declare module 'resin-components' {
 
 	class Text extends RenderableElementWithProps<DefaultProps, any> {}
 
-	interface TextareaProps extends StyledSystemProps, React.TextareaHTMLAttributes<HTMLTextAreaElement> {}
+	interface TextareaProps extends DefaultProps {
+		autoComplete?: string;
```
pimterry: This could even be `'on' | 'off'`

typings.d.ts
```diff
@@ -179,11 +205,17 @@ declare module 'resin-components' {
...
 		blank?: boolean;
 		disabled?: boolean;
+		download?: any;
+		href?: string;
+		hrefLang?: string;
+		media?: string;
+		rel?: string;
+		target?: string;
+		type?: string;
+		as?: string;
```
thgreasi: Are `type` and `as` necessary? Aren't those for `<link>` tags?
Couldn't find a place that these are used in the source.

typings.d.ts
thgreasi: The [spec defines way more](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete) than just those two...

typings.d.ts
LucianBuzzo: These are taken from the React typings

typings.d.ts
LucianBuzzo: This is taken from the react typings, I'm happy to leave it as is due to the large variety of possible values

typings.d.ts
LucianBuzzo: The download attribute doesn't actually have to have a value https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download
I guess this means that it could be a boolean.
This attribute is taken from the React typings so I would prefer to follow their leas.

thgreasi: Just to clarify [it can have a string value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Using_the_download_attribute_to_save_a_<canvas>_as_a_PNG) that will be used as the suggested filename. Pretty NIT though (IMO) to discuss this further...

typings.d.ts
thgreasi: Couldn't find those two to be part of `AnchorHTMLAttributes` and that's why I mentioned it.

typings.d.ts
pimterry: > The spec defines way more than just those two...

That's for `<input>`, not `<textarea>`, which has only two: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-autocomplete

pimterry: That said, it doesn't make much of a difference, so if we want to leave this as `string` that's fbm too.

PR #172, opened 2017-11-06
Fix bug where filters still get updated if cancel is pressed

### LINKED ISSUE #171: Filters: filters still get updated if cancel is pressed
See https://github.com/resin-io/resin-ui/issues/1517

LucianBuzzo:
Fixes #171

change-type: patch


src/components/Filters/index.js
```diff
@@ -261,14 +261,21 @@ class Filters extends React.Component {
...
+  }
+
   render () {
     const inputModels = sieve.makeFilterInputs(this.props.schema)
     const rules = this.props.rules || []
 
     return (
       <FilterWrapper mb={3}>
         <Flex justify='space-between'>
-          <Button primary onClick={() => this.toggleModal()}>
```
thgreasi: It seems that `toggleModal()` isn't used any more, can we drop it?

src/components/Filters/index.js
```diff
@@ -261,14 +261,21 @@ class Filters extends React.Component {
     this.props.setViews(views)
   }
 
+  openNewFilterModal () {
```
thgreasi: Would it make sense to be names `showNewFilterModel` since we already have `showEditModal`?

src/components/Filters/index.js
LucianBuzzo: Yep can do

PR #185, opened 2017-11-21
Add typings for sieve.makeFilterInputs

### LINKED ISSUE #184: Add typings for sieve.makeFilterInputs

LucianBuzzo:
Fixes #184

change-type: patch


typings.d.ts
```diff
@@ -269,6 +269,18 @@ declare module 'resin-components' {
...
 	class SchemaSieveClass {
 		filter<T>(items: T[], rule: FilterRule): T[];
+		makeFilterInputs(
+			schema: Schema,
+		): {
+			[key: string]: {
+				type: string;
+				name: string;
+				label: string;
+				availableOperators: object;
```
thgreasi: Isn't this an array?

LucianBuzzo: yep you're correct! good catch

PR #188, opened 2017-11-21
Add titleElement property to Modal component

### LINKED ISSUE #183: Change typings to allow Modal title to be an element
Currently `<Modal title={<h1>Title</h1>}>...</Modal>` works but the typings prohibit it.
This might not be straight-forward since the DefaultProps interface already contains `title?: string` and we can't override that.
Some possible "fixes" would be:
* remove the `title` from the DefaultProps interface
* create a `title-element` property

thgreasi:
Resolves #183
Change-Type: minor


src/components/Modal.js
```diff
@@ -40,34 +40,37 @@ const ModalPanel = styled.div`
...
+        top
+        right
+        bottom
+        left
+        onClick={() => props.cancel()}
+      />
+      <ModalPanel>
+        {!!title && (
+          <Text mb={50}>
+            <strong>{title}</strong>
```
LucianBuzzo: I think we should only be wrapping the `title` in a `<strong>` tag if it is a string

PR #193, opened 2017-11-22
change default horizontal padding of buttons to 20px

### LINKED ISSUE #192: change default horizontal padding of buttons
Currently, buttons have a min-width of 135px. Styleguide suggests that button's width is equal to the button's content plus 20px of left and right padding.

konmouz:
fixes #192

change-type: patch

- Changed standard buttons' padding to 20px
- Changed emphasized buttons' padding to 50px
- Changed standard square buttons' min-width to 40px
- Changed emphasized square buttons' min-width to 50px
- Added story for emphasized square buttons


LucianBuzzo:
@konmouz This has broken the fixed width buttons -->
screen shot 2017-11-22 at 17 47 51

src/components/Button.js
```diff
@@ -33,19 +33,38 @@ const getColor = (props, key, shade) => {
...
+  if (props.w == null && !props.square) {
+    // return 20
+    return props.emphasized ? 50 : 20
   }
   if (props.square) {
-    return props.emphasized ? props.theme.space[5] : props.theme.space[4]
+    return 0
   }
 
   return props.w
```
LucianBuzzo: This line is causing issues for fixed width buttons, as it sets the padding to the width of the button! I think this could be safely changed to:
```
if (props.square || props.w) {
   return 0
}
return props.emphasized ? 50 : 20
```

PR #195, opened 2017-11-28
Improve typings

LucianBuzzo:
change-type: patch


LucianBuzzo:
@thgreasi is `align` a conflicting prop?

thgreasi:
@LucianBuzzo I was talking about `value`

LucianBuzzo:
@thgreasi In this instance, we're not defining some new meaning for 'value'  so I think its ok. In the future with the advent of a full TS project, these types should be generated organically and these issue will be a thing of the past.

PR #198, opened 2017-11-30
Add a `normalize` function to Pine Type modules

### LINKED ISSUE #197: Pine types should have a 'normalize' function
Pine types should have a normalize function that accepts a value and tries to normalize to its intended type.
For example, the `Boolean` type should normalize a string value like `True`, `False`, `1`, and `0` into a true boolean and return it.

LucianBuzzo:
Fixes #197

change-type: minor


src/components/PineTypes/Boolean.js
```diff
@@ -20,6 +21,25 @@ export const rules = {
 
 export const validate = isBoolean
 
+export const normalize = value => {
+  if (isString(value)) {
+    switch (value.toLowerCase()) {
+    case 'true':
```
thgreasi: So, does prettier like aligning the `case`s right below the `switch`!?!?!? :alien: :satellite:

src/components/PineTypes/CaseInsensitiveText.js
```diff
@@ -28,6 +28,8 @@ export const rules = {
 
 export const validate = isString
 
+export const normalize = value => value.toString()
```
thgreasi: Would it make sense to we make it safe to run against null/undefined values (here and below)?

LucianBuzzo: yes indeed, I'll add tests for null and undefined to each of the types

Issue #203, opened 2017-12-04
Alert components are missing


thgreasi:
See: https://github.com/resin-io-modules/resin-components/issues/6

PR #206, opened 2017-12-05
Implement Alert component

thgreasi:
Normal (not in styleguide)
screenshot-2017-12-5 storybook 1
Word wrapping
screenshot-2017-12-5 storybook 2
Emphasized (as found in styleguide)
screenshot-2017-12-5 storybook 3
Outlined
screenshot-2017-12-5 storybook 4
Plaintext (with whitespace padding to align sensisibly with other nearby controls)
screenshot-2017-12-5 storybook 5

Resolves: #203
Change-Type: minor


konmouz:
@thgreasi, I have some issues:
- Not sure we need 'Normal', 'Outlined' should be used when a message is not emphasized.
- 'Plaintext' doesn't need an 'x' button, these messages can't be removed
- The font kerning looks a bit weird. Make sure you use Roboto 16px with 0 kerning everywhere except 'Plaintext' that should be 14px

thgreasi:
* Changed that. 'Outlined' is the new 'Normal'!
* Removed the 'x' button from the 'Plaintext' story
* Added a `font-kerning: none;` just in case, even though, this was not present in zeplin and I would guess that it just might be Firefox rendering things slightly differently. Re-checked the font weights & families.

screenshot from 2017-12-05 13-24-04

PR #207, opened 2017-12-05
Update modal buttons & title

### LINKED ISSUE #194: Update the Modal component button positioning based on the latest wireframes
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/z9kHu72JQ0nt5nMnlmCDKGqEHzx

thgreasi:
Right aligns the modal buttons.
Changes the modal buttons top margin
Hides the cancel button when a respective method is not provided.
Changes the tilte based on the styleguide.
Adds a titleDetails property based on the styleguide.
Adds extra modal stories and makes most of them to have the modals open by default.

Resolves #194
Change-Type: minor


src/stories/Modal.js
```diff
@@ -11,6 +12,7 @@ const Container = styled.div`
 class ModalDemo extends React.Component {
   constructor (props) {
     super(props)
+    this.props = props
```
LucianBuzzo: I think this is unnecessary and can be removed

thgreasi: Nice catch. I was probably testing something and forgot to remove this.

src/components/Modal.js
```diff
@@ -34,7 +44,7 @@ const ModalPanel = styled.div`
   box-shadow: 0px 0px 60px 1px rgba(0, 0, 0, 0.4);
   pointer-events: auto;
 
-  @media (max-width: 730px) {
+  @media (max-width: ${px(MODAL_WIDTH + 30)}) {
```
thgreasi: @LucianBuzzo How about exposing a `width` property and fallback to `MODAL_WIDTH` when `undefined`?
For example [this wireframe](https://app.zeplin.io/project/59ccc17f31f59fb5e145a3ac/screen/5a12fe6efe34e34e9dd215b5) uses 1000px wide modals.

LucianBuzzo: @thgreasi Yeah that sounds great

thgreasi: `calc` doesn't work inside media queries, changed that to use `max-width` instead.

PR #223, opened 2017-12-08
Don't throw an error if a rule contains invalid values

### LINKED ISSUE #221: Don't display invalid rules in filter summary
Connects to https://github.com/resin-io/resin-ui/issues/1593
If a rule does not have a valid corresponding data type it should not be displayed, as the filter logic has no way of actually operating with the rule.

### LINKED ISSUE #222: Don't throw an error if a filter rule has an incorrect type
It's just rude...

Connects to https://github.com/resin-io/resin-ui/issues/1593

LucianBuzzo:
Don't display filter rules with invalid values

Fixes #222
Fixes #221

change-type: minor


src/stories/Filters.js
```diff
@@ -88,6 +104,7 @@ class FiltersDemo extends React.Component {
   }
 
   setRules (rules) {
+    console.log(rules)
```
thgreasi: `console.log`

src/stories/Filters.js
```diff
@@ -63,7 +63,23 @@ class FiltersDemo extends React.Component {
   constructor (props) {
     super(props)
     this.state = {
-      rules: [],
+      // rules: [],
+      rules: [
```
thgreasi: Are these forgotten test rules?

LucianBuzzo: erk yes

src/stories/Filters.js
LucianBuzzo: eep good catch!

PR #234, opened 2017-12-12
Fixes underline buttons

### LINKED ISSUE #201: underline in buttons is larger than text
- Fix underline button: https://resin-io-modules.github.io/resin-components/?selectedKind=Button&selectedStory=Underline&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

konmouz:
fixes #201

change-type: patch


konmouz:
@thgreasi, how can we fix that? add some vertical margin?

thgreasi:
TL;DR Merge as is :+1: 

@konmouz I'm totally convinced that we need to keep with the same height out of the box, so I'm fine with this as is. For example I'm thinking the way that these buttons are used in the billing summary page.

If you insist that we should keep all buttons with the same same size then I would suggest using a pseudo-element for the line. From what I remember from the billing page it looked like:
```css
{
    position: relative;
    padding-left: 0;
    padding-right: 0;
    border: none;

    &:after {
        content: '';
        position: absolute;
        left: 0;
        right: 0;
        bottom: 6px;
        height: 1px;
        background: #youColorHere;
    }

    &:hover:after {
        height: 2px;
    }
}
```

PR #238, opened 2017-12-13
Modal: make cancel prop optional & allow JSX elements for the action

### LINKED ISSUE #237: Modal typings should have the cancel property as optional

thgreasi:
Resolves #237
Change-Type: patch


typings.d.ts
```diff
@@ -270,8 +270,8 @@ declare module 'resin-components' {
 			title?: string;
 			titleElement?: string | JSX.Element;
 			titleDetails?: string | JSX.Element;
-			action: string;
-			cancel: () => any;
+			action: string | JSX.Element;
```
LucianBuzzo: Why do we want to allow JSX elements here?

thgreasi: So that we can have buttons with icons & text.

LucianBuzzo: Ack sounds good

PR #245, opened 2017-12-15
Add type for Key Value Pairs

### LINKED ISSUE #244: Add a KeyValuePair type

LucianBuzzo:
Fixes #244

change-type: minor


src/components/Filters/Summary.js
```diff
@@ -131,6 +131,7 @@ class FilterSummary extends React.Component {
             return (
               <Box mb={10} mr={10} key={rule.id}>
                 <FilterDescription
+                  schema={this.props.schema[rule.name]}
```
thgreasi: This should also be added to the ViewsMenu.js and be passed to the FilterDescription component (so that we can save an restore views).

src/components/Filters/index.js
```diff
@@ -361,6 +366,7 @@ class Filters extends React.Component {
               saveView={(name, scope) => this.saveView(name, scope)}
               rules={rules}
               views={this.props.views || []}
+              schema={this.props.schema}
```
thgreasi: This should be also passed to the ViewsMenu component (so that we can save an restore views).

src/components/PineTypes/KeyValuePair.js
```diff
@@ -0,0 +1,149 @@
...
+const normalizeToCollection = value => (isArray(value) ? value : [value])
+
+export const rules = {
+  is: (target, value) => {
+    return some(normalizeToCollection(target), value)
+  },
+  'is not': (target, value) => {
+    return !some(normalizeToCollection(target), value)
+  },
+  'key is': (target, value) => {
```
thgreasi: Wouldn't it be great if this could be an object like:
```
'key is': {
  getLabel: (type) => { ... }, // this would be optional
  test: (target, value) => { ... },
}
```
So the operator could also say "tag name" just like the placeholder's input.
This way the SchemaSieve `getOperators()` method would check whether the prop is an object (or not) and whether a getLabel() method and use that to populate the array.

Maybe in the next iteration.

src/components/Filters/Summary.js
LucianBuzzo: Good spot. We need TS here :(

package.json
```diff
@@ -14,6 +14,7 @@
     "build": "rimraf dist && tsc",
     "build:storybook": "build-storybook -o .out",
     "test": "npm run lint && mocha --require ts-node/register test/*.spec.js",
+    "mocha": "mocha --require ts-node/register test/*.spec.js",
```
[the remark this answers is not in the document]
LucianBuzzo: yep

src/components/PineTypes/KeyValuePair.js
LucianBuzzo: I'll implement this suggestion in this PR as it will definitely make the feature more complete

LucianBuzzo:
@thgreasi I've made some updates and addressed all your comments. It feels a lot nicer now!

src/components/Filters/FilterDescription.js
```diff
@@ -12,15 +16,43 @@ const ButtonWrapper = styled.button`
   padding: 3px 8px;
 `
 
+const getOperatorText = (operator, type, schemaEntry) => {
+  const operatorItem = PineTypes[type].rules[operator]
+  if (isPlainObject(PineTypes[type].rules[operator]) && schemaEntry) {
```
thgreasi: NIT: I guess that this could be `isPlainObject(operatorItem)`.

thgreasi:
NIT: Also tags are not present in the story book, so you don't really know how to test this unless you check the code.

src/components/Filters/FilterDescription.js
LucianBuzzo: oops

LucianBuzzo:
@thgreasi I'll add a tag display to the story

thgreasi:
Woohoo ! Cool!

PR #247, opened 2017-12-18
Button: add plaintext variant

### LINKED ISSUE #246: Implement plaintext button

thgreasi:
Resolves #246
Change-Type: minor


thgreasi:
@konmouz this can be used both w/ and w/o icons, its just a matter of the content that we will be providing for each use.
I was hoping to use it as shown in the styleguide & the panel wireframes where the text font size is 16px:
* https://app.zeplin.io/project/59cbb35a8edcbe2d8fbe5678/screen/59cd74ed3b9450c01addba94
* https://app.zeplin.io/project/59cbb35a8edcbe2d8fbe5678/screen/59d4f58a68cf621d3adab6e5 (which is a list)

Removed the alternative colors from the storybook & added one with an icon.
Do we need to change the text to be 14px font size?

PR #251, opened 2017-12-20
Gracefully handle situations where a filter rule refers to a schema entry that is non-exist

### LINKED ISSUE #224: Gracefully handle situations where a filter rule refers to a schema entry that is non-existent
If a rules `name` property is a value that is not a key in the schema, the filter component will throw an error when trying to edit that rule.

LucianBuzzo:
The filter component will now actively filter out rules that refer to
non-existent schema entries.

Fixes #224

change-type: patch


test/filter.spec.js
```diff
@@ -1144,5 +1144,32 @@ describe('resin-filter', () => {
...
+      const inputs = sieve.makeFilterInputs(schema)
+      const input = inputs.test
+
+      // Set the input name to something that doesn't exist in the schema
+      input.name = 'foobar'
+
+      input.operator = 'contains'
+      input.value = 'abc'
+
+      console.log(input)
```
thgreasi: I guess we can get rid of this.

test/filter.spec.js
```diff
@@ -1144,5 +1144,32 @@ describe('resin-filter', () => {
 
       expect(sieve.filter(collection, input)).to.have.length(2)
     })
+
+    it('should behave correctly rule references a non-existent schema field', function () {
```
thgreasi: This seems weird to me, would it make more sense like `...when a rule references...`?

test/filter.spec.js
```diff
@@ -1144,5 +1144,32 @@ describe('resin-filter', () => {
...
+
+      // Set the input name to something that doesn't exist in the schema
+      input.name = 'foobar'
+
+      input.operator = 'contains'
+      input.value = 'abc'
+
+      console.log(input)
+
+      expect(sieve.filter(collection, input)).to.have.length(0)
```
thgreasi: When trying `npm test` this didn't fail on the old source!?!? Can you confirm this?

src/components/Filters/index.js
```diff
@@ -89,6 +89,9 @@ class Filters extends React.Component {
       edit: this.generateFreshEdit(),
       searchString: (existingRule && existingRule.value) || ''
     }
+
+    // Clean exsting rules on load
+    this.filterAndSetRules(rules)
```
thgreasi: **This might be an extreme case**, but this would not protect us in case the consumer mutates or assigns a new array with invalid rules after the component initializes. Should we add something similar to `componentWillReceiveProps`?

Alternatively, would it make sense to only have sanitization code in the render function, which the single place responsible to distribute the models in the subcomponents?
```js
const rules = this.filterInvalid(this.props.rules || [])
```
PS: I guess that the second will be costly, so let's forget it.

test/filter.spec.js
LucianBuzzo: @thgreasi Yep this is correct, the test here is more of a safegaurd against future regressions

test/filter.spec.js
LucianBuzzo: Yes you're right, this english is not good :(

src/components/Filters/index.js
LucianBuzzo: I would argue that at that point we've tried our best to help the consumer, if they have bugged code, they are on their own :)

test/filter.spec.js
LucianBuzzo: yes indeed

test/filter.spec.js
thgreasi: Would also love to have a test case that would fail without this PRs changes as a safegaurd against future regressions.

LucianBuzzo: The filtering logic actually behaves correctly in these situations, its the UI that fails :/
We need integration tests here ideally.

LucianBuzzo: I've created an issue for this https://github.com/resin-io-modules/resin-components/issues/254

PR #253, opened 2017-12-20
Add table component

### LINKED ISSUE #157: Add a table component

LucianBuzzo:
Fixes #157

change-type: minor


tsconfig.json
```diff
@@ -28,6 +28,7 @@
   "include": [
     "./src/**/*.js",
     "./src/**/*.ts",
-    "./src/**/*.d.ts"
+    "./src/**/*.d.ts",
+    "./typings.d.ts"
```
thgreasi: It seems that this doesn't affect the build at all.
There is some overlapping with `common-types.d.ts` and some differences for the same stuff as well.

src/utils.ts
```diff
@@ -3,6 +3,10 @@ import get = require('lodash/get');
 import isObject = require('lodash/isObject');
 import * as Color from 'color';
 
+interface ThemedDefaultProps extends DefaultProps {
```
thgreasi: This could be `ThemedStyledProps<DefaultProps, Theme>` or `DefaultProps & ThemeProps<Theme>` in case we want to use existing interfaces.

src/components/Table.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+interface TableHeaderItem {
+	field: string;
+	icon?: string;
+	label?: string | JSX.Element;
+	sortable?: boolean | TableSortFunction;
+}
+
+interface TableProps {
+	header: TableHeaderItem[];
```
thgreasi: This seems like a `columns` definition to me instead of `header`. What do you think?

src/components/Table.tsx
```diff
@@ -0,0 +1,177 @@
...
+	[key: string]: TableDataItem;
+}
+
+type TableSortFunction = <T>(a: T, b: T) => number;
+
+interface TableHeaderItem {
+	field: string;
+	icon?: string;
+	label?: string | JSX.Element;
+	sortable?: boolean | TableSortFunction;
```
thgreasi: I would also love to be able to pass a `render()` parameter to control how each field is presented!
Probably in a separate PR.

src/components/Table.tsx
```diff
@@ -0,0 +1,177 @@
...
+`
+
+const renderField = (row: TableDataRow, field: string) => {
+	const value = get(row, field);
+
+	if (!value) {
+		return '';
+	}
+
+	return value.hasOwnProperty('display') ? (value as any).display : value;
```
thgreasi: I think that this should be `isPlainObject(value) && value.hasOwnProperty('display')`.

thgreasi: So, if I get this right, each item value can be an object with a prerendered element as it's value?
In that case, the consumer is responsible to update the data-elements based on state changes?

IMO a render/display function on the column definition is more sensible and easier to reason, but I guess that you already have a use case that this makes huge difference. If that's the case then I'm fine with this.

(Had separation of data & presentation in mind, but I'm not sure whether this fits in a React world.)

src/components/Table.tsx
```diff
@@ -0,0 +1,177 @@
...
+										<HeaderButton
+											plaintext
+											primary={this.state.sortColumn === item.field}
+											onClick={() =>
+												this.toggleSort(item.field)}
+										>
+											<Flex align="center">
+												{item.label || item.field}
+												&nbsp;
+												<FaSort color={this.state.sortColumn === item.field ? theme.colors.info.main : ''} />
```
thgreasi: That's probably forgotten but I guess that the icons should change based on the current sorting state.

src/components/Table.tsx
```diff
@@ -0,0 +1,177 @@
...
+							}
+							return <th key={item.field}>{item.label}</th>;
+						})}
+					</tr>
+				</thead>
+				<tbody>
+					{map(
+						this.sortData(data), (row, i) => {
+						return (
+							<tr key={i}>
```
thgreasi: Could the consumer provide a custom `key` to improve the re-rendering process?

src/components/Table.tsx
LucianBuzzo: That's an interesting approach and one I hadn't thought of. The separation of data & presentation does get a bit blurry in the react world for sure.
The reason why I took this approach is because it makes it trivially easy to build interactive UI based on a data item. See here for a slightly messy example --> https://github.com/resin-io/resin-ui/blob/87caac6e2bd808d5ee5652bb3889bef60794f9c3/src/scripts/modules/environment-variables/config-variables.tsx#L234

The idea of having all the presentational information in the header (column) prop and then the data just being raw data is quite attractive.
I'm going to let this brew for an hour and come back to it :)

src/components/Table.tsx
LucianBuzzo: Interestingly, this is how the current sort headers work and how the zeplin designs show it. Lets talk to @konmouz and see if he has any opinions

src/components/Table.tsx
LucianBuzzo: Yep good idea

src/components/Table.tsx
thgreasi: That example is straightforward :+1:

src/common-types.d.ts
```diff
@@ -18,7 +18,6 @@ interface Theme {
 }
 
 interface StyledSystemProps {
-	theme: Theme;
```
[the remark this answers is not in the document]
LucianBuzzo: This meant that you had to include a `theme` prop when using the `<Button>` component :/

src/components/Button.tsx
```diff
@@ -12,9 +19,13 @@ interface ButtonProps extends DefaultProps, Coloring, Sizing {
 	iconElement?: JSX.Element;
 }
 
+interface ThemedButtonProps extends ButtonProps {
```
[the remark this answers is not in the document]
LucianBuzzo: We typically prefer interfaces over types, is there a reason to use a type here?

src/components/Table.tsx
LucianBuzzo: yep I like it

src/components/Button.tsx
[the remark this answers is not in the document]
thgreasi: Totally with you on interface preference (with the exception of empty interfaces that only inherit from others, where tslint prefers `type`s).

Actually I was just saying that we could use `ThemedStyledProps<ButtonProps, Theme>` or `ThemeProps<Theme>` (even in the interface) instead of defining the `theme` prop ourselves. Perhaps some of the built in styled-component methods could have better inference if we do so. No strong opinions though.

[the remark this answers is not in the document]
LucianBuzzo: Ah I see what you mean, yep sounds good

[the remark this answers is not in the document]
thgreasi: I guess we can keep this as is. Was NIT anyway, although the code would be DRYer even if we had such an interface in our `utils.ts`.

src/components/Table.tsx
```diff
@@ -0,0 +1,179 @@
...
+`;
+
+/**
+ * Get the value specified by the `field` value
+ * If a `render` function is available, use it to get the display value.
+ */
+const renderField = (row: TableDataRow, column: TableColumn) => {
+	const value = get(row, column.field);
+
+	if (!value) {
```
thgreasi: What if we wanted `0` to show up? Should this only apply for `null` & `undefined`?

src/components/Table.tsx
```diff
@@ -0,0 +1,179 @@
...
+		return '';
+	}
+
+	return column.render ? column.render(value) : value;
```
thgreasi: I think that when a render method exists, it should be fed with the raw value.
How about moving this before the `if (!value)` so that a render function can handle `undefined`/`null`/`0` in any crazy way?

src/components/Table.tsx
LucianBuzzo: Good catch

src/components/Table.tsx
LucianBuzzo: yep, I also think we should be passing in the full row to the render function as a second parameter, so that you can work with the presented data if necessary. For example, being able to access the devices uuid in the bulb component.

thgreasi: Totally makes sense :+1:

src/components/Table.tsx
```diff
@@ -22,7 +22,7 @@ interface TableColumn {
 	icon?: string;
 	label?: string | JSX.Element;
 	sortable?: boolean | TableSortFunction;
-	render?: (value: any) => string | JSX.Element;
+	render?: (value: any, context: any) => string | JSX.Element;
```
thgreasi: NIT: how about naming this as `dataItem` or `row`?

LucianBuzzo: I'll change it to row

PR #256, opened 2017-12-21
Fix Button styles when using `bg` and `color` props together

### LINKED ISSUE #255: Dropdown button doesn't behave as expected when setting color
When using the color prop on a `DropDownButton`, the color is not respected on hover.

LucianBuzzo:
Fixes #255

change-type: patch


src/components/Button.tsx
```diff
@@ -89,6 +89,13 @@ const Outline = Button.extend`
 		getColor(props, 'color', 'main') || props.theme.colors.text.main};
 	background: ${props => props.color || 'none'};
 	border: 1px solid;
+
+	&:hover,
+	&:focus,
+	&:active {
+		background-color: ${props => props.bg ?
+			darken(props.bg as string) : (getColor(props, 'bg', 'dark') || props.theme.colors.tertiary.dark)};
```
thgreasi: What's the case that this fallback to the tertiary color is used?

LucianBuzzo: This happens if you don't specify a primary / seconday or tertiary option. eg
```
<Button outline>button</Button>
```

Issue #259, opened 2017-12-22
Getting prettier-eslint errors


thgreasi:
Fresh `npm i` on master
> \> npm run prettify
```
Trace: prettier-eslint [TRACE]: requiring "prettier" module at "prettier"
    at Logger.trace (/home/teo/Resin.io/resin-components/node_modules/loglevel-colored-level-prefix/dist/index.js:54:24)
    at requireModule (/home/teo/Resin.io/resin-components/node_modules/prettier-eslint/dist/utils.js:433:12)
    at getPrettierConfig (/home/teo/Resin.io/resin-components/node_modules/prettier-eslint/dist/index.js:229:43)
    at format (/home/teo/Resin.io/resin-components/node_modules/prettier-eslint/dist/index.js:90:87)
    at MapSubscriber.project (/home/teo/Resin.io/resin-components/node_modules/prettier-eslint-cli/dist/format-files.js:305:55)
    at MapSubscriber._next (/home/teo/Resin.io/resin-components/node_modules/rxjs/operators/map.js:79:35)
    at MapSubscriber.Subscriber.next (/home/teo/Resin.io/resin-components/node_modules/rxjs/Subscriber.js:90:18)
    at AsyncSubject.Subject.next (/home/teo/Resin.io/resin-components/node_modules/rxjs/Subject.js:55:25)
    at AsyncSubject.complete (/home/teo/Resin.io/resin-components/node_modules/rxjs/AsyncSubject.js:46:35)
    at handlerFn (/home/teo/Resin.io/resin-components/node_modules/rxjs/observable/BoundNodeCallbackObservable.js:193:33)
    at tryToString (fs.js:457:3)
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:444:12)
```

Explicitly using the previous `prettier-eslint-cli` (4.4.0) doesn't fix the issue. It looks like some of their dependencies didn't respect semver.

LucianBuzzo:
@thgreasi can this be closed?

thgreasi:
Oh yea, good catch, considered no longer applicable since #260

PR #260, opened 2017-12-22
Add prettier & lint for Typescript files

thgreasi:
Also removes prettier-eslint-cli (since it was giving errors #259) in favor of the latest prettier & tslint (obviously np to revert that part if we don't like it).


package.json
```diff
@@ -41,12 +48,14 @@
...
-    "prettier-eslint-cli": "~4.6.0",
+    "prettier": "^1.9.2",
     "react": "^15.6.2",
     "react-dom": "^15.6.0",
     "rimraf": "^2.6.2",
     "ts-loader": "^3.0.2",
-    "ts-node": "^3.3.0"
+    "ts-node": "^3.3.0",
+    "tslint": "^5.8.0",
+    "typescript": "^2.6.2"
```
thgreasi: So, all this time the TS part of the build was using the local machine's typescript installation!

src/components/Alert.tsx
```diff
@@ -23,11 +24,11 @@ const getTitle = (props: AlertProps) => {
...
-	return props.danger ? 'Darn!' :
-		props.warning ? 'Warning!' :
-		props.success ? 'Great!' :
-		props.info ? 'Hey!' :
-		'';
+	return props.danger
+		? 'Darn!'
+		: props.warning
+			? 'Warning!'
+			: props.success ? 'Great!' : props.info ? 'Hey!' : '';
```
thgreasi: I hate how prettier formats multi-ternary expressions.
They have an issue to track this... but still no progress https://github.com/prettier/prettier/issues/737

package.json
LucianBuzzo: Doesn't look like it, though that is probably entirely by accident!
```
$ resin-components/ (255-dropdown-color) ll node_modules | grep typescript
drwxr-xr-x   15 lucianbuzzo  staff   510B 20 Dec 17:33 typescript/
drwxr-xr-x    8 lucianbuzzo  staff   272B 20 Dec 17:33 typescript-eslint-parser/
```

I did some digging and it looks like `prettier-eslint` installs typescript as a dependency.

thgreasi: Oh, that's cool(er) then :)

PR #268, opened 2018-01-02
Add selection functionality to Table component

### LINKED ISSUE #266: Add checkbox functionality to table component

LucianBuzzo:
Fixes #266

change-type: minor


src/components/Table.tsx
```diff
@@ -50,6 +53,7 @@ const BaseTable = styled.table`
 	> tbody {
 		> tr {
 			> td {
+				text-align: left;
```
thgreasi: **In a future releast**, this will probably have to be configurable as part of the column and as a fallback be inferred by the data type (if it's a primitive). eg: numeric columns would be aligned right by default.

I could reason that using a custom column render function gives us huge power, but it might be an overkill if you have to use it just for aligning stuff. Sensible defaults are also attractive :)

PS: this is probably not the place/PR to start such discussion.

src/components/Table.tsx
```diff
@@ -50,6 +53,7 @@ const BaseTable = styled.table`
...
 				font-size: 16px;
```
thgreasi: Out of context but isn't that supposed to be 14px in the body?

src/components/Table.tsx
```diff
@@ -81,42 +88,93 @@ const renderField = (row: TableDataRow, column: TableColumn) => {
...
 
-	sortData(data: TableDataRow[]): TableDataRow[] {
+	isChecked(item: T) {
+		const rowKey = this.props.rowKey;
+		if (!rowKey) {
+			return false;
+		}
+
+		const identifier = item[rowKey];
+		return !!find(this.state.checkedItems, x => x[rowKey] === identifier);
```
thgreasi: Could this be `some(this.state.checkedItems, { [rowKey]: identifier }`?

src/components/Table.tsx
```diff
@@ -81,42 +88,93 @@ const renderField = (row: TableDataRow, column: TableColumn) => {
...
 				: sortableValue;
-		}) as any) as TableDataRow[];
+		});
 
 		if (this.state.reverse) {
 			reverse(collection);
 		}
 
-		return collection;
+		return (collection as any) as T[];
```
thgreasi: I think that we can get rid of this casting if we use `sortBy<T>`.

src/components/Table.tsx
```diff
@@ -81,42 +88,93 @@ const renderField = (row: TableDataRow, column: TableColumn) => {
...
+		if (!rowKey) {
+			return false;
+		}
+
+		const identifier = item[rowKey];
+
+		const isChecked = !this.isChecked(item);
+		const checkedItems = isChecked
+			? this.state.checkedItems.concat(item)
+			: this.state.checkedItems.filter(x => x[rowKey] !== identifier);
```
thgreasi: I think that this can be `remove({ [rowKey]: identifier })` but I'm fine with `filter` as well in an effort to reference less lodash methods.

src/components/Table.tsx
```diff
@@ -81,42 +88,93 @@ const renderField = (row: TableDataRow, column: TableColumn) => {
...
+
+		if (this.props.onChecked) {
+			this.props.onChecked(checkedItems);
+		}
+
+		this.setState({ allChecked: false, checkedItems });
```
thgreasi: I'm not sure about the `allChecked: false` part.
Shouldn't this become `true` if we select all the table items one by one?

src/components/Table.tsx
```diff
@@ -134,6 +192,15 @@ export default class Table extends React.Component<
 			<BaseTable {...props}>
 				<thead>
 					<tr>
+						{this.props.onChecked && (
+							<th>
+								<input
```
thgreasi: I guess we could **in a later release** allow custom elements for this one as well.
Probably by making a union type for the column, where when `onChecked` is defined then
```ts
{
  ...
  render?: (...) =>` JSX.Element & { checked?: boolean; onChange: () => void };
  ...
}
```

src/components/Table.tsx
```diff
@@ -165,7 +232,16 @@ export default class Table extends React.Component<
 				<tbody>
 					{map(this.sortData(data), (row, i) => {
 						return (
-							<tr key={rowKey ? row[rowKey] : i}>
+							<tr key={rowKey ? (row[rowKey] as any) : i}>
+								{this.props.onChecked && (
+									<td>
+										<input
```
thgreasi: Should/Could we use the rendtion `<Input />` here, so that if/when we implement a custom checkbox, the table will get it automatically?

src/stories/Table.js
```diff
@@ -30,6 +30,10 @@ const columns = [
     label: 'First Seen',
     sortable: true,
     render: value => {
+      if (!value) {
```
thgreasi: Should this be `== null` so that it renders `0`s?

src/components/Table.tsx
```diff
@@ -11,25 +12,26 @@ import theme from '../theme';
...
+	columns: Array<TableColumn<T>>;
+	data: T[];
 	// Optionally provide a key that should be used as a unique identifier for each row
-	rowKey?: string;
+	rowKey?: keyof T;
+
+	// Only usable if a rowKey property is also provided.
+	// If an onChecked property is provided , then checkboxes will be renders,
+	// allowing rows to be selected.
+	onChecked?: (checkedItems: T[]) => void;
```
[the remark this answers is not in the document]
LucianBuzzo: `onCheck` sounds good here

src/components/Table.tsx
LucianBuzzo: @thgreasi Interesting, this is definitely something to think about as we move forward with he project

src/components/Table.tsx
LucianBuzzo: Good spot!

src/components/Table.tsx
LucianBuzzo: Classy!!

src/components/Table.tsx
LucianBuzzo: Nice, I thought there was a simple solution here :)

src/components/Table.tsx
LucianBuzzo: I like the object matching shorthand, I think the method we want here is `reject` though.

src/components/Table.tsx
LucianBuzzo: Yup, ill update it

src/components/Table.tsx
LucianBuzzo: Yep could do

src/components/Table.tsx
```diff
@@ -116,7 +158,41 @@ export default class Table extends React.Component<
...
+
+		const identifier = item[rowKey];
+
+		const isChecked = !this.isChecked(item);
+		const checkedItems = isChecked
+			? this.state.checkedItems.concat(item)
+			: reject(this.state.checkedItems, { [rowKey]: identifier });
+
+		if (this.props.onCheck) {
+			this.props.onCheck(checkedItems);
```
thgreasi: > Should we also pass `the` item whose checkbox was toggled?

LucianBuzzo: @thgreasi I don't think we should. I prefer a behaviour similar to filters, where we simply list the updated object collection (like `setRules`). For most of our use cases, knowing which items was just checked/unchecked is not useful.

thgreasi: Ok, I guess that if such a need ever appears, we could add it easily.

PR #273, opened 2018-01-03
Export Table from module

LucianBuzzo:
change-type: patch


thgreasi:
I would prefer to have this merged after #268 where we are changing the interfaces used by the Table component.

LucianBuzzo:
This will be resolved by https://github.com/resin-io-modules/rendition/pull/268

PR #277, opened 2018-01-05
Filter: autofocus save view modal input & handle form submits

### LINKED ISSUE #239: Change title of add filter modal
Change the title from `Add a new filter` to `Filter by`

### LINKED ISSUE #276: Filter save view modal should autofocus input & handle <enter>s

thgreasi:
Also uses cursor as the pointer for the respective button and fixes form submit for add/edit filter modal.

Resolves #276
Resolves #239
Change-Type: minor


src/components/Filters/index.js
```diff
@@ -330,15 +330,13 @@ class Filters extends React.Component {
           <div>
             <Modal
               title={
-                this.state.edit.id
-                  ? 'Update existing filter'
-                  : 'Add a new filter'
+                this.state.edit.id ? 'Update existing filter' : 'Filter by'
```
thgreasi: Would also make sense to me to change `Update existing filter` to something simpler like `Update filter` or even use `Filter by` for both cases. What do you think @LucianBuzzo ?

LucianBuzzo: Lets use `Filter by` in both cases

PR #289, opened 2018-01-09
Modal: use a css class to manipulate body element scroll

### LINKED ISSUE #282: The Modal component should use a class to manipulate body element scroll

thgreasi:
I was initially using `randomString()` (from the utils), but after some research on the styed-components repo & how other react modal components do such things, it seems that the community converges to using plain prefixed well known class names. The random class name doesn't add any benefit, while a well known one can be exploited in CSS selectors in the consumer projects.

This also:
* converts the Fixed & Modal components to tsx
* references the FixedProps & ModalProps from the typings/rendition.d.ts. If we like this and adopt it in the rest of the tsx components, the result will be DRYer and will no longer get in a situation where the component's internal interfaces are (accidentally) different from those that we expose in the typings.

Resolves #282
See: #44
Change-Type: minor


src/components/Modal.tsx
```diff
@@ -1,8 +1,9 @@
 import assign = require('lodash/assign');
 import * as React from 'react';
+import { ModalProps } from 'rendition';
```
LucianBuzzo: Sweet, is this how we can import typings from our definition without redefining them?

thgreasi: I think so, everything seems to work until now (also tested the integration with the UI).

PR #305, opened 2018-01-19
Add `onRowClick` property and highlight style to Table component

### LINKED ISSUE #303: Add option to Table to allow click action on row
You should be able to define a function that will be triggered when a table row is clicked.
It should run the provided function with the the row data of the row that is clicked.

LucianBuzzo:
Fixes #303

change-type: minor

screen shot 2018-01-19 at 10 31 04

screen shot 2018-01-19 at 10 31 29


src/components/Table.tsx
```diff
@@ -9,34 +9,22 @@ import some = require('lodash/some');
...
-	// Optionally provide a key that should be used as a unique identifier for each row
-	rowKey?: keyof T;
-
-	// Only usable if a rowKey property is also provided.
-	// If an onCheck property is provided , then checkboxes will be renders,
-	// allowing rows to be selected.
-	onCheck?: (checkedItems: T[]) => void;
-}
+	> td:first-child {
+		border-left: 3px solid ${theme.colors.info.main};
```
thgreasi: Ultra-NIT: this could be just shadow so that the browser only repaints instead of rerendering to calculate the new width.

src/components/Table.tsx
```diff
@@ -249,12 +247,20 @@ export default class Table<T> extends React.Component<
 				</thead>
 				<tbody>
 					{map(this.sortData(data), (row, i) => {
+						const isChecked = this.props.onCheck ? this.isChecked(row) : false;
 						return (
-							<tr key={rowKey ? (row[rowKey] as any) : i}>
+							<tr
+								data-checked={isChecked}
+								style={{ display: 'table-row' }}
```
thgreasi: Is this required?

src/components/Table.tsx
```diff
@@ -249,12 +247,20 @@ export default class Table<T> extends React.Component<
 				</thead>
 				<tbody>
 					{map(this.sortData(data), (row, i) => {
+						const isChecked = this.props.onCheck ? this.isChecked(row) : false;
 						return (
-							<tr key={rowKey ? (row[rowKey] as any) : i}>
+							<tr
+								data-checked={isChecked}
```
thgreasi: I was surprised to see a dash property in tsx!

src/components/Table.tsx
```diff
@@ -67,6 +57,14 @@ const BaseTable = styled.table`
 			&:nth-of-type(even) {
 				background-color: #f8f8f8;
 			}
+
+			&[data-checked='true'] {
+				${highlightStyle};
+			}
+
+			&:hover {
+				${(props: any) => (!!props.onRowClick ? highlightStyle : '')};
```
thgreasi: Loved this :+1: 
My only request in this review: Would it make sense to not add the `theme.colors.info.main` part in front for the hover effect and use it only for selection?

src/components/Table.tsx
LucianBuzzo: Do you mean `box-shadow`?

src/components/Table.tsx
LucianBuzzo: nope

src/components/Table.tsx
LucianBuzzo: Yep I think that makes sense

src/components/Table.tsx
thgreasi: yea, sorry for not clarifying

LucianBuzzo: Ah yep, good idea

PR #310, opened 2018-01-22
Only change Table row bg on hover if onRowClick or onCheck props are …

### LINKED ISSUE #309: Don't user hover background if table has no `onRowClick` prop set

LucianBuzzo:
…used

Fixes #309

change-type: minor


LucianBuzzo:
I'll pitch it to a product call where we can discuss the behaviour with @konmouz

LucianBuzzo:
Approved in product call

PR #312, opened 2018-01-22
Allow a custom JSX element to be added to the top of a Table

### LINKED ISSUE #311: Allow a custom JSX element to be added into a Table
The element is provided as `rowPrefix` and will be rendered before the main rows in the Table. These rows are not affected by sorting.

LucianBuzzo:
Fixes #311

change-type: minor


src/components/Table.tsx
```diff
@@ -243,6 +243,7 @@ export default class Table<T> extends React.Component<
 					</tr>
 				</thead>
 				<tbody>
+					{this.props.rowPrefix}
```
thgreasi: Would it make sense to:
* name this `bodyPrefix` just in case we need to add a `headPrefix/Suffix` in the future?
* have a plural name & type? (perhaps not)
* not apply the hover effect on these items?

LucianBuzzo: I'm in agreement on the name change to `bodyPrefix`.
It shouldn't be plural, as it must be a single JSX element.
If the user doesn't want the default hover effect, it is up to them to override it with a style.

thgreasi: > It shouldn't be plural, as it must be a single JSX element.

Yea, I was thinking whether we might want more than one items, but I guess that's too much.

LucianBuzzo: Looking at the work I'm doing at the moment, I just realised that I have need for many rows as a prefix :joy: d'oh!

PR #324, opened 2018-01-23
Add BadgeSelect component

### LINKED ISSUE #323: Add BadgeSelect component
screen shot 2018-01-23 at 15 18 46

LucianBuzzo:
Fixes #323

change-type: minor


src/components/BadgeSelect.tsx
```diff
@@ -0,0 +1,98 @@
+import has = require('lodash/has');
+import includes = require('lodash/includes');
+import * as React from 'react';
+import { BadgeSelectProps } from 'rendition';
+import styled from 'styled-components';
+import Badge from './Badge';
+import DropDownButton from './DropDownButton';
+import Text from './Text';
+
+interface BadgeSelectState {
```
thgreasi: NIT: how about moving this right above the component?
I was getting lost toooooo easily before starting following that rule.

src/components/BadgeSelect.tsx
```diff
@@ -0,0 +1,98 @@
...
+	padding-right: 10px;
+	padding-top: 3px;
+	padding-bottom: 3px;
+
+	&:hover {
+		background-color: #f0f3f7;
+	}
+`;
+
+class BadgeSelect extends React.Component<BadgeSelectProps, BadgeSelectState> {
```
thgreasi: Would it make sense to name like `BadgeSelector` or `BadgeDropdown`? No strong feelings on this though.

src/components/BadgeSelect.tsx
```diff
@@ -0,0 +1,98 @@
...
+	background: none;
+	display: block;
+	margin-top: 5px;
+	margin-bottom: 5px;
+	text-align: left;
+	min-width: 182px;
+	padding-left: 10px;
+	padding-right: 10px;
+	padding-top: 3px;
+	padding-bottom: 3px;
```
thgreasi: I think that we also need to define a cursor in here.

src/components/BadgeSelect.tsx
```diff
@@ -0,0 +1,98 @@
...
+							)}
+						{this.state.selected != null &&
+							!includes(this.props.items, this.state.selected) &&
+							this.state.selected}
+					</Text>
+				}
+			>
+				{this.props.items.map(item => (
+					<ButtonWrapper key={item} onClick={() => this.setSelected(item)}>
+						<Badge text={item} />
```
thgreasi: NIT: would it make sense to make these block level?

src/components/BadgeSelect.tsx
```diff
@@ -0,0 +1,98 @@
...
+					</ButtonWrapper>
+				))}
+
+				{!!this.props.extra &&
+					this.props.extra.map(item => (
+						<ButtonWrapper key={item} onClick={() => this.setSelected(item)}>
+							<Text>{item}</Text>
```
thgreasi: Would it make sense to use a white badge for these, to properly align them with the rest of the items?

LucianBuzzo: I'd much prefer that the non-badge text is aligned with the side of the badges, not the text within them.

src/components/BadgeSelect.tsx
LucianBuzzo: :+1:

src/components/BadgeSelect.tsx
LucianBuzzo: I considered both of these, and settled on BadgeSelect as being the most semantic

src/components/BadgeSelect.tsx
LucianBuzzo: I don't understand what you mean by 'block level', can you explain please?

src/components/BadgeSelect.tsx
LucianBuzzo: can do

src/components/BadgeSelect.tsx
thgreasi: I meant to have 100% width.

LucianBuzzo: Ah ok, the designs are not 100% width

PR #331, opened 2018-01-26
Modal: make action prop optional and default to 'Ok'

### LINKED ISSUE #330: Modal: Add default for 'action' prop

LucianBuzzo:
Fixes #330

change-type: minor


LucianBuzzo:
@thgreasi The styleguide just shows the text `Main CTA`. We typically use sentence case for CTA text, and I think 'Ok' is a generic enough confirmation word to use here.

screen shot 2018-01-26 at 16 30 07

thgreasi:
@LucianBuzzo totally agree about using ok, I was just wondering how the sentence case of acronyms like ok should be  :)

LucianBuzzo:
@thgreasi Figuring out the 'proper' way to write ok turns out to be a bit of a minefield
- http://www.writing-skills.com/ok-ok-okay-how-do-you-write-ok
- https://en.wikipedia.org/wiki/OK
- https://english.stackexchange.com/questions/108213/must-ok-only-be-written-in-capital-letters

@sonyagreen @zwalchuk You guys might be able to give us some useful info, how should we display the work 'OK' on the modal confirmation?

sonyagreen:
Oooh, good resin style guide question. As much as I love a good grammatical issue, I think this is more about the look and feel of our product, and what's most universal.
* is "OK" the word we want? Would we consider something a bit more cheeky like "Got it!" 
* is there an industry standard for this?

zwalchuk:
I don't think I'm qualified to comment on the appropriate amount of cheekiness for a button, but if we do stick with OK I think it should definitely be all capital.

LucianBuzzo:
Thanks for the feedback @zwalchuk @sonyagreen 
I think we should go with "OK" for the CTA text.

PR #345, opened 2018-01-31
Add badges to README

LucianBuzzo:
change-type: patch

The gitter link isn't working just yet, but should be set up very soon

badges


README.md
```diff
@@ -2,7 +2,11 @@
 
 [![npm version](https://badge.fury.io/js/rendition.svg)](http://badge.fury.io/js/rendition)
 [![Build Status](https://travis-ci.org/resin-io-modules/rendition.svg?branch=master)](https://travis-ci.org/resin-io-modules/rendition)
+[![Dependency Status](https://img.shields.io/david/resin-io-modules/rendition.svg)](https://david-dm.org/resin-io-modules/rendition)
+[![npm](https://img.shields.io/npm/dm/rendition.svg?maxAge=2592000)]()
+[![GitHub last commit](https://img.shields.io/github/last-commit/resin-io-modules/rendition.svg)]()
```
thgreasi: Empty url :police_car: . How about pointing it to https://www.npmjs.com/package/rendition ?

LucianBuzzo: 👍

README.md
```diff
@@ -2,7 +2,11 @@
 
 [![npm version](https://badge.fury.io/js/rendition.svg)](http://badge.fury.io/js/rendition)
 [![Build Status](https://travis-ci.org/resin-io-modules/rendition.svg?branch=master)](https://travis-ci.org/resin-io-modules/rendition)
+[![Dependency Status](https://img.shields.io/david/resin-io-modules/rendition.svg)](https://david-dm.org/resin-io-modules/rendition)
+[![npm](https://img.shields.io/npm/dm/rendition.svg?maxAge=2592000)](https://www.npmjs.com/package/rendition)
+[![GitHub last commit](https://img.shields.io/github/last-commit/resin-io-modules/rendition.svg)](https://github.com/resin-io-modules/rendition/commits/master)
 [![style: styled-components](https://img.shields.io/badge/style-%F0%9F%92%85%20styled--components-orange.svg?colorB=daa357&colorA=db748e)](https://github.com/styled-components/styled-components)
+[![Join the chat at https://gitter.im/resin-io-modules/rendition](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/resin-io-modules/rendition?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
```
pimterry: This gitter chat link doesn't work for me, I just get a 404. I think nowadays we mostly avoid having per-project gitters like this, since they're a hassle to keep an eye on and to maintain, and we push people towards the general resin one, or the forums.

README.md
```diff
@@ -2,7 +2,11 @@
 
 [![npm version](https://badge.fury.io/js/rendition.svg)](http://badge.fury.io/js/rendition)
 [![Build Status](https://travis-ci.org/resin-io-modules/rendition.svg?branch=master)](https://travis-ci.org/resin-io-modules/rendition)
+[![Dependency Status](https://img.shields.io/david/resin-io-modules/rendition.svg)](https://david-dm.org/resin-io-modules/rendition)
```
pimterry: This currently puts a big badge saying 'insecure' at the top of our readme, which doesn't seem ideal. Looks like our moment version needs a bump.

README.md
pimterry: This is quite a few badges, and when they end up on https://www.npmjs.com/package/rendition you'll have less space than you do on github, so I think you're breaking this out across multiple lines of badges there. Not a huge problem, but it might be good to trim these down a little.

README.md
LucianBuzzo: Yep, I noticed that myself, I'll follow up with another PR to bump the version ASAP

README.md
LucianBuzzo: Sure, I'd like to go ahead with the badges in this PR, and then we can trim them back at a later point

LucianBuzzo: Moving forwards rendition is going to be pushed as a separate entity/brand to resin, similar to etcher, which is why we it has its own gitter chatroom. I'm not sure why its getting a 404

LucianBuzzo: On second thoughts, I'm going to kill off the 'downloads' and 'last commit' badges

pimterry: Yep, I understand we're going to push it as it's own open-source project, but is gitter definitely our preferred route for user communication with our open source projects? My impression was that in general we were moving away from it everywhere we can, not setting up new ones.

Fine if we've discussed this already somewhere, but might be worth a #process around this if not. Could set up a Rendition section in the forum instead, for example, or... something else. Or maybe it's fine, I just don't know.

LucianBuzzo: I'll bring it up via #process

LucianBuzzo: For now I'm going to remove the gitter badge for now so we can move forward with this PR

PR #354, opened 2018-02-09
Set type and label attributes correctly when editing extra filters

### LINKED ISSUE #353: Set type and label attributes correctly when editing extra filters

LucianBuzzo:
Fixes #353

change-type: patch


src/components/Filters/index.tsx
```diff
@@ -243,24 +247,51 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
...
+		filterBase: FilterModel,
+		attribute: keyof FilterModel,
+		value: any,
+	) {
+		let filter = { ...filterBase };
 
-		if (attribute === 'name' && update.name !== value) {
-			update = this.setDefaultEditData(update, value);
+		if (attribute === 'name' && filter.name !== value) {
+			filter = this.setDefaultEditData(filter, value) as FilterModel;
```
pimterry: Preexisting nit: that method should really be called `getDefaultEditData`, right? It doesn't mutate anything afaict, it just gives you the data.

LucianBuzzo: Good point!

PR #355, opened 2018-02-09
Allow prefixes on BadgeSelect list

LucianBuzzo:
change-type: minor


src/components/BadgeSelect.tsx
```diff
@@ -91,6 +98,13 @@ class BadgeSelect extends React.Component<BadgeSelectProps, BadgeSelectState> {
 							<Text>{item}</Text>
 						</ButtonWrapper>
 					))}
+
+				{!!this.props.extraSuffix &&
+					this.props.extraSuffix.map(item => (
+						<ButtonWrapper key={item} onClick={() => this.setSelected(item)}>
+							<Text>{item}</Text>
+						</ButtonWrapper>
```
pimterry: Might be nice to have a light separator between prefix/suffix items?

LucianBuzzo: ack, good idea 👍

LucianBuzzo: @pimterry Here's what I've come up with
screen shot 2018-02-09 at 16 11 37

src/components/BadgeSelect.tsx
```diff
@@ -79,16 +80,38 @@ class BadgeSelect extends React.Component<BadgeSelectProps, BadgeSelectState> {
...
 				}
 			>
+				{!!this.props.extraPrefix &&
+					this.props.extraPrefix.map(item => (
+						<ButtonWrapper key={item} onClick={() => this.setSelected(item)}>
+							<Text fontSize={2}>{item}</Text>
+						</ButtonWrapper>
+					))}
+
+				{!!this.props.extraPrefix && (
```
pimterry: There's a subtle tiny sort-of bug here (and below): if I set `extraPrefix: []` then I'll get a divider, because `!![] === true`

LucianBuzzo: I'll create an issue and resolve it at a later date: https://github.com/resin-io-modules/rendition/issues/356

PR #359, opened 2018-02-13
Add Terminal component

### LINKED ISSUE #308: Add terminal component

LucianBuzzo:
Fixes #308

change-type: minor


src/components/Terminal.tsx
```diff
@@ -0,0 +1,250 @@
...
+			}
+
+			window.addEventListener('resize', this.resize);
+
+			this.resize();
+		}, 100);
+	}
+
+	componentWillUnmount() {
+		if (this.tty) {
```
pimterry: Why do we need this test? Looks like `this.tty` is set in the constructor, and never changed.

src/components/Terminal.tsx
```diff
@@ -0,0 +1,250 @@
...
+			window.removeEventListener('resize', this.resize);
+			this.tty.destroy();
+		}
+	}
+
+	destroy() {
+		window.removeEventListener('resize', this.resize);
+
+		if (!this.props.persistent) {
```
pimterry: `persistent` is always used because you've passed in a `ttyInstance` that you don't want destroyed, is that right?

Might be nice to remove this parameter, and just use `!this.props.ttyInstance` here instead - if you pass in an instance you're always responsible for cleaning it up, if you don't then it always gets cleaned up for you.

src/components/Terminal.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+	open() {
+		const rootElement = ReactDOM.findDOMNode(this);
+		const element = rootElement.querySelector(`#${this.mountElementId}`);
+		this.tty.open(element as HTMLElement);
+		this.tty.focus();
+	}
+
+	resize() {
+		if (this.tty) {
```
pimterry: Don't think you need this check either

src/components/Terminal.tsx
LucianBuzzo: yep agreed

PR #366, opened 2018-03-01
V4

LucianBuzzo:
Upgrades to react 16 and version bumps all dependencies.
Tooltips have been refactored to use an inline property.
Some components have been converted to TypeScript.


src/components/Badge.tsx
```diff
@@ -37,5 +38,5 @@ export default withTheme(
 				{text}
 			</BaseBadge>
 		);
-	},
-);
+	}),
+) as React.ComponentClass<BadgeProps>;
```
pimterry: Why is this cast required?

src/components/PineTypes/Boolean.js
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react'
 import Select from '../Select'
-import * as isBoolean from 'lodash/isBoolean'
-import * as isString from 'lodash/isString'
+const isBoolean = require('lodash/isBoolean')
+const isString = require('lodash/isString')
```
pimterry: Why do we need this?

src/hoc.tsx
```diff
@@ -5,6 +5,9 @@ import styled from 'styled-components';
 import { color, fontSize, space, width } from 'styled-system';
 import tag from 'tag-hoc';
 import blacklist from './blacklist';
+import { Tooltips } from './tooltips';
+
+const tt = new Tooltips();
```
pimterry: Nit: `tooltip` would be a little clearer

src/tooltips.ts
```diff
@@ -0,0 +1,244 @@
...
+							CLICK_TIMEOUT,
+						);
+					};
+					props.onClick = (e: Event) => showFn(e) || hideFn() || oldFn(e);
+				} else {
+					const oldMEFn = props.onMouseEnter || noop;
+					props.onMouseEnter = (e: Event) => showFn(e) || oldMEFn(e);
+
+					const oldMLFn = props.onMouseLeave || noop;
+					props.onMouseLeave = (e: Event) => this.hide() || oldMLFn(e);
```
pimterry: Here (and twice more above) there's bit of clever `() => a || b` to run `a` & then `b`. That works, but it's a little confusing and a little risky - if we accidentally return something truthy in any `a`, then `b` will never run.

Could these just be `() => { a(); b(); }` instead?

src/stories/Tooltip.js
```diff
@@ -1,37 +1,179 @@
...
+- Modal
+- Navbar
+- PineTypes
+- Provider
+- ProgressBar
+- Select
+- Table
+- Text
+- Textarea
+*/
```
pimterry: Leftover todo - are we doing this? If not now, it'd be nice to move it into an issue instead.

src/tooltips.ts
```diff
@@ -0,0 +1,244 @@
...
+	private tooltipElementArrow: HTMLDivElement;
+	private hideTimeout: number;
+	private initialised: boolean = false;
+
+	initialiseElements() {
+		if (this.initialised || !document || !document.body) {
+			return;
+		}
+
+		const tooltipElement = document.createElement('div');
```
pimterry: This is interesting. Why can't we use React + styled components to do all this tooltip stuff?

You probably don't want to put it in the downstream tree, but you could define a component here and separately call ReactDOM.render() with it to put it in the page yourself. I think that should still let you do this The React Way, and make a lot of this much simpler.

LucianBuzzo: Good idea

src/components/Badge.tsx
```diff
@@ -22,7 +23,7 @@ const BaseBadge = styled(Box)`
 `;
 
 export default withTheme(
-	({ small, text, theme, ...props }: ThemedBadgeProps) => {
+	hoc(({ small, text, theme, ...props }: ThemedBadgeProps) => {
```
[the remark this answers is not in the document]
LucianBuzzo: +1 for `asRendition`

src/components/Badge.tsx
LucianBuzzo: To keep me sane! I think we just started using this pattern to get consistent output between `compose` and `withTheme`. The styled components typings are not great..

src/components/PineTypes/Boolean.js
LucianBuzzo: I think jest was choking on the import syntax here

PR #367, opened 2018-03-08
Add Terminal methods to typings

LucianBuzzo:
change-type: patch


typings/rendition.d.ts
```diff
@@ -420,7 +420,14 @@ declare module 'rendition' {
 		config?: Xterm.ITerminalOptions;
 	}
 
-	class Terminal extends RenderableElementWithProps<TerminalProps, any> {}
+	class Terminal extends RenderableElementWithProps<TerminalProps, any> {
+		tty: Xterm.Terminal;
```
thgreasi: This is currently `private`!
If this is going to be `public` then I would also love to see the `readonly` keyword as well, so that nobody can mess with the components internals.

LucianBuzzo: In some instances we want the flexibility to mess with the internals a bit. For that reason I'll change `tty` to `readonly`. In the future I'll look at a complete abstraction of the internal mechanics.

LucianBuzzo: https://github.com/resin-io-modules/rendition/issues/368

PR #369, opened 2018-03-11
Fix bug where a persistent Terminal cannot be created

LucianBuzzo:
- Adds a story for persistent Terminals
- Updates to xterm.js 3.2.0
- Fixes bug where theme would become unset on persistent Terminals

change-type: patch


src/components/Terminal.tsx
```diff
@@ -200,17 +213,18 @@ class Terminal extends React.Component<ThemedTerminalProps, {}> {
...
+	// Explicitly calling '.destroy()' will always work, even with the 'persistent' property
 	destroy() {
 		window.removeEventListener('resize', this.resize);
 
-		// If the Terminal was not instantiated with an existing tty instance,
-		// destroy the instance when the component unmounts
-		if (!this.props.ttyInstance) {
-			this.tty.destroy();
-		}
+		this.tty.destroy();
```
thgreasi: Should we also set it to null or something?

src/stories/assets/repl.js
```diff
@@ -1,14 +1,19 @@
...
     this.replRoot = {}
+    this.gotLineCallback = gotLineCallback
+
+    this.open()
+  }
+
+  open () {
     this.iframe = document.createElement('iframe')
-    element.appendChild(this.iframe)
+    document.body.appendChild(this.iframe)
```
thgreasi: Should this have some styling to be hidden and not take space in the page?
image

src/components/Terminal.tsx
LucianBuzzo: @thgreasi Is this good practice for GC cycles? I would have thought that it would get ditched once the React component is removed

src/stories/assets/repl.js
LucianBuzzo: yep good idea

src/components/Terminal.tsx
thgreasi: Well, browsers can GC cyclic references, but this would make it a bit easier for them.

Issue #371, opened 2018-03-15
Should have an unstyled click-to-copy element


thgreasi:
See: https://app.zeplin.io/project/5a3a81ffa3beb747c481b1b0

PR #372, opened 2018-03-15
Add generic TextWithCopy component

thgreasi:
Needed for https://github.com/resin-io/resin-ui/issues/1810
This is a bit more generic than CodeWithCopy.
Didn't unify the two, since in the UI we are using the CSS classes if the nested elements and that would be a breaking change. Should better plan to do that in the next major release though.
Suggestions for a better name are welcome but I couldn't make my mind on that matter. Maybe ElementWithCopy, ClickToCopy or something.

Resolves: #371 
See: https://app.zeplin.io/project/5a3a81ffa3beb747c481b1b0
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/W-uoeMYre-N9F8IRrkCW9pL7nHn
Change-Type: minor


typings/rendition.d.ts
```diff
@@ -450,6 +450,23 @@ declare module 'rendition' {
 
 	class Text extends RenderableElementWithProps<TextProps, any> {}
 
+	namespace Text {
```
LucianBuzzo: Why do we need this? If its required for some reason can you add a comment to the typing explaining its purpose please.

thgreasi: Will do.
Without this the typings do not allow you to do `<Text.span>`. This is almost exactly what we did for `<Heading.h1,2,3,4...>` (even though the typings do not allow the plain `<Heading>` there and I'm not sure whether that's on purpose).

PR #376, opened 2018-03-28
Rebuild Filters component and SchemaSieve to use JSON schema

LucianBuzzo:


__tests__/SchemaSieve.spec.js
```diff
@@ -0,0 +1,1547 @@
...
+          is_less_than: 'less than'
+        }
+      },
+      'Key Value Pair': {
+        v4Type: 'object',
+        testValue: { tag_name: 'test', tag_value: 'test' },
+        operators: {
+          is: 'is',
+          is_not: 'is_not',
+          key_is: 'key is',
```
thgreasi: Would it make sense to have the value here be `name is`?

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,157 +1,516 @@
...
+	return pickBy(collection, m => every(validators, v => v(m)));
+};
 
-			if (operator! in filterTests[type].rules) {
-				const result = filterTests[type].rules[operator!];
-				return isPlainObject(result)
-					? result.test(target, value)
-					: result(target, value);
-			}
+export const getDataModel = (schema?: JSONSchema6) => {
```
thgreasi: Would it make sense to move this to `DataTypes/index.tsx` so that it's closer to the objects that it manages/exposes? In that case the `DataTypes` folder/"namespace" would have the full responsibility of resolving and providing, eh, the DataTypes... and this file will be uncoupled from each specific DataType available.

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,157 +1,516 @@
...
+		return DataTypes.boolean;
+	}
+	if (type === 'number') {
+		return DataTypes.number;
+	}
+
+	return null;
+};
+
+export const createFullTextSearchFilter = (
```
thgreasi: Would it make sense to move the FullTextSearch related methods to a different file?
This is yet an other implementation of a filter type, so it might not have to be next to the sieve methods, even though re-exporting is required.

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,157 +1,516 @@
...
 
-	filterArray(collection: any[], input: FilterRule | FilterRule[]) {
-		return collection.filter(item => this.test(item, input));
+	return map(model.operators, (value, key) => ({
+		slug: key,
+		label: value.getLabel(subSchema),
+	}));
+};
+
+const convertToBooleanSignature = (
```
thgreasi: Could we move all the migration/legacy/v3 related pieces in a separate migration module?

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,157 +1,516 @@
...
+const convertToStringSignature = (
+	legacyFilter: v3.BaseFilter,
+): FilterSignature => {
+	const signature = {
+		field: legacyFilter.name,
+		value: legacyFilter.value,
+		operator: '',
+	};
+
+	switch (legacyFilter.operator) {
```
thgreasi: It seems that this switch could be replaced by a mapping object and have something like
```
signature.operator = legacyFilterMap[legacyFilter.operator] || legacyFilter.operator || '';
```
Probably also applies to other methods.

src/components/Filters/index.tsx
```diff
@@ -81,291 +89,217 @@ const FilterWrapper = styled(Box)`
...
+		const item = currentEdit[index];
+		currentEdit.splice(index, 1, { ...item, operator });
+		this.setState({
+			edit: currentEdit,
+		});
 	}
 
-	showEditModal(rule: FilterRule) {
+	setEditValue(value: string, index: number) {
+		console.log('setting edit value', value);
```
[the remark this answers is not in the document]
thgreasi: These two look exactly almost identical.
How about combining into one `setEditProps(index: number, props: { operator?: string; value?: string; })` (or typed as the two objects used in here with `|`)?

src/components/Filters/index.tsx
```diff
@@ -81,291 +89,217 @@ const FilterWrapper = styled(Box)`
...
+			const matchIndex = findIndex(currentFilters, { $id });
+			currentFilters.splice(matchIndex, 1, filter);
+			filters = currentFilters.slice();
 		} else {
-			update.value = '';
+			filters = currentFilters.concat(filter);
 		}
 
-		return update;
+		console.log('FILTERS', filters);
```
thgreasi: `console.log` :police_car:

src/components/Filters/index.tsx
```diff
@@ -81,291 +89,217 @@ const FilterWrapper = styled(Box)`
 	position: relative;
 `;
 
+interface EditModel {
```
thgreasi: This is identical to `FilterSignature` and we are cross-assigning vars with casting. Should we dedup them or make the one extend the other?

src/components/Filters/index.tsx
```diff
@@ -81,291 +89,217 @@ const FilterWrapper = styled(Box)`
...
-			set(filter, attribute, value);
-		}
+	editFilter(filter: JSONSchema6) {
+		const { schema } = this.props;
 
-		return filter;
+		const signatures = SchemaSieve.decodeFilter(schema, filter);
+
+		this.setState({
+			edit: signatures as EditModel[],
```
thgreasi: Can we move that cast inside `decodeFilter`?

src/components/Filters/index.tsx
```diff
@@ -81,291 +89,217 @@ const FilterWrapper = styled(Box)`
...
 
-	handleFilterFormChange(value: string, attribute: keyof FilterModel) {
-		const update = this.updateFilter(
-			this.state.edit as FilterModel,
-			attribute,
-			value,
+	removeFilter({ $id }: JSONSchema6) {
+		this.setState(
+			prevState => ({
+				filters: prevState.filters.filter(f => f.$id !== $id),
```
thgreasi: How about [`_.reject`](http://devdocs.io/lodash~4/index#reject)?
(filtering the filters!!!)

src/components/Filters/index.tsx
```diff
@@ -81,291 +89,217 @@ const FilterWrapper = styled(Box)`
...
-
-		if (store) {
-			store.data = store.data.filter(v => v.id !== view.id);
-
-			this.props.setViews(views);
-		}
+	deleteView({ id }: FiltersView) {
+		this.setState(
+			prevState => ({
+				views: prevState.views.filter(v => v.id !== id),
```
thgreasi: > How about [`_.reject`](http://devdocs.io/lodash~4/index#reject)?

src/stories/__Tools_JSONSchema.js
```diff
@@ -0,0 +1,100 @@
...
+const ajvKeywords = require('ajv-keywords')
+const metaSchema6 = require('ajv/lib/refs/json-schema-draft-06.json')
+
+const ajv = new Ajv()
+ajvKeywords(ajv)
+ajv.addMetaSchema(metaSchema6)
+
+window.ajv = ajv
+
+class Validator extends React.Component {
```
thgreasi: Woohoo!

typings/rendition.d.ts
```diff
@@ -64,54 +65,28 @@ declare module 'rendition' {
...
+	interface DataTypeModel {
+		operators: {
+			[key: string]: {
+				getLabel: (schema: JSONSchema6) => string;
+			};
 		};
+		decodeFilter(filter: JSONSchema6): null | FilterSignature;
+		createFilter(
+			field: string,
+			o: string,
```
thgreasi: I would love to see a better name here.

typings/rendition.d.ts
```diff
@@ -508,4 +496,65 @@ declare module 'rendition' {
 	}
 
 	const Theme: Theme;
+
+	namespace v3 {
```
thgreasi: Could we move this to an other file and re-export as part of the namespace here?
Or maybe we can also declare `declare module 'rendition' { namespace V3 { /*...*/ } }` in the other file and just import it here.

LucianBuzzo:
Thanks for the review @thgreasi !

package-lock.json
```diff
@@ -0,0 +1,15352 @@
+{
+  "name": "rendition",
```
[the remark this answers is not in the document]
LucianBuzzo: Yep it was intentional

__tests__/SchemaSieve.spec.js
```diff
@@ -0,0 +1,1547 @@
...
+        },
+        'Entry 5': {
+          foo: 'bar'
+        }
+      }
+
+      it('should correctly test values using the "is" operator', function () {
+        const filter = sieve.createFilter(schema, [{
+          field: 'Tag',
+          operator: 'is',
```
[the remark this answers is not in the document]
LucianBuzzo: Good idea, I'll add one

__tests__/SchemaSieve.spec.js
```diff
@@ -0,0 +1,1547 @@
...
+
+      expect(sieve.getDataModel(schema)).toBe(null)
+    })
+
+    it('should return a model for string types', () => {
+      const schema = {
+        type: 'string'
+      }
+
+      expect(sieve.getDataModel(schema)).toContainAllKeys(dataModelKeys)
```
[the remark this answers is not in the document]
LucianBuzzo: Do you think we should check against the actual expected model?

__tests__/SchemaSieve.spec.js
```diff
@@ -0,0 +1,1547 @@
...
+          expect(
+            stripId(sieve.migrateLegacyFilter(schema, legacyFilter))
+          ).toEqual(
+            stripId(filter)
+          )
+        })
+      })
+    })
+
+    it(`should convert legacy 'Full Text Search' filters`, () => {
```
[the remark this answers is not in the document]
LucianBuzzo: Yep, but I will add a test for this

src/components/DataTypes/boolean.tsx
```diff
@@ -0,0 +1,84 @@
...
+	is: {
+		getLabel: (_s: JSONSchema6) => 'is',
+	},
+};
+
+type OperatorSlug = keyof typeof operators;
+
+export const decodeFilter = (filter: any): FilterSignature | null => {
+	const field = Object.keys(filter.properties!).shift()!;
+	const operator: OperatorSlug = filter.title;
```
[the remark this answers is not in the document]
LucianBuzzo: I thought a lot about this and I settled on putting the operator as the `title` and the human readable description as the `description`. I'm stretching the spec a bit here, but this seemed to make the most sense

__tests__/SchemaSieve.spec.js
```diff
@@ -0,0 +1,1547 @@
...
+    Object.keys(schema.properties).forEach(type => {
+      it(`should return an array of operators for the '${type}' type`, () =>
+        sieve.getOperators(schema, type).forEach((operator) =>
+          expect(operator).toContainAllKeys(['slug', 'label'])
+        )
+      )
+    })
+  })
+
+  describe('.migrateLegacyFilter()', () => {
```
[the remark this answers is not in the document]
LucianBuzzo: Yep, good idea, I've moved them to separate file now

__tests__/SchemaSieve.spec.js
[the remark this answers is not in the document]
thgreasi: I thought so, but if you think that there is no added benefit then plz just ignore this.

src/components/DataTypes/boolean.tsx
```diff
@@ -0,0 +1,84 @@
...
+	return {
+		field,
+		operator,
+		value,
+	};
+};
+
+export const createFilter = (
+	field: string,
+	o: string,
```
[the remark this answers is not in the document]
LucianBuzzo: Good idea, I've updated the other type files as well

src/components/DataTypes/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as booleanType from './boolean';
```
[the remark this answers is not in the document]
LucianBuzzo: yes definitely!

src/components/Filters/Summary.tsx
```diff
@@ -85,29 +65,20 @@ class FilterSummary extends React.Component<
 		});
 	}
 
-	render() {
-		if (every(this.props.rules, r => !isValidRule(r))) {
-			return null;
-		}
+	setExistingId(e: Event) {
```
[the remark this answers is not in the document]
LucianBuzzo: good catch

src/components/Filters/Summary.tsx
```diff
@@ -147,24 +118,31 @@ class FilterSummary extends React.Component<
 						</form>
 					</Modal>
 				)}
-				<Flex wrap>
-					{this.props.rules.map(rule => {
-						if (!isValidRule(rule)) {
```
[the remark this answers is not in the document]
LucianBuzzo: I'm going to handle it in the filterdescription component

src/components/Filters/ViewsMenu.tsx
```diff
@@ -136,60 +133,47 @@ class ViewsMenu extends React.Component<ViewsMenuProps, ViewsMenuState> {
...
-													</ViewListItem>
-												);
-											})}
-										</UnstyledList>
-									</Box>
-								);
-							})}
+							map(groupedViews, (views: FiltersView[], scope) => (
+								<Box key={scope}>
+									{(!!scope && scope !== 'null') && (
```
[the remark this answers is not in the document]
LucianBuzzo: urgh, completely unnecessary too! I'll tidy this up

src/components/Filters/index.tsx
```diff
@@ -1,41 +1,49 @@
...
+import Summary from './Summary';
+
+interface FilterInput {
+	schema: JSONSchema6;
+	value: any;
+	operator: string;
+	onUpdate: (value: any) => void;
+}
+
+const FilterInput = (props: any) => {
```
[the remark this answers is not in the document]
LucianBuzzo: yep done

src/components/Filters/index.tsx
```diff
@@ -81,291 +89,217 @@ const FilterWrapper = styled(Box)`
...
-		this.props.setRules(this.filterInvalidRules(rules));
 	}
 
-	generateFreshEdit() {
-		if (!this.props.schema) {
-			return {};
+	getCleanEditModel(field?: string) {
+		const { schema } = this.props;
+		if (!field) {
+			field = Object.keys(schema.properties as JSONSchema6).shift()!;
```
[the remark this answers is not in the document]
LucianBuzzo: good catch

src/components/Filters/index.tsx
LucianBuzzo: I always forget about reject :) it's so neat

typings/rendition.d.ts
LucianBuzzo: I prefer to keep it here for simplicities sake

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,157 +1,173 @@
...
-			return this.filterArray(collection, input);
-		}
-		if (isPlainObject(collection)) {
-			return this.filterObject(collection, input) as Partial<T>;
-		}
+	return pickBy(collection, m => every(validators, v => v(m)));
+};
 
-		throw new Error('collection argument must be either object or array.');
+export const getDataModel = (schema?: JSONSchema6) => {
```
thgreasi: > Would it make sense to move this to `DataTypes/index.tsx` so that it's closer to the objects that it manages/exposes? In that case the `DataTypes` folder/"namespace" would have the full responsibility of resolving and providing, eh, the DataTypes... and this file will be uncoupled from each specific DataType available.

src/stories/Filters.js
```diff
@@ -35,71 +33,68 @@ const Tag = ({ tag }) => (
...
+    Name: {
+      title: 'Pokemon Name',
+      type: 'string'
+    },
+    Description: {
+      title: 'Pokemon Name',
+      type: 'string'
+    },
+    Abilities: {
+      title: 'Pokemon Name',
```
thgreasi: I think that these titles should be changed.
image

src/components/Filters/index.tsx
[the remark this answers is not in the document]
thgreasi: > These two look exactly almost identical.
How about combining into one `setEditProps(index: number, props: { operator?: string; value?: string; })` (or typed as the two objects used in here with `|`)?

^ ?

package.json
```diff
@@ -58,7 +58,7 @@
     "ts-loader": "^3.0.2",
     "ts-node": "^3.3.0",
     "tslint": "^5.8.0",
-    "typescript": "^2.6.2"
+    "typescript": "^2.8.1"
```
thgreasi: I guess we I should resume working on the respective UI PR as well :) :+1:

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,157 +1,173 @@
...
+		const subSchema = schema.properties![field] as JSONSchema6;
+		const model = getDataModel(subSchema);
+
+		if (!model || !model.operators.hasOwnProperty(operator)) {
+			return {};
 		}
 
-		return filterTests[type].validate(value);
-	}
+		return (model.createFilter as (...args: any[]) => JSONSchema6)(
```
thgreasi: Could we define a base signature to cast `createFilter`?
```
(
field: string,
operator: string,
value: any,
subSchema: JSONSchema6,
) => JSONSchema6
```
This way we can reduce the amounts of `any` used and have a place that actually checks that all `createFilter` implementations are compatible.

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,157 +1,173 @@
...
+			field,
+			operator as any,
```
thgreasi: We can get rid of this, even if we change the `createFilter` cast right above.

package-lock.json
```diff
@@ -0,0 +1,15352 @@
...
+    },
+    "babel-plugin-transform-es2015-modules-amd": {
+      "version": "6.24.1",
+      "resolved": "https://registry.npmjs.org/babel-plugin-transform-es2015-modules-amd/-/babel-plugin-transform-es2015-modules-amd-6.24.1.tgz",
+      "integrity": "sha1-Oz5UAXI5hC1tGcMBHEvS8AoA0VQ=",
+      "dev": true,
+      "requires": {
+        "babel-plugin-transform-es2015-modules-commonjs": "6.26.0",
+        "babel-runtime": "6.26.0",
+
```
thgreasi: !!! the lock still points to ts 2.8.1 !!!

LucianBuzzo:
@thgreasi I can't reply to your quoted comments :(

thgreasi:
@LucianBuzzo wait, what? Why? You mean that there is no reply button on GH or that my comments are a bit meaningless?

LucianBuzzo:
@thgreasi Theres no reply button!
screen shot 2018-04-03 at 14 24 53

src/components/DataTypes/object.tsx
```diff
@@ -0,0 +1,385 @@
...
+export const createFilter = (
+	field: string,
+	operator: OperatorSlug,
+	value: any,
+	schema: JSONSchema6,
+): JSONSchema6 => {
+	const { title } = schema;
+	const base: ObjectFilter = {
+		title: operator,
+		description: `${title || field} ${operators[operator].getLabel(
```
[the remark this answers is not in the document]
LucianBuzzo: I prefer to KISS then DRY

src/components/Filters/FilterDescription.tsx
```diff
@@ -19,53 +15,22 @@ const ButtonWrapper = styled.button`
...
-			<em>{props.rule.value}</em>
-		)}
-	</Flex>
+const FilterDescriptionInner = (props: { filter: JSONSchema6 }) => (
+	<Box>
+		{!!props.filter.anyOf &&
+			props.filter.anyOf.map((f, i) => (
+				<React.Fragment key={i}>
+					{i > 0 && <em> or </em>}
+					<span>{f.description}</span>
```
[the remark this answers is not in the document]
LucianBuzzo: 👍

LucianBuzzo:
>Is this used anywhere? After a quick search I didn't find any place that is referenced, only found that it's defined in here and for the full text search filter.
The `required` declaration is required for the JSON schema to validate correctly, its actually part of the filtering logic.

LucianBuzzo:
>NIT: Since this is the only method that returns JSX notation(/element factory/however else it's called anyway), would it make more sense to name these as render?

It's very likely that I will expand these modules to be able to `Display` data as well, so keeping this named `Edit` is helpful

LucianBuzzo:
>These two look exactly almost identical.
How about combining into one setEditProps(index: number, props: { operator?: string; value?: string; }) (or typed as the two objects used in here with |)?

I'd really prefer to keep these separate, for the sake of simplicity

thgreasi:
Woohoo!

Issue #378, opened 2018-04-04
Anchors do not work inside the Modal component


thgreasi:
That's probablt because of `utils.stopEvent()` also doing a `preventDefault()` before `stopPropagation`
See: https://github.com/resin-io-modules/rendition/blob/master/src/components/Modal.tsx#L119
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/QxsM0RLkhpgoA2-O_Qd7MZPdrVT

PR #379, opened 2018-04-04
Modal: Fix anchors inside modals

thgreasi:
Tried this in the 2FA & device tag nested modals.

Resolves: #378 
Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #378


src/components/Modal.tsx
```diff
@@ -116,7 +115,10 @@ class Modal extends React.Component<ModalProps, any> {
 			>
 				<ModalBackdrop z={8888} bg="rgba(0,0,0,0.4)" top right bottom left />
 				<ModalSizer>
-					<ModalPanel w={w || width || DEFAULT_MODAL_WIDTH} onClick={stopEvent}>
+					<ModalPanel
+						w={w || width || DEFAULT_MODAL_WIDTH}
+						onClick={(e: Event) => e.stopPropagation()}
```
pimterry: It'd be nice to avoid arrow functions in render. Some background here: https://medium.freecodecamp.org/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36.

Not a big deal, but since we might want to do this in lots of places, imo it'd be slightly better to avoid this and simplify things with a single `stopPropagation` function in util, and just use that everywhere.

Issue #283, opened 2018-01-05
Create a Form component that can generate forms from a json schema


LucianBuzzo:
We can use `react-json-schema` and utilise the ability to override the default fields and widgets, replacing them with rendition fields https://github.com/mozilla-services/react-jsonschema-form#customizing-the-default-fields-and-widgets

LucianBuzzo:
This means we need to provide overrides for the following default fields:

- [ ] ArrayField
- [ ] BooleanField
- [ ] DescriptionField
- [ ] NumberField
- [ ] ObjectField
- [ ] SchemaField
- [ ] StringField
- [ ] TitleField
- [ ] UnsupportedField

And the following default widgets:

- [ ] BaseInput
- [ ] PasswordWidget
- [ ] RadioWidget
- [ ] UpDownWidget
- [ ] RangeWidget
- [ ] SelectWidget
- [ ] TextWidget
- [ ] DateWidget
- [ ] DateTimeWidget
- [ ] AltDateWidget
- [ ] AltDateTimeWidget
- [ ] EmailWidget
- [ ] URLWidget
- [ ] TextareaWidget
- [ ] HiddenWidget
- [ ] ColorWidget
- [ ] FileWidget
- [ ] CheckboxWidget
- [ ] CheckboxesWidget

PR #386, opened 2018-04-16
[Beta] Add Form component for generating forms from JSON schemas

LucianBuzzo:
Add Form component for generating forms from JSON schemas

Connects to #283

change-type: minor

screen shot 2018-04-20 at 12 13 10


thgreasi:
Is it expected that the storyboard errors?
```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports. Check the render method of `FormDemo`.
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `FormDemo`.
    at invariant (http://localhost:6006/static/preview.bundle.js:8833:15)
    at throwOnInvalidElementType (http://localhost:6006/static/preview.bundle.js:42646:3)
    at createFiberFromElement (http://localhost:6006/static/preview.bundle.js:42617:13)
    at reconcileSingleElement (http://localhost:6006/static/preview.bundle.js:45298:23)
    at reconcileChildFibers (http://localhost:6006/static/preview.bundle.js:45354:35)
    at reconcileChildrenAtExpirationTime (http://localhost:6006/static/preview.bundle.js:45484:30)
    at reconcileChildren (http://localhost:6006/static/preview.bundle.js:45475:5)
    at updateHostComponent (http://localhost:6006/static/preview.bundle.js:45783:5)
    at beginWork (http://localhost:6006/static/preview.bundle.js:46226:16)
    at performUnitOfWork (http://localhost:6006/static/preview.bundle.js:49054:16)
```

src/components/Form/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+		console.log('FORMDATA', formData);
+
+		this.setState({ formData }, () => {
+			if (this.props.onChange) {
+				this.props.onChange(formData);
+			}
+		});
+	}
+
+	submit() {
```
thgreasi: NIT: Depending on how render heavy is the `<Form>` component, you might want to consider changing this to `submit = () => {/*...*/}` (instead of using `.bind()`) and the respective property can then be `onSubmit={this.submit}`, which will cause less re-renders due to property change.

src/components/Form/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+		this.change = this.change.bind(this);
+	}
+
+	change(target: string, value: any) {
+		console.log(target, value);
+		const { formData } = this.state;
+		set(formData, target, value);
+
+		console.log('FORMDATA', formData);
```
thgreasi: `console.log` :police_car:

src/components/Form/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+		this.state = {
+			formData: {},
+		};
+
+		this.change = this.change.bind(this);
+	}
+
+	change(target: string, value: any) {
+		console.log(target, value);
```
thgreasi: `console.log` :police_car:

src/components/Form/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	submit() {
+		if (this.props.onSubmit) {
+			this.props.onSubmit(this.state.formData);
+		}
+	}
+
+	render() {
+		const { schema, onChange, onSubmit, value, ...props } = this.props;
```
thgreasi: I think that `onSubmit` isn't used.

LucianBuzzo: onSubmit isn't used, but its a neat way to stop the prop getting passed to `FormWrapper`

thgreasi:
Something really weird is happening:
peek 2018-04-20 19-32
:alien: :satellite: ?

LucianBuzzo:
@thgreasi Looks like some kind of `setState` race condition, I'll investigate further

LucianBuzzo:
@thgreasi The issue is resolved :+1:

README.md
```diff
@@ -36,6 +36,14 @@ Wrap your application in the `<Provider>` component so that child components can
 
 All components support [styled-system][styled-system] attributes, allowing you to use `fontSize`, `color`, `px` etc.
 
+## Unstable/Beta components
+
+Some components in the [storybook](https://resin-io-modules.github.io/rendition)
+are marked with a `[beta]` label. These components are experimental and still
+under development. They should not be used in a production application.
```
thgreasi: If-you-are-in-the-mood: could also mention that their API might change w/o notice.

__tests__/components/Form.spec.js
```diff
@@ -0,0 +1,153 @@
...
+        <Provider>
+          <Form schema={schema} onChange={callback} />
+        </Provider>
+      )
+
+      const input = component.find('input')
+      input.simulate('change', { target: { value } })
+      component.update()
+
+      setTimeout(() => {
```
thgreasi: If-you-are-in-the-mood: maybe we could use a since `waitForComponentUpdate` (or something) that could
* call `component.update()`
* call `setTimeout` and return a promise (or use Bluebird's `Promise.delay()`)

Additionally this could let us drop the `done()` calls (since we will be returning the promise instead form within the `it()`) and maybe in the future even let use `await` :)

src/stories/Form.js
```diff
@@ -0,0 +1,110 @@
...
+          <Txt monospace>
+            <pre>{JSON.stringify(this.state.formData, null, 4)}</pre>
+          </Txt>
+        </Box>
+      </Flex>
+    )
+  }
+}
+
+storiesOf('[beta] Form', module)
```
thgreasi: If-you-are-in-the-mood: How about moving the `[beta]` past to the end so that the it doesn't look like a mis-ordering in the stories menu
image

Or someone could argue about a whole different ordering for all the beta components (eg last), but probably doesn't worth it.

src/unstable/components/Form/FieldTemplate.tsx
```diff
@@ -0,0 +1,56 @@
...
+	}
+	return (
+		<LabelElement className="control-label" htmlFor={id}>
+			{label}
+			{required && <span className="required">{REQUIRED_FIELD_SYMBOL}</span>}
+		</LabelElement>
+	);
+};
+
+const FieldTemplate = (props: any) => {
```
thgreasi: How about typing those two `any`s?
Seems straightforward.

src/unstable/components/Form/FieldTemplate.tsx
```diff
@@ -0,0 +1,56 @@
...
+		children,
+		errors,
+		help,
+		description,
+		hidden = false,
+		required = false,
+		displayLabel = true,
+	} = props;
+	if (hidden) {
+		return children;
```
thgreasi: I would expect that `hidden` would mean `return null` :confused: 
Would using something like `noLabel` make more sense (or maybe I'm missing something).

src/stories/Form.js
```diff
@@ -0,0 +1,110 @@
...
+      title: 'National Pokedex Number',
+      type: 'number'
+    },
+    caught: {
+      type: 'boolean'
+    },
+    first_seen: {
+      title: 'First seen',
+      type: 'string',
+      format: 'date-time'
```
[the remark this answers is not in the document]
LucianBuzzo: I'll address this in a future PR once this is merged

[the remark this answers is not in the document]
thgreasi: I guess that's fine, since the component is beta :)
Just open an issue or leave a comment so that we don't forget this.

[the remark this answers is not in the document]
LucianBuzzo: I'll go back through and open issues once this PR is merged

src/unstable/components/Form/FieldTemplate.tsx
LucianBuzzo: These are the props passed to the component by `react-jsonschema-form`, so we can't change them. I actually lifted this component from the module's source  https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/fields/SchemaField.js#L97

src/unstable/components/Form/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+	componentWillReceiveProps(nextProps: FormProps) {
+		if (!isEqual(this.props.value, nextProps.value)) {
+			this.setState({
+				value: nextProps.value,
+			});
+		}
+	}
+
+	change = (data: any) => {
```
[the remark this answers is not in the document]
LucianBuzzo: I can type it by using this https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-jsonschema-form/index.d.ts#L171

typings/rendition.d.ts
```diff
@@ -595,3 +595,22 @@ declare module 'rendition' {
...
+declare module 'rendition/dist/unstable' {
+	import { JSONSchema6 } from 'json-schema';
+	import { Component } from 'react';
+	import { UiSchema } from 'react-jsonschema-form';
+	import { BoxProps, RenderableElementWithProps } from 'rendition';
+
+	interface FormProps extends BoxProps {
+		schema: JSONSchema6;
+		submitButtonText?: string;
+		hideSubmitButton?: boolean;
```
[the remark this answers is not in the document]
LucianBuzzo: I gave this some though and I prefer this being a bit verbose.

PR #391, opened 2018-04-16
Allow filtering on nested data structures by flattening the schema

### LINKED ISSUE #388: <Filters> component should 'flatten' schema
If you pass the following schema to the `<Filters>` component:
```
{
  "additionalProperties": true,
  "properties": {
    "data": {
      "properties": {
        "actor": {
          "format": "uuid",
          "type": "string"
        },
        "payload": {
          "additionalProperties": true,
          "properties": {
            "message": {
              "type": "string"
            }
          },
          "required": [
            "message"
          ],
          "type": "object"
        },
        "target": {
          "format": "uuid",
          "type": "string"
        },
        "timestamp": {
          "format": "date-time",
          "type": "string"
        }
      "type": "object"
    }
  },
  "type": "object"
}
```

I would expect to be able to filter on:
- `actor`
- `message`
- `target`
- `timestamp`

LucianBuzzo:
Fixes #388

change-type: minor


src/components/Filters/SchemaSieve.ts
```diff
@@ -142,3 +149,125 @@ export const getOperators = (schema: JSONSchema6, field: string) => {
...
+		}
+
+		flattenSchema(value, delimiter, newKey, acc);
+	});
+
+	return accumulator;
+};
+
+// Restores a schema that has been flattened with `flattenSchema()`
+export const hydrateSchema = (
```
pimterry: Hydrate doesn't sound like the opposite of flatten, it sounds like it's deserializing data. `unflatten`?

src/components/Filters/index.tsx
```diff
@@ -104,40 +105,55 @@ interface FiltersState {
...
 			editingFilter: null,
 			filters: filters || [],
 			views: views || [],
+			schema: flatSchema,
+			edit: [this.getCleanEditModel(null, flatSchema)],
 		};
+	}
 
-		// Clean exsting rules on load
+	componentDidUpdate(prevProps: FiltersProps) {
```
pimterry: You shouldn't be updating this in `componentDidUpdate`. That'll get called _after_ the props changed & caused a rerender, so you're going to end up doing two renders here unnecessarily, and this logic will run useless for lots of other updates.

Instead you want `componentWillReceiveProps`, or even better `getDerivedStateFromProps` (if you're happy targetting only super recent React versions).

src/components/Filters/index.tsx
```diff
@@ -167,9 +183,20 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
 		});
 	}
 
+	emitFilterUpdate() {
+		if (!this.props.onFiltersUpdate) {
+			return;
+		}
+
+		console.log(this.state.filters);
```
pimterry: Console.log

src/components/Filters/SchemaSieve.ts
```diff
@@ -142,3 +149,125 @@ export const getOperators = (schema: JSONSchema6, field: string) => {
 		label: value.getLabel(subSchema),
 	}));
 };
+
+// Reduces a multi level schema to a single level
+export const flattenSchema = (
+	schema: JSONSchema6,
+	delimiter: string = DELIMITER,
+	parentKey: string = '',
+	accumulator?: JSONSchema6,
```
pimterry: This function works in two ways:

* It might be a pure function that carefully clones & returns data
* It might be a side-effect function that returns data through this argument, and also directly

That's fairly surprising, and makes some of this hard to follow. It'll get way clearer if you split this into two functions:

* One that takes an accumulator, updates it with extra data from a schema, and is potentially recursive
* One that just sets up an accumulator, calls the first, and then returns the accumulator itself

src/components/Filters/index.tsx
```diff
@@ -104,40 +105,55 @@ interface FiltersState {
...
+		if (!isEqual(prevProps.schema, this.props.schema)) {
+			this.setState({
+				schema: SchemaSieve.flattenSchema(this.props.schema),
+			});
+		}
 	}
 
-	getCleanEditModel(field?: string) {
-		const { schema } = this.props;
+	getCleanEditModel(field?: string | null, schema?: JSONSchema6) {
```
pimterry: Why does this and `getOperators` both now take a schema as a parameter? Doesn't seem to be used here at all. Is this for something else?

src/components/Filters/SchemaSieve.ts
LucianBuzzo: Good idea

src/components/Filters/SchemaSieve.ts
LucianBuzzo: Good point, I'll rework this and make it clearer

src/components/Filters/index.tsx
LucianBuzzo: This is purely to make everything easier when setting up the state in `constructor`. As `getCleanEditModel` and `getOperators` both now use `state` internally, they can't be called before `this.state` has been instantiated. I get around this by allowing you to specify the schema.

src/components/Filters/index.tsx
```diff
@@ -167,9 +183,20 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
...
 	}
 
+	emitFilterUpdate() {
+		if (!this.props.onFiltersUpdate) {
+			return;
+		}
+
+		console.log(this.state.filters);
+		this.props.onFiltersUpdate(
+			this.state.filters.map(filter => SchemaSieve.hydrateSchema(filter)),
```
thgreasi: Please also update this one

src/components/Filters/SchemaSieve.ts
```diff
@@ -142,3 +149,136 @@ export const getOperators = (schema: JSONSchema6, field: string) => {
...
+		}
+	}
+
+	if (schema.anyOf) {
+		accumulator.anyOf = schema.anyOf.map(subSchema =>
+			flattenAccumulator(subSchema, delimiter, parentKey),
+		);
+	}
+
+	forEach(schema.properties, (value: JSONSchema6, key) => {
```
thgreasi: How about using `reduce` instead of `forEach` and remove all `!` from the lambda?
PS: `reduce` is actually semantically more appropriate.

thgreasi: The value is typed as `boolean | JSONSchema6`, would it make sense remove the type from the arg and check the actual type in the callback? It feels a bit safer.

src/components/Filters/SchemaSieve.ts
```diff
@@ -142,3 +149,136 @@ export const getOperators = (schema: JSONSchema6, field: string) => {
...
+	if (!unflattenedSchema.properties) {
+		return unflattenedSchema;
+	}
+
+	unflattenedSchema.properties = {};
+
+	forEach(schema.properties, (value: JSONSchema6, field) => {
+		// If the property hasn't been flattened, we can skip it
+		if (!startsWith(field, DELIMITER)) {
+			return (unflattenedSchema.properties![field] = value);
```
thgreasi: Please change these `return` statement to be on a separate line and not return anything.

src/components/Filters/SchemaSieve.ts
```diff
@@ -142,3 +149,136 @@ export const getOperators = (schema: JSONSchema6, field: string) => {
...
+		let head = unflattenedSchema;
+
+		const keys = trimStart(field, delimiter).split(delimiter);
+
+		while (keys.length) {
+			const key = keys.shift()!;
+
+			// Don't overwrite an existing entry
+			if (!head.properties![key]) {
+				head.properties![key!] = keys.length
```
thgreasi: I think that the `!` is not needed in `key!`.

src/components/Filters/SchemaSieve.ts
```diff
@@ -142,3 +149,136 @@ export const getOperators = (schema: JSONSchema6, field: string) => {
...
+		const requiredKeys = unflattenedSchema.required;
+
+		unflattenedSchema.required = [];
+
+		requiredKeys.forEach(requiredField => {
+			const keys = trimStart(requiredField, delimiter).split(delimiter);
+
+			let head = unflattenedSchema;
+
+			while (keys.length) {
```
thgreasi: This could be a `for...of` and avoid the table manipulation, since we are not using `keys` inside the loop, but if you just prefer to keep it symmetric with the previous piece of code then :+1:

src/components/Filters/SchemaSieve.ts
```diff
@@ -142,3 +149,136 @@ export const getOperators = (schema: JSONSchema6, field: string) => {
...
+			if (!head.properties![key]) {
+				head.properties![key!] = keys.length
+					? {
+							type: 'object',
+							properties: {},
+						}
+					: value;
+			}
+
+			head = head.properties![key] as JSONSchema6;
```
thgreasi: Can we be sure that this isn't a boolean as the typings say?

src/components/Filters/index.tsx
```diff
@@ -104,40 +105,55 @@ interface FiltersState {
...
+			edit: [this.getCleanEditModel(null, flatSchema)],
 		};
+	}
 
-		// Clean exsting rules on load
+	componentWillReceiveProps(prevProps: FiltersProps) {
+		// If the schema prop updates, also update the internal 'flat' schema
+		if (!isEqual(prevProps.schema, this.props.schema)) {
+			this.setState({
+				schema: SchemaSieve.flattenSchema(this.props.schema),
```
thgreasi: Should we also call `getCleanEditModel` in here?

src/components/Filters/index.tsx
thgreasi: How about doing something like:
```
const edit = [];
this.state= {
	/*...*/
	schema: flatSchema,
	edit,
};
edit.push(this.getCleanEditModel(null));
```
and drop the added schema property? Would that be too :imp::imp::imp: ?

src/components/Filters/index.tsx
```diff
@@ -104,40 +105,55 @@ interface FiltersState {
...
 			field,
 			operator,
 			value: '',
 		};
 	}
 
-	getOperators(field: string) {
-		const { schema } = this.props;
+	getOperators(field: string, schema?: JSONSchema6) {
+		if (!schema) {
```
thgreasi: > Could this be in the method declaration args as a default value `schema: JSONSchema6 = this.state.schema`?

src/components/Filters/index.tsx
```diff
@@ -104,40 +105,55 @@ interface FiltersState {
...
+			this.setState({
+				schema: SchemaSieve.flattenSchema(this.props.schema),
+			});
+		}
 	}
 
-	getCleanEditModel(field?: string) {
-		const { schema } = this.props;
+	getCleanEditModel(field?: string | null, schema?: JSONSchema6) {
+		if (!schema) {
```
thgreasi: Could this be in the method declaration args as a default value `schema: JSONSchema6 = this.state.schema`?

src/components/Filters/index.tsx
LucianBuzzo: yeah that would be fine

LucianBuzzo:
@pimterry I thought about it quite a bit and I prefer this approach, as it keeps the UI logic fairly simple

thgreasi:
I would also agree that any operations on the schema seem to better match the achieve class than being part of the Filter **UI** component. If it makes the UI mucheasier than fine. I guess that we could improve it later.

src/components/Filters/SchemaSieve.ts
LucianBuzzo: Good point, I'll change the implementation

PR #392, opened 2018-04-16
Add property to Filters to control which subcomponents are rendered

### LINKED ISSUE #380: Split filters into four sub-components
Right now it's only possible to use filters wholly, with the add filters button, search bar, views button, and active filters. Instead, it'd be great if we could have a base wrapper component that takes these subcomponents optionally.

LucianBuzzo:
Fixes #380

change-type: minor


src/components/Filters/index.tsx
```diff
@@ -286,43 +284,66 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
...
 
+	shouldRenderComponent(mode: FilterRenderMode): boolean {
+		// If a render mode is not specified, render all components
+		if (!this.props.renderMode) {
+			return true;
+		}
+
+		const allowedModes = Array.isArray(this.props.renderMode) ? this.props.renderMode : [ this.props.renderMode ];
+
+		if (allowedModes.indexOf('all') > -1) {
```
Shou: How come we're not using `.includes('all')`?

LucianBuzzo: I'll change it to use includes

Issue #385, opened 2018-04-14
Gauge: add the option to display the counts inside each pie piece


thgreasi:
HQ: https://github.com/resin-io/hq/issues/1271

dimitrisnl:
Forgot to close this one

PR #396, opened 2018-04-17
Show partial totals in Gauge component

dimitrisnl:
Connects-To: #385
Change-Type: minor

screen shot 2018-04-30 at 10 07 04 am

screen shot 2018-04-30 at 10 14 44 am

screen shot 2018-04-30 at 10 07 23 am

---- Autogenerated Waffleboard Connection: Connects to #385


dimitrisnl:
@LucianBuzzo , couple questions here.

1. Why do the tests fail? I get

```
-     className="sc-dnqmqq gDpXJY sc-gzVnrw iYSyBq"
+     className="sc-dnqmqq gDpXJY sc-gzVnrw jBacMi"
```

so the problem is with the added class.

2. How can I make it look 'nicer'?

3. I used `partialTotals` as a name for the prop. Is it in line with the naming conventions?

LucianBuzzo:
@dimitrisnl See this line from the test output
```
Snapshot Summary
 › 2 snapshot tests failed in 2 test suites. Inspect your code changes or run `npm run jest -- -u` to update them.
```

If you've made changes to a component you'll need to regenerate the snapshots for them, see https://facebook.github.io/jest/docs/en/snapshot-testing.html#updating-snapshots

I don't think `partialTotals` makes much sense, how about `showSegmentTotals`?

You could try using the `isLight` method from the `colors` package and then changing the number color accordingly https://www.npmjs.com/package/color#luminosity

dimitrisnl:
screen shot 2018-04-17 at 2 40 52 pm

dimitrisnl:
screen shot 2018-04-18 at 1 28 28 pm

It seems like there are some serious issues while using data with big disparity.

LucianBuzzo:
@dimitrisnl I think we might just need to implement the totals as tooltips?

dimitrisnl:
@LucianBuzzo This is what I'm working on atm

Other approaches
1. Hide the values with they are less than 10% of the total - Doesn't look really good.
2. Add the totals in the legend - Way too small, last place to look.
3. Use another pie solution for the library, with build in capabilities - Better avoid that

dimitrisnl:
Couldn't make tooltips happen, and ended up using a simple title tag
untitled

thgreasi:
Already better that what we already have.
Would it make sense to have these numbers on the legend instead?
It would probably make them appear one per line which would increase the height of the respective card in the Dashboard, so that's probably a bad idea for that use case...

dimitrisnl:
The negative with using titles is that it has a slight delay. WIth tooltips, you get immediate feedback. 

As for the legends, I thought about that too but it does not look great. The font size is pretty small, and it feels really cluttered

thgreasi:
Oh wait... If it's SVG then we can do CSS only tooltips using `:hover { content: attr(...); position absolutel....}`, can't we?

dimitrisnl:
@thgreasi Huh sounds interesting. Will give it a go!

dimitrisnl:
@thgreasi Not sure I can use pseudo elements with SVG paths

LucianBuzzo:
@dimitrisnl Can we not just use chart.js? That demo looks lovely

dimitrisnl:
screen shot 2018-04-26 at 2 34 39 pm

@LucianBuzzo Went ahead and tried it.  There were certain caveats but everything seems to be working alright.

dimitrisnl:
Actually there a number of issues to be addressed.

1. Fix the failing tests
2. Do the proper actions after props change

LucianBuzzo:
@dimitrisnl Looks awesome!

dimitrisnl:
@LucianBuzzo Have some serious issues with the jest snapshot generation.

I get a weird `TypeError: Cannot read property 'length' of null`. Not quite sure what should I be looking for.

package.json
```diff
@@ -76,13 +76,15 @@
...
     "ajv-keywords": "^3.1.0",
     "babel-core": "^6.26.0",
+    "chart.js": "^2.7.2",
     "color": "^2.0.0",
     "color-hash": "^1.0.3",
     "copy-to-clipboard": "^3.0.8",
     "grid-styled": "^3.2.1",
     "lodash": "^4.17.4",
     "moment": "^2.20.1",
     "paths-js": "^0.4.5",
```
LucianBuzzo: I think we can remove `paths-js` now

src/components/Gauge.js
```diff
@@ -76,119 +53,124 @@ const StatusGaugeWrapper = styled.div`
...
+class StatusGauge extends React.Component {
+  constructor (props) {
+    super(props)
+
+    const { data } = props
+
+    const values = data.map(x => x.value)
+    const labels = data.map(x => x.name)
+    const colors = data.map(x => x.color)
+    const count = values.reduce((carry, item) => carry + item, 0)
```
LucianBuzzo: This means that the component won't update if you change the props after it mounts

dimitrisnl:
Really weird `Navbar` & `Form` fail, yet I have made no changes in their respective files.

Updating the snapshots and including them.

src/components/Gauge.js
```diff
@@ -1,53 +1,30 @@
...
-      text-decoration: none !important;
-    }
-
-    .gauge__graph__total-label {
-      transform: translateX(0);
-    }
-  }
-
-  .gauge__graph {
+  .gauge__doughnut {
```
thgreasi: Could this still be `.gauge__graph` since we [are already using this in the UI to add some margins](https://github.com/resin-io/resin-ui/blob/26868a2ecb592025e4922c2aad8ef39d3841073b/src/scripts/modules/applications-list/applications-list.less#L27).
We should confirm whether this would affect the UI in a negative way.

src/components/Gauge.js
```diff
@@ -76,119 +53,141 @@ const StatusGaugeWrapper = styled.div`
...
 
-const GaugeTotalText = styled.text`
-  font-family: ${font};
-  font-size: ${props => px(props.count > 999 ? 20 : 26)};
-  fill: #5e5e5e;
-  letter-spacing: -1px;
-  text-anchor: middle;
-  transform: translateX(0);
-`
+let DoughnutChart = null
```
thgreasi: !!! The Gauge will be appearing several times in the page, so this should probably be a property of the StatusGauge class.

src/components/Gauge.js
```diff
@@ -76,119 +53,141 @@ const StatusGaugeWrapper = styled.div`
...
+
+    const data = this.formatGraphData(values, colors, labels, count)
+    const options = this.formatGraphOptions(count)
+    const pluginOptions = this.formatPluginOptions(
+      doughnutDefaults,
+      font,
+      count
+    )
+
+    // eslint-disable-next-line
```
thgreasi: In my tests this seems to not be needed.

src/components/Gauge.js
```diff
@@ -76,119 +53,141 @@ const StatusGaugeWrapper = styled.div`
...
+      tooltips: { enabled: count > 0 }
+    }
+  }
+
+  formatPluginOptions (defaultValues, font, count) {
+    return {
+      primaryLabel: {
+        ...defaultValues.primaryLabel,
+        font,
+        size: count > 99 ? (count > 999 ? 22 : 30) : 38,
```
thgreasi: How about preserving the inequality check order so that we can drop the parenthesis, like:
```
count > 999 ? 22 :
count > 99 ? 30 :
38
```

src/components/Gauge.js
```diff
@@ -76,119 +53,141 @@ const StatusGaugeWrapper = styled.div`
...
+    }
+  }
+
+  render () {
+    /**
+     * data: [{
+     *   key: string
+     *   value: number
+     *   color: string
+     *   name: string
```
thgreasi: It seems that you would love to use TS in here :)

src/components/Gauge.js
```diff
@@ -76,119 +53,141 @@ const StatusGaugeWrapper = styled.div`
...
+      datasets: [
+        {
+          backgroundColor: count > 0 ? colors : this.props.placeholderColor,
+          data: count > 0 ? values : [1]
+        }
+      ]
+    }
+  }
+
+  formatGraphOptions (count) {
```
thgreasi: How about `getGraphOptions` and `getLabelOptions`?
Although I would also be fine with combining those two into a single one.

src/components/Gauge.js
```diff
@@ -76,119 +53,141 @@ const StatusGaugeWrapper = styled.div`
...
+      elements: {
+        arc: {
+          borderWidth: 0
+        }
+      },
+      tooltips: { enabled: count > 0 }
+    }
+  }
+
+  formatPluginOptions (defaultValues, font, count) {
```
thgreasi: We could reference `defaultValues` and `font` from the imports directly instead of passing them as args, since they are not modified at all.

src/chartPlugins.js
```diff
@@ -0,0 +1,46 @@
+export const doughnutDefaults = {
+  primaryLabel: {
+    text: '',
+    font: 'Arial',
+    color: '#5e5e5e',
+    size: 38,
+    offsetFromCenter: 10
+  },
+  secondaryLabel: {
```
thgreasi: NIT: how about splitting to doughnutPrimaryLabelStyles and doughnutSecondaryLabelStyles?
Even though those two could just be two consts inside the gauge component itself to be near their sole consumer.
If you feel that the component got too large for a single file, then it might be time to have a folder for it.

src/components/Gauge.js
dimitrisnl: Tried so, had some major issues with Chart.js types.

src/components/Gauge.js
```diff
@@ -76,119 +53,146 @@ const StatusGaugeWrapper = styled.div`
...
+    this.generateDoughnut(this.props)
+  }
+
+  componentWillReceiveProps (nextProps) {
+    this.generateDoughnut(nextProps)
+  }
+
+  componentWillUnmount () {
+    if (this.DoughnutChart) {
+      this.DoughnutChart.destroy()
```
thgreasi: We should also assign it to `null` after calling `destroy()`, otherwise would be called for a second time in `generateDoughnut()` if the components mounts again.
We could actually have a `destroy()` method as part of the component, encapsulating the whole `if`.

src/components/Gauge.js
thgreasi: If-you-are-still-in-the-mood: You could use some `any` :)

src/components/Gauge.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+	.gauge__legend-name {
+		font-size: 11px;
+	}
+`;
+
+class StatusGauge extends React.Component<GaugeProps, {}> {
+	private DoughnutChart: Chart;
+	private DoughnutCtx: CanvasRenderingContext2D | null;
+
```
thgreasi: Can we also have a constructor to do the `super(props)` call, like we do in all components?

src/components/Gauge.tsx
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	.gauge__legend-name {
+		font-size: 11px;
+	}
+`;
+
+class StatusGauge extends React.Component<GaugeProps, {}> {
+	private DoughnutChart: Chart;
+	private DoughnutCtx: CanvasRenderingContext2D | null;
```
thgreasi: I think that these should start with a lower case letter, especially the chart instance.

src/components/Gauge.tsx
```diff
@@ -0,0 +1,209 @@
...
+		this.generateDoughnut(this.props);
+	}
+
+	componentWillReceiveProps(nextProps: GaugeProps) {
+		this.generateDoughnut(nextProps);
+	}
+
+	componentWillUnmount() {
+		if (this.DoughnutChart) {
+			this.DoughnutChart.destroy();
```
thgreasi: We should probably also set `this.DoughnutChart` to `null` right after `destroy()`, otherwise `destroy()` will be called again if the components re-mounts.
We could actually encapsulate this as a `destroy()` method for this component.

src/components/Gauge.tsx
dimitrisnl: Oh, I included it but seems like I removed it afterwards. WIll do.

src/chartPlugins.js
dimitrisnl: @thgreasi I decalre them in the `chartPlugins` mostly for fallback reasons. Should I move them in the `Gauge` component?

dimitrisnl:
THANK YOU BOT

dimitrisnl:
Let's see if @LucianBuzzo has any objections, and I'll happily merge it

src/components/Gauge.tsx
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+
+	render() {
+		return (
+			<StatusGaugeWrapper {...this.props}>
+				<div className="gauge__graph">
+					<canvas
+						ref={canv =>
+							(this.doughnutCtx = canv ? canv.getContext('2d') : null)
```
LucianBuzzo: This is really weird, you assign `canv` to `this.doughnutCtx` and then immediately overwrite it I think. Should the `=` here be `===` ?
Either way I'd prefer to see the private property just be the Canvas, and have the `getContext` call happen on demand in `generateDoughnut`.

src/components/Gauge.tsx
```diff
@@ -0,0 +1,216 @@
...
+		}
+		// Redraw the chart on props update
+		this.destroyChart();
+
+		const ctx = this.doughnutCtx;
+
+		const { title } = props;
+		const { values, colors, labels, count } = this.normalizeProps(props.data);
+
+		const data = this.getGraphData(values, colors, labels, count);
```
LucianBuzzo: Another weird bit of logic here, can you just remove `normalizeProps` and pass `props.data` directly to `getGraphData`. At the moment this is unnecessarily complicated.

src/components/Gauge.tsx
```diff
@@ -0,0 +1,216 @@
...
+			options,
+			plugins: [doughnutLabel],
+		});
+	}
+
+	normalizeProps(data: ChartEntry[]) {
+		const values = data.map(x => x.value);
+		const labels = data.map(x => x.name);
+		const colors = data.map(x => x.color);
+		const count = values.reduce(
```
LucianBuzzo: You're iterating over the data array four times here. I'd prefer to see whole result constructed from a single reduce function, something like this should work:
```
_.reduce(data, (carry, item) => {
	carry.values.push(item.value);
	carry.labels.push(item.name);
	carry.colors.push(item.color);
	carry.count += item.value
	return carry
}, { values: [], labels: [], colors: [], count: 0 })
```

src/components/Gauge.tsx
dimitrisnl: Yeah, completely overlooked this one. Will sort it out.

src/components/Gauge.tsx
dimitrisnl: That would be better

Issue #400, opened 2018-04-17
Remove jest-extended from project


LucianBuzzo:
I encountered a number of bugs in the `jest-extended` project, which means that many tests are giving false positives.
https://github.com/jest-community/jest-extended/issues/121
https://github.com/jest-community/jest-extended/issues/119

PR #402, opened 2018-04-17
397 nested filter

LucianBuzzo:
Fixes the carnage left by the `jest-extended` bug 
connects to #400


src/components/DataTypes/date-time.tsx
```diff
@@ -1,10 +1,14 @@
 import { JSONSchema6 } from 'json-schema';
 import assign = require('lodash/assign');
+import memoize = require('lodash/memoize');
+import * as moment from 'moment';
 import * as React from 'react';
 import { DataTypeEditProps, InputProps } from 'rendition';
 import * as utils from '../../utils';
 import Input from '../Input';
 
+(moment as any).suppressDeprecationWarnings = true;
```
pimterry: What deprecation messages are these? Why do we need to depend on deprecated functionality?

src/components/DataTypes/date-time.tsx
```diff
@@ -14,6 +18,11 @@ const getDefaultDate = (): string => {
 	return `${today}T00:00`;
 };
 
+// Normalize a timestamp to a RFC3339 timestamp, which is required for JSON
+// schema.
+// http://momentjs.com/docs/#/displaying/format/#default-format
```
pimterry: This link doesn't work for me, I think you want https://momentjs.com/docs/#default-format

src/components/DataTypes/date-time.tsx
LucianBuzzo: I was trying to clean up test output a little bit, on reflection its best to leave this in!

PR #405, opened 2018-04-18
[Filters] Use flattening methods on inbound and outbound views and filters

### LINKED ISSUE #404: Filters component should 'unflatten' views, the same way it does with the filters themselves

LucianBuzzo:
Fixes #404

change-type: patch


src/components/Filters/index.tsx
```diff
@@ -111,25 +111,43 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
 		const { filters, schema, views } = this.props;
 
 		const flatSchema = SchemaSieve.flattenSchema(schema);
+		const flatViews = this.flattenViews(views || []);
+		const flatFilters = (filters || []).map(filter =>
```
pimterry: If you like, you can actually just put defaults like this in the deconstruction above:

```ts
const { filters = [], schema, views = [] } = this.props;

...
```

src/components/Filters/index.tsx
```diff
@@ -111,25 +111,43 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
...
+		if (nextProps.filters && !isEqual(nextProps.filters, this.props.filters)) {
 			this.setState({
-				schema: SchemaSieve.flattenSchema(this.props.schema),
+				filters: nextProps.filters.map(filter =>
+					SchemaSieve.flattenSchema(filter),
+				),
+			});
+		}
+
+		if (nextProps.views && !isEqual(nextProps.views, this.props.views)) {
```
pimterry: Because each of these checks check whether `views`/`filters` are truthy & different, not just different, with this change the state won't update at all if you remove the filters or views prop entirely.

Is that right? Sounds like a bug.

LucianBuzzo: You're right, that sounds like a bug. If there were filters/views and they get set to a falsey value, then we should set the state to an empty array, like we do in the constructor

src/components/Filters/index.tsx
```diff
@@ -307,7 +343,9 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
 			return true;
 		}
 
-		const allowedModes = Array.isArray(this.props.renderMode) ? this.props.renderMode : [ this.props.renderMode ];
+		const allowedModes = Array.isArray(this.props.renderMode)
```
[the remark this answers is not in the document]
LucianBuzzo: Great idea

src/components/Filters/index.tsx
```diff
@@ -111,25 +111,43 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
...
 
 		this.state.edit.push(this.getCleanEditModel());
 	}
 
-	componentWillReceiveProps(prevProps: FiltersProps) {
+	componentWillReceiveProps(nextProps: FiltersProps) {
 		// If the schema prop updates, also update the internal 'flat' schema
-		if (!isEqual(prevProps.schema, this.props.schema)) {
+		if (!isEqual(nextProps.schema, this.props.schema)) {
+			this.setState({
```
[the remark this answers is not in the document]
LucianBuzzo: This is a damn good idea

LucianBuzzo:
I've added some additional testing for the `<Filters>` component as well

src/components/Filters/index.tsx
```diff
@@ -108,29 +110,49 @@ interface FiltersState {
...
 			schema: flatSchema,
 			edit: [],
 		};
 
 		this.state.edit.push(this.getCleanEditModel());
 	}
 
-	componentWillReceiveProps(prevProps: FiltersProps) {
+	componentWillReceiveProps(nextProps: FiltersProps) {
+		const newState: any = {};
```
thgreasi: Could this be types as `Partial<FiltersState>`?

LucianBuzzo: I tried this and it turned into a bit of a nightmare trying to make the compiler happy!

thgreasi: Maybe `{ [key: string]: any }` could be slightly better.

Issue #407, opened 2018-04-19
Adding filters without triggering Select onChange results in empty string value


Shou:
Steps to reproduce:

1. Click add filter button
2. Change to any property of a type that renders a select, e.g. booleans or finite enumerables
3. Save
4. Filter value is now empty string instead of first option's value

PR #408, opened 2018-04-19
Fix add filter selects' default value

Shou:
We ensure <select>s in the filter adding modal default to the top value,
instead of an empty string until the `onChange` event is triggered.

Fixes: #407
Change-Type: patch


src/components/DataTypes/enum.tsx
```diff
@@ -107,19 +108,30 @@ export const createFilter = (
...
-		{...props}
-		value={value}
-		onChange={(e: React.FormEvent<HTMLSelectElement>) =>
-			onUpdate(e.currentTarget.value)
-		}
-	>
-		{map(schema.enum, (item: string) => <option key={item}>{item}</option>)}
-	</Select>
-);
+type DataTypeSelectProps = DataTypeEditProps & SelectProps & { slim?: boolean };
```
Shou: I'm guessing we want to consolidate the type declaration above and this one somewhere, but I'm not familiar enough with `rendition` to know where.

Shou:
@lucianbuzzo what do you think of this now?

src/components/Filters/index.tsx
```diff
@@ -416,6 +417,15 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
 				>
 				{map(this.state.edit, ({ field, operator, value }, index) => {
 					const operators = this.getOperators(field);
+					const fieldObject = this.state.schema.properties![field];
+
+					if (value === '') {
```
LucianBuzzo: We shouldn't be updating state as part of the render method like this, Can you move this logic into `setEditField` ?

Shou: @LucianBuzzo that won't solve the problem as it won't be called until the user triggers the `onChange` event. We need to somehow set the value as a default.

LucianBuzzo: @Shou You can put your logic in `getCleanEditModel` as that gets called on construction and when the field selector changes. I just messed around with it and this seems to work a treat
```ts
	getCleanEditModel(field?: string | null) {
		const schema = this.state.schema;
		if (!field) {
			field = Object.keys(schema.properties!).shift()!;
		}
		const operator = this.getOperators(field).shift()!.slug;
		let value: any = '';

		const subschema = schema.properties![field];
		if (typeof subschema !== 'boolean') {
			if (subschema.enum) {
				value = subschema.enum[0] || '';
			}

			if (subschema.type === 'boolean') {
				value = true;
			}
		}

		return {
			field,
			operator,
			value,
		};
	}
```

Shou: Sorry it took so long here – yeah that works, thanks! I've updated the PR with those changes now.

Issue #383, opened 2018-04-11
Table: should be able to handle clickable rows & ckeckboxes at the same time


thgreasi:
Expected: events originating from the checkbox of the selection column should not trigger the `onRowClick` handler.

ATM: clicking the checkbox also triggers the `onRowClick` handler as well.

PR #417, opened 2018-05-07
Table: add cell props & make the cells clickable instead of the rows

thgreasi:
Moves the row checkbox outside the anchor, working around an issue in the UI where we couldn't prevent the anchor click when clicking the checkbox

Resolves: #383
Change-Type: minor


package.json
```diff
@@ -72,8 +72,8 @@
     "grid-styled": "2.0.0-10",
     "lodash": "^4.17.4",
     "moment": "^2.20.1",
-    "paths-js": "^0.4.5",
-    "react-aria-tooltip": "^1.1.1",
+    "paths-js": "^0.4.7",
```
dimitrisnl: Do we still use `path-js`? I removed it after refactoring `DeviceStatusGauge`

thgreasi: That's on the v3 branch that still uses path-js.

Issue #419, opened 2018-05-14
Filter with dropdown options should have the first value selected by default


dimitrisnl:
### Expected behavior

When I want the first option from the drop-down, clicking save should suffice.

### Expected behavior

The first option isn't actually selected by default, and the filter will not be applied

### Screenshots
![screen shot 2018-05-14 at 2 49 21 pm](https://user-images.githubusercontent.com/4951004/39995833-35887280-5786-11e8-89d8-e083f6c53136.png)

![screen shot 2018-05-14 at 2 49 09 pm](https://user-images.githubusercontent.com/4951004/39995831-35670ba4-5786-11e8-88df-cc1714fcd6ae.png)

dimitrisnl:
closed as duplicate of #407

PR #420, opened 2018-05-14
Fix default values of filters

dimitrisnl:
---- Autogenerated Waffleboard Connection: Connects to #419


dimitrisnl:
@LucianBuzzo Would you mind taking a look when you're available?

LucianBuzzo:
@dimitrisnl Is this the same as https://github.com/resin-io-modules/rendition/pull/408 ? If it is, could you pick up where @Shou left it?

dimitrisnl:
@LucianBuzzo Oh, I completely missed this PR. Will do so.

PR #425, opened 2018-05-23
Select component improvements

LucianBuzzo:
This PR is comprised of three commits
- Convert the `Select` component to TS
- Write more complex tests for the `Select` component
- Add a `disabled` prop to the `Select` component

Connects to #424


src/components/Select.tsx
```diff
@@ -0,0 +1,73 @@
...
+		px(props.emphasized ? props.theme.space[5] : props.theme.space[4])};
+  font-size: inherit;
+  border: 1px solid ${props => props.theme.colors.gray.main};
+  padding-top: 0px;
+  padding-bottom: 0px;
+  padding-left: 16px;
+  padding-right: ${props =>
+		px(props.emphasized ? props.theme.space[5] : props.theme.space[4])};
+  background-color: white;
+  -webkit-appearance: none;
```
dimitrisnl: I know we can't use something like [autoprefixer](https://github.com/postcss/autoprefixer) but it would be great to drop the vendor prefixes somehow.

src/components/Select.tsx
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
+import { compose } from 'recompose';
+import { SelectProps } from 'rendition';
+import styled, { StyledFunction, withTheme } from 'styled-components';
+import asRendition from '../asRendition';
+import { radius } from '../theme';
+import { px } from '../utils';
+
+const Base = (styled.select as StyledFunction<
```
thgreasi: How about using `withProps that does exactly this cast for you?
```
import { px, withProps } from '../utils';

const Base = withProps<{ emphasized?: boolean }>()(styled.select)`...`;
```

src/components/Select.tsx
```diff
@@ -0,0 +1,75 @@
...
+  appearance: none;
+  width: 100%;
+  min-width: 90px;
+
+  &:hover {
+    box-shadow: 0 0 4px 1px rgba(0, 0, 0, 0.1);
+  }
+`;
+
+const Wrapper = (styled.span as StyledFunction<SelectProps>)`
```
thgreasi: Could be `withProps<SelectProps>()(styled.span)` but it might be an overkill for this case.
PS: Even though I like it more since it also re-exposes the span prop typings.

src/components/Select.tsx
```diff
@@ -0,0 +1,75 @@
...
+	onChange,
+	value,
+	...props
+}: SelectProps) => {
+	return (
+		<Wrapper emphasized={emphasized} {...props}>
+			<Base
+				disabled={disabled}
+				emphasized={emphasized}
+				value={value === null ? '' : value}
```
thgreasi: I guess that `undefined` has a special meaning here.

src/components/Select.tsx
LucianBuzzo: `styled-components` auto prefixes by default, so I can just remove these

src/components/Select.tsx
LucianBuzzo: The JSX attribute for `value` can be `string | number | string[] | undefined`, where undefined just means that there is no value specified.

src/components/Select.tsx
LucianBuzzo: cool

src/components/Select.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { compose } from 'recompose';
+import { SelectProps } from 'rendition';
+import styled, { withTheme } from 'styled-components';
+import asRendition from '../asRendition';
+import { radius } from '../theme';
+import { px, withProps } from '../utils';
+
+const Base = withProps<
+	{ emphasized?: boolean } & React.HTMLProps<HTMLSelectElement>
+>()(styled.select)`
```
thgreasi: This should have ended up as `withProps<{ emphasized?: boolean }>()(styled.select)`. Did you face any issues?

src/components/Select.tsx
```diff
@@ -0,0 +1,73 @@
...
+  border-radius: ${px(radius)};
+  height: ${props =>
+		px(props.emphasized ? props.theme.space[5] : props.theme.space[4])};
+  font-size: inherit;
+  border: 1px solid ${props => props.theme.colors.gray.main};
+  padding-top: 0px;
+  padding-bottom: 0px;
+  padding-left: 16px;
+  padding-right: ${props =>
+		px(props.emphasized ? props.theme.space[5] : props.theme.space[4])};
```
thgreasi: Somehow the styles for `Base` ended up with two space indentation except from the lines that use `px` which are indented with two tabs !? :alien: :satellite:

PR #432, opened 2018-06-08
Add missing methods to SchemaSieve typings

LucianBuzzo:
change-type: patch


LucianBuzzo:
@thgreasi Snapped, I triggered the merge before I saw your comment, I'll add it as an issue!

PR #435, opened 2018-06-08
Add filter handling for arrays of strings

### LINKED ISSUE #434: Handle filtering of arrays

LucianBuzzo:
Fixes #434

change-type: minor


src/components/DataTypes/array.tsx
```diff
@@ -0,0 +1,184 @@
...
+		getLabel: (_s: JSONSchema6) => 'contains',
+	},
+	not_contains: {
+		getLabel: (_s: JSONSchema6) => 'does not contain',
+	},
+};
+
+type OperatorSlug = keyof typeof operators;
+
+interface StringFilter extends JSONSchema6 {
```
thgreasi: Should this be `ArrayFilter` instead?

src/components/DataTypes/array.tsx
```diff
@@ -0,0 +1,184 @@
...
+					},
+				},
+			],
+		});
+	}
+
+	return base;
+};
+
+export const Edit = ({
```
thgreasi: Maybe we should check the value of `items.type` and load the `Edit` component from the respective DataType (string, number, etc)

Issue #440, opened 2018-06-13
Have a standalone search field


dimitrisnl:
As per @konmouz the search field that can be been seen [here](https://resin-io-modules.github.io/rendition/?selectedKind=Filters&selectedStory=Render%20modes&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) should be a component by itself.

Use case: The `Search the docs` prompt in the `Need Help?` bubble

PR #441, opened 2018-06-13
Include Search Component

dimitrisnl:
---- Autogenerated Waffleboard Connection: Connects to #440


dimitrisnl:
screen shot 2018-06-13 at 4 30 25 pm

I updated so that it can be resizable enough for cases like
image

typings/rendition.d.ts
```diff
@@ -322,6 +322,14 @@ declare module 'rendition' {
...
 	> {}
 
+	interface SearchProps extends DefaultProps {
+		dark?: boolean;
+		disabled?: boolean;
+		placeholder?: string;
+		value: string;
+		onChange: (value: any) => void;
+	}
+
```
LucianBuzzo: You need to add a class definition here as well, something like:
```
class Search extends RenderableElementWithProps<SearchProps, any> {}
```

Issue #453, opened 2018-06-15
We don't fetch font weight of 700


dimitrisnl:
Thus bold titles don't render correctly.

dimitrisnl:
working on it as part of https://github.com/resin-io-modules/rendition/issues/431

Issue #431, opened 2018-06-07
Add Card component


thgreasi:
The `box-shadow` should be `0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12), 0 3px 1px -2px rgba(0, 0, 0, 0.2)`.
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/7YsQn5uXJCydF9Bbh_rJKMUqBpC
See: https://app.zeplin.io/project/59cbb35a8edcbe2d8fbe5678/screen/59d4f58a68cf621d3adab6e5

dimitrisnl:
Simple card? No header?

thgreasi:
Ideally all the variants of it + the one with lines as found in the billing page for the invoices.

dimitrisnl:
Gotcha

PR #455, opened 2018-06-15
Card component

dimitrisnl:
---- Autogenerated Waffleboard Connection: Connects to #453
---- Autogenerated Waffleboard Connection: Connects to #431


dimitrisnl:
screen shot 2018-06-15 at 4 32 08 pm

screen shot 2018-06-15 at 4 32 27 pm

screen shot 2018-06-15 at 4 46 49 pm

dimitrisnl:
### Notes:

- Allowed custom property to set `min-height`
- The colors are based from these designs https://app.zeplin.io/project/59cbb35a8edcbe2d8fbe5678/screen/59d4f58a68cf621d3adab6e5 - Pretty sad that I couldn't find them in our colors file. Should we revisit what colors we run? cc @LucianBuzzo @konmouz @thgreasi 
- The header can take as JSX element as a call to action - though it was better that way
- !! When we have a list there is different margin under the header that when we simply have a paragraph. Accounted for that case, but still seems inconsistent.

typings/rendition.d.ts
```diff
@@ -171,6 +171,16 @@ declare module 'rendition' {
 		text: string;
 	}
 
+	interface CardProps extends DefaultProps {
+		title?: string;
+		cta?: any;
+		rows?: any;
+		children?: any;
```
dimitrisnl: Updating these typing atm

Issue #460, opened 2018-06-20
Button labels are misaligned


dimitrisnl:
**Describe the bug**
The buttons have a small padding at the bottom that makes the label misaligned vertically

**Expected behavior**
Button label should be centered vertically

PR #462, opened 2018-06-20
Remove vertical padding from Button

dimitrisnl:
Connects-To: #460
Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #460


dimitrisnl:
We're using flex, with `align: center` so any extra padding was not needed.

dimitrisnl:
screen shot 2018-06-20 at 12 41 16 pm

Minor cosmetic change, but hey.

Issue #467, opened 2018-06-21
Update buttons as per styleguide


dimitrisnl:
**Describe the bug**
The issue is about the default button state, but I opened it as a generic issue, in case I need to do any more touchups.

**Expected behavior**
![screen shot 2018-06-21 at 11 21 57 am](https://user-images.githubusercontent.com/4951004/41706786-54f1ffa6-7545-11e8-8906-ecd2d4a5a063.png)

The border should be different

Issue #461, opened 2018-06-20
Buttons transition


dimitrisnl:
**Is your feature request related to a problem? Please describe.**
Buttons feel a bit static

**Describe the solution you'd like**
Could try out a small transition delay of 0.5s or 1s

dimitrisnl:
Yes. Should have clarified that.

Something like `transition: <property> .1s ease-in-out`

dimitrisnl:
I was going through refactoring the `Need help` bubble in the dashboard to React. So I'm checking how rendition is up to date with @konmouz style guide, hence the issues that might pop up.

PR #468, opened 2018-06-21
Update Buttons

dimitrisnl:
-https://app.zeplin.io/project/59cbb35a8edcbe2d8fbe5678/screen/59cd74ed3b9450c01addba94
---- Autogenerated Waffleboard Connection: Connects to #461
---- Autogenerated Waffleboard Connection: Connects to #467


dimitrisnl:
examples

screen shot 2018-06-21 at 3 07 48 pm

screen shot 2018-06-21 at 3 07 59 pm

src/theme.ts
```diff
@@ -21,27 +21,27 @@ const colors = {
...
+		light: '#A6D989',
+		extraLight: '#EDF7E7',
+		dark: '#1e4707',
 	},
 	info: {
 		main: '#2297DE',
-		semilight: '#90CBEE',
-		light: '#E9F4FB',
+		light: '#90CBEE',
+		extraLight: '#E9F4FB',
```
thgreasi: Oh wait. These are breaking changes here! These are exported and we use them in the UI.
Please reconsider not changing the names if it makes sense for now.
If we do want to rename these, then please also update the typings.

dimitrisnl: Ah I see, though we only use them in two places in 2FA.

dimitrisnl: Unfortunately the colours before had big variance for something called semilight, so it would be best to keep it somehow consistent.

Do we really want to mark it as major when the UI change is so minor?

LucianBuzzo: @dimitrisnl Can you change this to a major version bump and add a migration doc please? I'm not opposed to doing major bumps on small things like this, we just need to be clear about what pieces of the API changed

dimitrisnl: Btw I'm not using it the new properties in this PR so I could discard these changes.

LucianBuzzo: @dimitrisnl If you'd like to discard these changes, for now, that's fine as well. You could make an issue to review the color swatches for v5

dimitrisnl: Sure this would be for the best @LucianBuzzo . Cherry picking the changes atm

dimitrisnl:
Won't merge it just yet - let's shit the v4 first

dimitrisnl:
Feel like it's a good time to merge it what do you think @LucianBuzzo @thgreasi ?

LucianBuzzo:
@dimitrisnl Yes please!

PR #471, opened 2018-06-21
Make tooltips work with disabled elements

### LINKED ISSUE #470: Tooltips never disappear when used with disabled elements
**Describe the bug**
The `onMouseLeave` event is never fired on React elements, which means the tooltip is never hidden if the event type is 'hover'. See https://github.com/facebook/react/issues/4251#issuecomment-334266778

**To Reproduce**
Steps to reproduce the behaviour:
```jsx
<Button
    tooltip={'I will never leave'}
    disabled={true}
>
    Hover over me
</Button>
```

LucianBuzzo:
Fixes #470

change-type: patch


src/tooltips.tsx
```diff
@@ -294,6 +308,10 @@ export class Tooltips {
 					props.onMouseEnter = (e: Event) => {
 						showFn(e);
 						oldMEFn(e);
+
+						if ((e.target as any).disabled) {
```
thgreasi: Should we do that only for buttons? We "can" have the `disabled` attr on many elements. On the other hand, it will probably not hurt anyone to use this approach even for disabled inputs, I guess.
Your call.

src/tooltips.tsx
```diff
@@ -242,6 +243,19 @@ export class Tooltips {
 
 		document.body.appendChild(tooltipRoot);
 
+		// This is a special case for handling disabled elements. When a tooltip is
+		// shown on a tooltip element, the flag is activated and the next time this
+		// event listener is called the tooltip is hidden. This happens because
+		// a mouseover event will be triggered when the cursor leaves a disabled
+		// element.
+		// See: https://github.com/facebook/react/issues/4251#issuecomment-334266778
+		document.addEventListener('mouseover', () => {
```
thgreasi: We should also remove the listener at some point.
Do we need to also check the origin of the event?
We could probably add a single event listener for all tooltip components, on the first `componentDidMount()` call or something, and then pub-sub to the rest of them.

LucianBuzzo: The `Tooltips` class is a singleton and only sets up a single event listener, so AFAIK this is not a concern

src/tooltips.tsx
LucianBuzzo: AFAIK the `mouseleave` event won't be called for any disabled element, so we need to keep this logic here.

thgreasi:
Sorry for the confusion

PR #475, opened 2018-06-22
Add markdown widget to Form component

### LINKED ISSUE #472: Add markdown widget to Form component
**Describe the solution you'd like**
A user should be able to use `format: 'markdown'` in a schema to create a field
The form input should be a textarea with a "preview" tab that shows the rendered markdown output.

LucianBuzzo:
Fixes #472

change-type: minor


src/extra/Form/markdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+			autofocus,
+			onBlur,
+			onFocus,
+			options,
+			schema,
+			formContext,
+			...inputProps
+		} = this.props;
+
+		const change = ({ target: { value } }: any) => {
```
thgreasi: How about using `React.ChangeEvent<HTMLTextAreaElement>` instead of `any`?

src/extra/Form/markdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+					placeholder="This text supports markdown"
+					readOnly={readonly}
+					disabled={disabled}
+					autoFocus={autofocus}
+					value={value || ''}
+					{...inputProps}
+					onChange={change}
+					onBlur={
+						onBlur &&
+						((event: any) => onBlur(inputProps.id, event.target.value))
```
thgreasi: How about `React.FocusEvent<HTMLTextAreaElement>` instead of `any`?
PS: `onBlur?: FocusEventHandler<T>;` so that should be the correct event.

src/extra/Form/markdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+					}
+					onFocus={
+						onFocus &&
+						((event: any) => onFocus(inputProps.id, event.target.value))
```
thgreasi: How about `React.FocusEvent<HTMLTextAreaElement>` instead of `any`?

src/unstable/components/Form/widgets/TabbedBox.tsx
```diff
@@ -0,0 +1,63 @@
+import * as _ from 'lodash';
+import * as React from 'react';
+import { Box, Button, Flex, Txt } from '../../../../';
+
+interface TabbedBoxProps {
+	tabs: string[];
+	help: JSX.Element;
```
thgreasi: Would it make sense to have this as possibly `undefined`?

src/unstable/components/Form/widgets/TabbedBox.tsx
```diff
@@ -0,0 +1,63 @@
...
+								</Button>
+							);
+						})}
+					</Flex>
+
+					{this.props.help}
+				</Flex>
+
+				{_.map(this.props.children, (child, index) => {
+					if (this.state.show !== index) {
```
thgreasi: How about using `_.filter` or `_.find`?

src/extra/Form/mermaid.tsx
```diff
@@ -41,64 +22,49 @@ class Widget extends React.Component<FormWidgetProps, WidgetState> {
...
+					value={value || ''}
+					{...inputProps}
+					onChange={change}
+					onBlur={
+						onBlur &&
+						((event: any) => onBlur(inputProps.id, event.target.value))
+					}
+					onFocus={
+						onFocus &&
+						((event: any) => onFocus(inputProps.id, event.target.value))
```
thgreasi: We could easily change these as well but definitely not mandatory atm.

src/extra/Form/markdown.tsx
LucianBuzzo: good idea

src/extra/Form/mermaid.tsx
LucianBuzzo: I'll change them in this PR

src/unstable/components/Form/widgets/TabbedBox.tsx
LucianBuzzo: I can sidestep any looping and just use an index accessor here

Issue #502, opened 2018-07-15
Table: data parameter ignores changes after init


Kraku:
**Describe the bug**
When render function gets triggered with new data (props) table remains unchanged.

**To Reproduce**
```js
render() {
    let brightness, active;

    if (this.props.devices.length > 0) {
      brightness = this.props.devices[0].brightness;
      active = this.props.devices[0].active;
    }

    return (
      <div className="ContentTable">
        <p>First element:</p>
        <p>Brightness: {brightness}, Active: {active ? 'yes' : 'nope'}</p>

        <hr />

        <Table columns={this.columns} 
               data={this.props.devices} />
      </div>
    );
  }
```

**Expected behavior**
Table content should be updated when new data comes.

**Screenshots**
Before props update:
<img width="406" alt="screen shot 2018-07-15 at 1 50 34 pm" src="https://user-images.githubusercontent.com/4125286/42733729-e0c2e00a-8836-11e8-87a0-850d1367ef44.png">
After props update:
<img width="381" alt="screen shot 2018-07-15 at 1 50 46 pm" src="https://user-images.githubusercontent.com/4125286/42733728-e0a517be-8836-11e8-96ce-f23e8ba32383.png">

**Specifications (please complete the following information):**
  - Browser: Chrome 67
  - OS: macOS sierra
  - Environment: production

thgreasi:
Hi,
Could you include the whole [table column definition](https://github.com/resin-io-modules/rendition/blob/52bacec277c1cc5da9673114eb5cbcb72d281e9b/typings/rendition.d.ts#L412-L421)?

> When render function gets triggered with new data (props) table remains unchanged.

If the row doesn't get updated, then probably the respective `render()` method of the column definition object doesn't get called when the aforementioned change takes place. Could you confirm that (eg by adding some `console.log`s`)?

The [`render()`](https://github.com/resin-io-modules/rendition/blob/52bacec277c1cc5da9673114eb5cbcb72d281e9b/typings/rendition.d.ts#L419) method is invoked with two parameters, the value that's stored in the [`field`](https://github.com/resin-io-modules/rendition/blob/52bacec277c1cc5da9673114eb5cbcb72d281e9b/typings/rendition.d.ts#L416) property of the row item and the row item itself.
In order to reduce re-renders, each row get's rendered (by the `render()` method) only when the provided item has changed (using `_.isEqual`). As a result, your `render()` method should not rely on other scope variables, but just on its properties.

[...]

PR #505, opened 2018-07-19
Table: Add additional tests to check changes to data prop

LucianBuzzo:
Connects to #502

Change-type: patch


__tests__/components/Table.spec.js
```diff
@@ -361,6 +361,81 @@ describe('Table component', () => {
...
+
+      expect(cellsText).toEqual(dataText)
+    })
+
+    it('should update the table when data changes and a render function is used', () => {
+      const field = 'Name'
+      const cols = [
+        {
+          field,
+          render: (value) => value
```
thgreasi: How about adding some extra text around the value?

__tests__/components/Table.spec.js
```diff
@@ -361,6 +361,81 @@ describe('Table component', () => {
...
+          Name: 'Blastoise',
+          description: 'consecteur adipiscing elit, sed do eiusmod'
+        }
+      ]
+
+      const component = mount(
+        <Table columns={cols} data={data} />
+      )
+
+      data.push({
```
thgreasi: Would it make sense to also check the rendered component before the push as well?
Might be redundant though since we probably already check that elsewhere.

__tests__/components/Table.spec.js
```diff
@@ -361,6 +361,81 @@ describe('Table component', () => {
...
+      })
+
+      component.setProps({ data })
+
+      const cellsText = component.find('[data-display="table-body"] [data-display="table-cell"]')
+        .map(element => element.text())
+
+      const dataText = _.map(data, field)
+
+      expect(cellsText).toEqual(dataText)
```
thgreasi: How about also modifying an existing item? We could check:
* "modifying" the array in an immutable way (like we do in react) to set a new `Name`.
* replacing the array with `_.cloneDeep()` could let us check the respective `_.isEqual()` check in the row component. I'm not sure whether we can actually test that the table did a re-render though. Oh maybe by spying on the `render()` function that we provide. This would surface regressions to that mechanism :)
* We might also want to add a failing test where we modify one of the the array elements in a mutable way. This could surface regressions that would cause unnecessary re-renders of the table.

__tests__/components/Table.spec.js
```diff
@@ -361,6 +361,172 @@ describe('Table component', () => {
...
+      ]
+
+      const component = mount(
+        <Table columns={cols} data={data} />
+      )
+
+      const cellsText = component.find('[data-display="table-body"] [data-display="table-cell"]')
+        .map(element => element.text())
+
+      const dataText = _.map(data, (field))
```
thgreasi: I was initially confused, but I guess that we don't need the parentheses, it seems weird when compared with an actual mapping function.
Or maybe even drop the `field` var since we only use it a few places and IMO it would improve readability.

src/stories/README/Table.md
```diff
@@ -19,6 +19,26 @@ The `columns` property should be an array of objects with the following properti
 | render | <code>(value: any, row: T) => string &#124; number &#124; number &#124; JSX.Element &#124; null</code> | - | Use a custom render function to display the value in each column cell. This function will be called with the value of the `field` provided and the row data (`T`) |
 | sortable | <code>boolean &#124; (a: T, b: T) => number</code> | - | If true, the column will be sortable using an alphanumeric sort, alternatively a function can be provided allowing finer grained control over sorting |
 
+## Notes
+
+For performance reasons table rows are only re-rendered if the row properties
+have changed. For this reason, rows **will not** re-render if data is mutated:
```
thgreasi: How about `rows **will not** re-render if their properties are mutated`?

src/stories/README/Table.md
```diff
@@ -19,6 +19,26 @@ The `columns` property should be an array of objects with the following properti
...
+update (index) {
+  const newData = this.state.data
+  const element = newData[index]
+  newData[index] = Object.assign({}, element, { active: !element.active })
+  this.setState({ data: newData })
+}
+```
+
+Additionally, because of this rendering behaviour the `render` functions in
+`columns` should only render values provided to the render function, as
```
thgreasi: How about `only use values` so that we reduce the repetition of the `render` word a bit?

thgreasi: Maybe we could say something like
> When a `render()` functions are provided, they must act like pure functions with respect to their props. They should only rely on their arguments to generate their output and not use any context variables). If you are using a context variable in your `render()` function, then changes on that variable will not cause a re-render of the component and will not be reflected on the table.

As per https://reactjs.org/docs/components-and-props.html#functional-and-class-components

src/stories/Table.js
```diff
@@ -48,6 +49,36 @@ const columns = [
...
+      PokeDex: _.cloneDeep(PokeDex)
+    }
+
+    this.changeName = this.changeName.bind(this)
+  }
+
+  changeName (e) {
+    const { PokeDex } = this.state
+
+    PokeDex[0] = Object.assign({}, PokeDex[0], { Name: e.target.value })
```
thgreasi: I would like to also see a `slice()` here since it's a best practice, even though the table component will render on each `setState()`.

PR #517, opened 2018-07-23
migrations: Fix filter migration bug

LucianBuzzo:
- Prevent migrations from running on values that have already been
migrated

Change-type: patch


src/migrations.ts
```diff
@@ -234,18 +234,26 @@ const getLegacyFilterSignature = (
 // Converts v3 filters to v4 filters
 export const migrateLegacyFilter = (
 	schema: JSONSchema6,
-	legacyFilter: v3.Filter,
+	legacyFilter: v3.Filter | JSONSchema6,
 ): JSONSchema6 => {
-	if (legacyFilter.name === 'Full text search') {
-		return createFullTextSearchFilter(schema, legacyFilter.value);
+	// Check if this filtr has already been converted, and return immediately if
```
thgreasi: `filtr` => `filter`, or is it a codename like landr :smile:  ?

__tests__/migrations.spec.js
```diff
@@ -223,6 +223,59 @@ describe('migrations', () => {
...
+          }
+        ],
+        operator: 'is true',
+        value: '',
+        id: 'tOYR2EqLICRfS1cS'
+      }
+
+      const migratedFilter = migrations.migrateLegacyFilter(newSchema, oldFilter)
+
+      expect(
```
thgreasi: Could we also add an expectation for the new filter object as well and not just check that re-migrating is not altering the filter?

LucianBuzzo: Good idea

PR #531, opened 2018-07-25
Table: Add property for highlighting a row

LucianBuzzo:
Change-type: minor


src/components/Table.tsx
```diff
@@ -395,11 +407,15 @@ export default class Table<T> extends React.Component<
 					{this.props.tbodyPrefix}
 					{map(this.sortData(data), (row, i) => {
 						const isChecked = this.props.onCheck ? this.isChecked(row) : false;
+						const isHighlighted = this.props.highlightedRow
+							? this.isHighlighted(row)
+							: false;
```
pimterry: It'd be a little nicer to do this check inside isHighlighted. Right now, you have to know that calling it without checking this first might do unpredictable things.

typings/rendition.d.ts
```diff
@@ -433,6 +433,10 @@ declare module 'rendition' {
 		// Optionally provide a key that should be used as a unique identifier for each row
 		rowKey?: keyof T;
 		tbodyPrefix?: JSX.Element | JSX.Element[];
+		// Highlights a row. This property requires that you have provided a
+		// `rowKey` property: the row with a `rowKey` property that matches this
+		// value is highlighted.
+		highlightedRow?: number | string;
```
pimterry: The type here should match the cast for identifier in isHighlighted, since they represent the same thing. I think the `any` there is actually right - your key could be absolutely anything, including objects or arrays or symbols or whatever.

src/stories/README/Table.md
```diff
@@ -59,4 +59,5 @@ cause a re-render of the component and will not be reflected on the table.
 | `rowCheckboxAttributes`    | `object` | - | - | Attributes to pass to the checkbox element used in a row |
 | `rowKey`    | `key of T` | - | - | A field on a row that contains a unique identifier, can help speed up render performance and is required for the `onCheck` property |
 | `toBodyPrefix`    | <code>JSX.element &#124; JSX.Element</code> | - | - | JSX element(s) to display at the top of the table body |
+| `highlightedRow`    | <code>string &#124; number</code> | - | - | Highlights a row. This property requires that you have provided a `rowKey` property: the row with a `rowKey` property that matches this value is highlighted.
```
thgreasi: Heads-up to also update this along with the respective typings.

typings/rendition.d.ts
thgreasi: Maybe `T[keyof T]` even though this doesn't make it a lot better than `any`.

src/components/Table.tsx
```diff
@@ -213,6 +215,16 @@ export default class Table<T> extends React.Component<
...
 	}
 
+	isHighlighted(item: T) {
+		const rowKey = this.props.rowKey;
+		if (!rowKey) {
+			return false;
+		}
+
+		const identifier = item[rowKey];
+		return this.props.highlightedRow === (identifier as any);
```
thgreasi: How about changing this to allow highlighting more than one rows? Sounds symmetric to how how you can have more than one checked rows.

src/components/Table.tsx
LucianBuzzo: good idea

src/components/Table.tsx
LucianBuzzo: Good idea

Issue #538, opened 2018-07-31
ProgressBar: Should allow setting custom colors


thgreasi:
**Is your feature request related to a problem? Please describe.**
We need this for the DeviceProgress componet, so that the ProgressBar color matches that of the device status icon.

**Describe the solution you'd like**
It would be great to add the `color` and `background` to the typings of the ProgressBar, since this already works.

LucianBuzzo:
@thgreasi Can we close this issue now that your PR is merged?

thgreasi:
Absolutely, it seems that I used `Connects-to` instead of `Resolves`...

PR #540, opened 2018-07-31
ProgressBar: Add typings for the color & background

thgreasi:
Connects-to: #538
Change-type: minor
Signed-off-by: Thodoris Greasidis <thodoris@resin.io>
---- Autogenerated Waffleboard Connection: Connects to #538


thgreasi:
Nice catch @dimitrisnl, yea we should.

PR #554, opened 2018-08-15
Button: Fixed buttons alignment

### LINKED ISSUE #552: Buttons not aligning properly
**Describe the bug**
When two buttons are put in a container, they are not aligned properly

**To Reproduce**
Minimum code to reproduce:
```
import { Button } from 'rendition'
 import FaHome from 'react-icons/lib/fa/home' 
...
// Inside render() method:
<div> 
  <Button color='black'><FaHome /></Button> 
  <Button color='black' iconElement={<FaHome />} /> 
</div>
```

**Expected behavior**
Buttons should be top (and bottom) aligned

**Screenshots**
screen shot 2018-08-14 at 19 34 14

**Specifications (please complete the following information):**
  - Browser: whatsmybrowser.org/b/UAABJ3P
  - OS: OSX 10.11.6 El Capitan

bomberaya:
Fixes #552 Buttons not aligning properly
---- Autogenerated Waffleboard Connection: Connects to #552


dimitrisnl:
Hello @bomberaya!

You might want to run `jest --updateSnapshot` in order to update the Button snapshots.

Also, your commit message must include the following tags so that Version Bot can function properly

```
Connects-to: #552
Change-type: patch
```

Thank you for your contribution!

bomberaya:
Hello @dimitrisnl 


I did a `git fetch upstream` and my master branch is up to date with yours.

I did a `npm run jest -- --updateSnapshot` and committed all modified snapshots.

I also tried committing just `Button.spec.js.snap`, and then just the modified shapshots that were previously failing their respsective tests, but to no avail.

I must be missing something here, since I still have three failing tests (in the components that I never modified).

dimitrisnl:
@bomberaya Thank you for bringing this to our attention. I'll open an issue to address this weird snapshot behavior.

For now, I would suggest the following:

- Do a rebase, keeping the initial commit as is but including the appropriate footers tags.
- Include a second commit with all the generated snapshots. 
Something like `tests: Include updated Jest snapshots`. 
Keep in mind that the footer tags are mandatory for each commit.
- Push these two commits, the changes should be these

screen shot 2018-08-17 at 12 50 34 pm

Please ping me if you need any assistance!

dimitrisnl:
@bomberaya Have you checked how this behaviour affects other elements, like `CodeWithCopy`?

LucianBuzzo:
@bomberaya Are you planning to continue work on this PR? If not I'll close the PR, otherwise please let us know. Thanks again for your contribution :smile:

bomberaya:
Hi Lucian,

Unfortunately I'm not able to build the project so I cannot reliably fix
the issue without risking breaking something else.
Sorry about that :(

dimitrisnl:
@bomberaya We could make the `vertical-align` property of the button component accept a specific set of values, falling back to `top`.

Now in order to solve the conflict with `CodeWithCopy`, we can pass the `vAlign` prop here (https://github.com/resin-io-modules/rendition/blob/master/src/components/CodeWithCopy.tsx#L56) with its value set to `middle`. Like `<Button vAlign='middle' />`

How does that sound to you?

dimitrisnl:
@bomberaya Can you expand on not being able to build the project? Could you show me any logs?

bomberaya:
@dimitrisnl Ok, I did it

I had to update my npm (I was at version `5.6.0`, now I'm at `6.4.1`)
and to add a missing script `"storybook": "start-storybook -p 6006",` to `package.json`

Now everything is set up and I'm able to see the changes that I make to the components

bomberaya:
@dimitrisnl Re: vertical-align:

I like your suggestion to add `vAlign` property to `Button` and make it default to `top`.
And set it to `middle` in `CodeWithCopy` component.

Your suggestion would make the `Button` component more versatile in that regard.

The alternate solution to the concrete issue in `CodeWithCopy` could also be
to add `vertical-align: middle;` to `.code-with-copy__copy` class,
since the `Button` in `CodeWithCopy` component is already being styled using this class.

Also, we should probably check all the components whose snapshots are getting updated when running `jest --updateSnapshots` because all those components could be affected by this change.
Right?

dimitrisnl:
@bomberaya That's correct. It would be for the best to make sure that all the dependent components have not been affected by the change.

I would prefer having the `vAlign` property. Let the `Button` component be responsible for the styling instead.

bomberaya:
@dimitrisnl I didn't mean to close this PR, I did an interactive rebase and removed all of my commits because I need to start over.

bomberaya:
Created new PR:
https://github.com/resin-io-modules/rendition/pull/590

Issue #559, opened 2018-08-17
Tables: Allow to add classes to each row


dimitrisnl:
Would like to apply some specific styling in certain rows based on the data each one carries.

PR #563, opened 2018-08-20
table: Add conditional classes to rows

dimitrisnl:
Connects-to: #559
Change-type: minor
---- Autogenerated Waffleboard Connection: Connects to #559


typings/rendition.d.ts
```diff
@@ -449,6 +449,7 @@ declare module 'rendition' {
 		// `rowKey` property: the row with a `rowKey` property that matches this
 		// value is highlighted.
 		highlightedRows?: any;
+		getRowClass: (row: T) => string;
```
LucianBuzzo: This function should return an array of strings, representing each class name.

dimitrisnl: @LucianBuzzo Ah, so I should move the `join` part inside the render method of the row instead.

LucianBuzzo: @dimitrisnl Yep, I think that will be the best approach

Issue #567, opened 2018-08-23
Allow table columns to access the same data field


dimitrisnl:
Accessing the same field will result in

```
vendor.js:71318 Warning: Encountered two children with the same key, `public_key`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

PR #568, opened 2018-08-23
tables: Generate unique key for each column

dimitrisnl:
Connects-to: #567
Change-type: patch

This change is meant to allow the user to select the same field for display, and not have the console complain about having duplicate keys.

### DIFF
--- a/src/components/Table.tsx
+++ b/src/components/Table.tsx
@@ -164,7 +164,7 @@ class TableRow<T> extends React.Component<TableRowProps<T>, {}> {
 						/>
 					</span>
 				)}
-				{map(columns, column => {
+				{map(columns, (column, idx) => {
 					const cellAttributes = isFunction(column.cellAttributes)
 						? column.cellAttributes(data, get(data, column.field))
 						: column.cellAttributes || {};
@@ -176,7 +176,7 @@ class TableRow<T> extends React.Component<TableRowProps<T>, {}> {
 							data-key={keyAttribute}
 							onClick={this.props.onRowClick}
 							{...cellAttributes}
-							key={column.field}
+							key={`${column.field}-${idx}`}
 						>
 							{renderField(data, column)}
 						</a>
@@ -413,10 +413,10 @@ export default class Table<T> extends React.Component<
 								/>
 							</div>
 						)}
-						{map(columns, item => {
+						{map(columns, (item, idx) => {
 							if (item.sortable) {
 								return (
-									<div data-display="table-cell" key={item.field}>
+									<div data-display="table-cell" key={`${item.field}-${idx}`}>
 										<HeaderButton
 											data-field={item.field}
 											plaintext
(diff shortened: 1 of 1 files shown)


dimitrisnl:
NOTE: I'm using prefix+index in case we want to have the functionality of removing/adding columns that would mess with having a simple incrementing counter as key.

src/components/Table.tsx
```diff
@@ -176,7 +176,7 @@ class TableRow<T> extends React.Component<TableRowProps<T>, {}> {
 							data-key={keyAttribute}
 							onClick={this.props.onRowClick}
 							{...cellAttributes}
-							key={column.field}
+							key={`${column.field}-${idx}`}
```
pimterry: We shouldn't really be using the index of the column in the key. This is a little better than using the index alone in the key (i.e. `key={idx}`), but still has the same problems for the cases where you do have two columns on the same field. 

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318 is good background on how this can go wrong and how to fix it in the more general case. With this, if you do somehow swap the position of two columns with the same field, then React won't notice and you might get the wrong state, and when you change the index of a column in general you'll have to rebuild the whole component, even if that's unnecessary. It's not a huge problem, but it is the kind of thing that could catch us out painfully later on.

What is this case where we have two columns for the same data like this? Is there some other property or difference between them we could use?

src/components/Table.tsx
```diff
@@ -438,7 +438,7 @@ export default class Table<T> extends React.Component<
 								);
 							}
 							return (
-								<div data-display="table-cell" key={item.field}>
+								<div data-display="table-cell" key={`${item.field}-${idx}`}>
```
pimterry: (ditto for both of these)

src/components/Table.tsx
dimitrisnl: @pimterry  This is the case that I'm working on. Both these columns have to access the public_key in order to display their content.

```js
{
	label: 'Fingerprint',
	field: 'public_key',
	sortable: false,
	render: (publicKey: ResinSdk.SSHKey['public_key']) => (
		<Txt>{getSSHKeyFingerPrint(publicKey)}</Txt>
	),
},
{
	label: 'Key',
	field: 'public_key',
	sortable: false,
	render: (publicKey: ResinSdk.SSHKey['public_key'], key) => (
		<Flex>
			<PublicKeyField onClick={() => this.handleKeyDisplay(key)}>
				{publicKey}
			</PublicKeyField>
			<TextWithCopy copy={publicKey} showCopyButton="always" />
		</Flex>
	),
},
```

dimitrisnl: Another approach is to expose a `key` property in order to hardcode another value

dimitrisnl:
@pimterry @LucianBuzzo 

I solved it in the dashboard. Created a new interface that extends ResinSdk.SshKeys,  and includes the fingerprint as a property. Can safely close this.

Issue #539, opened 2018-07-31
Theme: update the `statusUpdating` color to Primary.light #5FB3E7


thgreasi:
After discussion with @konmouz 2018-07-31 since the current color is similar enough to the one in the styleguide.

PR #571, opened 2018-08-27
Update the statusUpdating color to Primary.light

amdomanska:
connects-to: #539
Change-type: patch
Signed-off-by: amdomanska <aga@resin.io>

### DIFF
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -1,7 +1,9 @@
+/* used multiple times below */
+const primaryLight = '#5FB3E7';
 const colors = {
 	primary: {
 		main: '#2297DE',
-		light: '#5FB3E7',
+		light: primaryLight,
 		dark: '#1E87C7',
 	},
 	secondary: {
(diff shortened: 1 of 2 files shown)


LucianBuzzo:
@amdomanska The resin-ci commit check is failing because it wants a line break between the commit subject and the footer.
Can you use the `resolves` or `fixes` keyword instead of `connects-to`? That way the attached issue will be automatically closed after merge.

thgreasi:
How using the following commit message:
```

File: src/theme.ts [hunk 1]
```diff
@@ -1,6 +1,8 @@
+/* used multiple times below */
+const primaryMain = '#2297DE';
 const colors = {
 	primary: {
-		main: '#2297DE',
+		main: primaryMain,
```
thgreasi: Was this by accident? The issue & PR title says that we should use `Primary.light` instead of `main`.
cc: @konmouz in case we prefer to go with `Primary.main`

amdomanska: Nope, my mistake o.O Ooops.

thgreasi:
@resin-ci retest

Issue #384, opened 2018-04-11
Table: the last table column needs some padding on the right


thgreasi:
![image](https://user-images.githubusercontent.com/1295829/38623388-37d179f2-3dae-11e8-8d3d-07f93c9da965.png)

PR #572, opened 2018-08-28
Table: Add right padding

amdomanska:
Add right padding and adjust left padding to maintain table layout

Resolves: #384
Change-type: patch
Signed-off-by: amdomanska <aga@resin.io>

Previously:
tableprev
tableprevinspect

Now:
tablenow
tablenowinspect

### DIFF
--- a/src/components/Table.tsx
+++ b/src/components/Table.tsx
@@ -39,7 +39,8 @@ const BaseTable = styled.div`
 			> [data-display='table-cell'] {
 				display: table-cell;
 				text-align: left;
-				padding-left: 40px;
+				padding-left: 20px;
+				padding-right: 20px;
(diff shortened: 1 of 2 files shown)


src/components/Table.tsx
```diff
@@ -61,7 +62,8 @@ const BaseTable = styled.div`
 				text-align: left;
 				padding-top: 14px;
 				padding-bottom: 14px;
-				padding-left: 40px;
+				padding-left: 20px;
+				padding-right: 20px;
```
thgreasi: NIT: since we are now defining all padding sides, how about using the shorthand `padding: 14px 20px;` (and the respective values in the above diff chunk)?

PR #577, opened 2018-08-29
Button: Add type prop to typings

### LINKED ISSUE #576: Button typings should allow the "type" attribute
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

LucianBuzzo:
Fixes #576

Change-type: patch


LucianBuzzo:
@resin-ci retest

PR #582, opened 2018-09-04
Add pager functionality to the Table component

LucianBuzzo:
connects to #581


dimitrisnl:
Awesome, I really like the pager position addition too

PR #584, opened 2018-09-05
table: Remove Box wrapper

dimitrisnl:
Connects-to: #583
Change-type: patch

It would be better for the parent element to decide on the sizing of the Table, than having an extra dom node in between. 
It's also a bit of a problem when combined with the `TableWithCustomColumns` wrapper in the UI.

### DIFF
--- a/src/components/Table.tsx
+++ b/src/components/Table.tsx
@@ -19,7 +19,6 @@ import styled from 'styled-components';
 
 import theme from '../theme';
 import Button from './Button';
-import { Box } from './Grid';
 import Pager from './Pager';
 
 const highlightStyle = `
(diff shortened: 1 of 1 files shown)


src/components/Table.tsx
```diff
@@ -439,7 +438,7 @@ export default class Table<T> extends React.Component<
 		const sortedData = this.sortData(data).slice(lowerBound, upperBound);
 
 		return (
-			<Box {...props}>
+			<React.Fragment>
```
pimterry: Is there a reason we're not using the built-in `<> ... </>` syntax for this? https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html. Should be equivalent, and it's supported in React 16.2+ while it looks like we're using 16.3.1.

dimitrisnl: Was used to write it that way, nothing more. Will update it 👍

Issue #586, opened 2018-09-05
Restore the ability to pass rest props in Table component


dimitrisnl:
The Box wrapper that was removed would accept the rest props giving the ability to style it further. We should include `{...props}` in the Table component.

PR #587, opened 2018-09-05
table: Allow styling using rest props

dimitrisnl:
Connects-to: #586
Change-type: patch

### DIFF
--- a/src/components/Table.tsx
+++ b/src/components/Table.tsx
@@ -451,7 +451,7 @@ export default class Table<T> extends React.Component<
 						/>
 					)}
 
-				<BaseTable>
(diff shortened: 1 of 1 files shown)


dimitrisnl:
I removed `{...props}` from the `Box` wrapper, but it would be nice to have that included.

dimitrisnl:
Would you suggest a more thorough alternative? Maybe including the Pager too? I really want to avoid having a wrapper component.

thgreasi:
I would say that we should cut a patch release with this as is, since the previous change stopped us from styling the `<Table>`, and we can reiterate for the `<Pager>`s.
cc: @LucianBuzzo

PR #590, opened 2018-09-12
button: Fix buttons alignment

bomberaya:
Fixes the alignment of buttons in row

Connects-to: #552
Change-type: patch


bomberaya:
@dimitrisnl 

> We could make the vertical-align property of the button component accept a specific set of values, falling back to top.

I went for a string property instead, because I didn't want to constrain the user to a subset of allowed values, and didn't want to copy all of the values supported by the CSS vertical-align property. For example, it can also contain a percentage value.

PR #596, opened 2018-09-19
tests: Remove .only call from Table tests

LucianBuzzo:
Change-type: patch


LucianBuzzo:
Closing in favor of https://github.com/resin-io-modules/rendition/pull/597

PR #597, opened 2018-09-19
Table: Defensively program against null or undefined data values

### LINKED ISSUE #589: Table crashes if data prop is null

LucianBuzzo:
Fixes #589

Change-type: patch

### DIFF
--- a/src/components/Table.tsx
+++ b/src/components/Table.tsx
@@ -286,7 +286,7 @@ export default class Table<T> extends React.Component<
 	}
 
 	setRowSelection = (selectedRows: T[]): void => {
-		const rowKey = this.props.rowKey;
+		const { rowKey, data } = this.props;
 
 		if (!rowKey) {
 			return;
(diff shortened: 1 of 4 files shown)


src/components/Table.tsx
```diff
@@ -286,9 +286,9 @@ export default class Table<T> extends React.Component<
 	}
 
 	setRowSelection = (selectedRows: T[]): void => {
-		const rowKey = this.props.rowKey;
+		const { rowKey, data } = this.props;
 
-		if (!rowKey) {
+		if (!rowKey || !data) {
```
thgreasi: How about moving this a bit lower, so that we can always do `setRowSelection([])`?
I'm thinking about the case that the table at some point had data and some of the rows were selected and at some point the property get's null. Even in that case the user should be able to try to clear the selections w/o issues.

src/components/Table.tsx
```diff
@@ -301,14 +301,20 @@ export default class Table<T> extends React.Component<
...
 	toggleAllChecked = () => {
+		const { data } = this.props;
+
+		if (!data) {
+			return;
+		}
+
 		const allChecked = !this.state.allChecked;
-		const checkedItems = allChecked ? this.props.data.slice() : [];
+		const checkedItems = allChecked ? data.slice() : [];
```
thgreasi: How about doing a `this.props.data || [])?
We might also want to have `allChecked = allChecked && !_.isEmpty(this.props.data)` to avoid an awkward situation that the all check marked is ticked but there are no data at all.

src/components/Table.tsx
LucianBuzzo: Good idea, I'll add a test case for this as well

src/components/Table.tsx
LucianBuzzo: Good idea, as above, I'll add a test case

Issue #602, opened 2018-09-20
Tooltip crashes when run in sandbox


amdomanska:
Right after tooltip param is added it works fine but any next change (even not connected to the component with tooltip, it might be just adding some text above or below) makes it crash with TypeError Cannot read property 'clientHeight' of null.

1. start with: https://codesandbox.io/s/3yp8zrwz75
2. check that tooltip works fine
3. remove line <h2> or make any other changes
4. tooltip crashes
5. refresh the page to make tooltip work again

**Expected behavior**
tooltip works fine all the time

sradevski:
@thgreasi @LucianBuzzo is this still relevant? Do we need to do something or can we close it?

PR #603, opened 2018-09-20
Tooltip: Fix from crashing when the tooltipElement gets cleared

thgreasi:
Resolves: #602
Change-type: patch
Signed-off-by: Thodoris Greasidis <thodoris@resin.io>

### DIFF
--- a/src/tooltips.tsx
+++ b/src/tooltips.tsx
@@ -154,10 +154,6 @@ class TooltipComponent extends React.Component<{}, TooltipComponentState> {
 		boundingRect.top += window.scrollY;
 		boundingRect.left += window.scrollX;
 
-		if (!this.tooltipElement) {
-			return;
-		}
-
 		if (!placement || placement === 'top') {
 			return {
 				top:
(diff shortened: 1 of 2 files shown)


src/tooltips.tsx
```diff
@@ -152,52 +154,80 @@ class TooltipComponent extends React.Component<{}, TooltipComponentState> {
 		boundingRect.top += window.scrollY;
 		boundingRect.left += window.scrollX;
 
-		const { placement, containerStyle, innerStyle, arrowStyle } = options;
-
-		// Set the contents of the tooltip using `ìnnerText` and adjust the styles
-		// now, so that the height can be properly calculated
-		if (this.tooltipElementInner) {
-			this.tooltipElementInner.innerText = tooltipText;
+		if (!this.tooltipElement) {
```
LucianBuzzo: We already check for this, so the additional check is redundant. I think we need to create a repeatable failing test case and then patch around it. This solution feels too much like guesswork.

thgreasi: Oh! Nice spot! It seems that I duplicated this during the refactor.

thgreasi: The fact is though that I couldn't reproduce what we saw in sandbox under normal circumstances :(

PR #607, opened 2018-09-20
docs: Compile separate component README files into main README

### LINKED ISSUE #598: Compile readme files into main readme
We should improve the discoverability of component documentation and one way to do this would be to compile all of the separate component README files into the main repo README. The individual readme files should still appear alongside the stories.

LucianBuzzo:
Fixes #598

Change-type: patch


scripts/generateDocs.js
```diff
@@ -0,0 +1,26 @@
...
+const path = require('path')
+
+const README_DIR = './src/stories/README';
+const REPLACE_TARGET = '{% COMPONENT_DOCS %}'
+
+const fileNames = fs.readdirSync('./src/stories/README')
+
+const docs = []
+
+for (const name of fileNames) {
```
thgreasi: NIT: How about using a `.map()` like we do right after?

scripts/generateDocs.js
```diff
@@ -0,0 +1,26 @@
...
+  const file = fs.readFileSync(path.join(README_DIR, name), 'utf8')
+  // Reduce size of markdown headers to better fit the main README format
+  docs.push(file.replace(/(^|\n)#/g, '$&##'))
+}
+
+const content = fileNames.map((file) => {
+  const name = file.replace('.md', '')
```
thgreasi: We might want to run prettier against this (even if it's just once and not as part of our build process) since in some places we use `;` while we don't in other.

LucianBuzzo: good idea

scripts/generateDocs.js
LucianBuzzo: yep good idea

PR #619, opened 2018-09-27
Introducing the Toggle switch button

alanhortz:
Here an addition to the lib, a toggle switch button.


alanhortz:
Todo : Create a separate branch for the feature

PR #620, opened 2018-09-27
Introducing the Toggle swith

alanhortz:


LucianBuzzo:
Hi @alanhortz Thankyou very much for your contribution. At resin, we are using rendition as the foundation for our web dashboard ( https://dashboard.resinstaging.io ), which already has an opinionated design for toggle switches:

screen shot 2018-10-02 at 09 31 15

At the moment this is using the Font Awesome "toggle-on" icon, although I really like the animated implementation you've presented in this PR. If we could update this PR to match this pre-existing toggle switch design that would be great :+1:

LucianBuzzo:
@resin-ci retest

src/components/Toggle.js
```diff
@@ -0,0 +1,130 @@
...
+    transition: all 0.3s cubic-bezier(0.175, 0.885, 0.32, 1.275),
+      padding 0.3s ease, margin 0.3s ease;
+    box-shadow: 0 0 0 2px rgba(0, 0, 0, 0.1), 0 4px 0 rgba(0, 0, 0, 0.08);
+  }
+  &:active::after {
+    padding-right: 1.6em;
+  }
+  ${props =>
+    props.on &&
+    `
```
LucianBuzzo: I'd prefer to see long style blocks like this encapsulated in a separate variable or function to keep the code readable.

src/components/Toggle.js
```diff
@@ -0,0 +1,130 @@
...
+    if (this.props.on !== nextProps.on) {
+      this.setState({ on: nextProps.on })
+    }
+  }
+
+  render () {
+    const { disabled, emphazised, ...props } = this.props
+    return (
+      <span>
+        <Input type='checkbox' checked={this.state.on} onChange={() => {}} />
```
LucianBuzzo: Why does this input need to be here?

LucianBuzzo: Why does this element have a noop `onChange` callback?

src/components/Toggle.js
```diff
@@ -0,0 +1,130 @@
...
+        <Button
+          {...props}
```
LucianBuzzo: Because extra props aren't applied to the top-level element of this component it can lead to some strange behaviour. For example if you're setting padding using the styled-system shorthand ( `p={3}` ) the padding is applied *inside* the `<span>` element. Additional props should be applied to the outermost/top-level element to ensure consistent behaviour across components.

typings/rendition.d.ts
```diff
@@ -346,6 +346,13 @@ declare module 'rendition' {
 
 	class Search extends RenderableElementWithProps<SearchProps, any> {}
 
+	class Toggle extends RenderableElementWithProps<
+		DefaultProps & {
+			brand: JSX.Element;
```
LucianBuzzo: What is the `brand` property? It seems to be completely unused

src/components/Toggle.js
alanhortz: Good idea, I'll do it

src/components/Toggle.js
alanhortz: To enhance accessibility, but this can be removed in favour of the attributes  role="checkbox" and aria-checked={true/false} on the button

alanhortz: To let the button control the state (referring to the aria-label="Toggle" on the button). Without this onChange handler, the checkbox would not have been updated. 

Again, we can remove the input see my [comment](https://github.com/resin-io-modules/rendition/pull/620/files/4727df91355229f1ed44a3919ef9a01e435ab625#r221955494) above

src/components/Toggle.js
alanhortz: Ok, got it !

typings/rendition.d.ts
alanhortz: You right not used, stupid copy/paste ...

LucianBuzzo:
Closed pending a comprehensive design review

PR #621, opened 2018-10-02
Form: Add unique class name for every field

LucianBuzzo:
This change makes it much easier to deterministically test forms using
tools like Puppeteer.

Change-type: minor


LucianBuzzo:
@thgreasi Good point, I'll update this PR to use classes instead, since the desired effect is the same

Issue #637, opened 2018-10-31
Tooltips can disrupt layout


LucianBuzzo:
If you have a long page layout, then trigger a tooltip at the bottom of the page, then reduce the length of the page (by removing elements) the page will stay the same length even though the tooltip and the element it was attached to are no longer visible.

Issue #377, opened 2018-03-31
Tooltip stays open after its attached element disappears (v4.0.0-beta2)


Shou:
If the element a tooltip is attached to disappears while the tooltip is open, the tooltip doesn't fade out as it should.

PR #638, opened 2018-10-31
637 tooltip layout

LucianBuzzo:
This fixes two issues:
- Tooltips disrupting layout #637
- Tooltips persisting after the bound element is removed #377

### DIFF
--- a/src/tooltips.tsx
+++ b/src/tooltips.tsx
@@ -201,7 +201,13 @@ class TooltipComponent extends React.Component<{}, TooltipComponentState> {
 	}
 
 	hide(): void {
-		this.setState({ show: false });
+		this.setState({
+			show: false,
+			coordinates: {
+				top: 0,
+				left: 0,
+			},
+		});
 	}
 
 	render() {
--- a/src/tooltips.tsx
+++ b/src/tooltips.tsx
@@ -13,6 +13,25 @@ const TARGET_OFFSET = 3;
 // The time in ms to show a tooltip after being triggered by a click event
 const CLICK_TIMEOUT = 1000;
 
+const createOnRemoveObserver = (
+	element: Node,
+	onDetachCallback: () => void,
+) => {
+	const observer = new MutationObserver(() => {
+		if (!document.contains(element)) {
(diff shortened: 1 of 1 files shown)


src/tooltips.tsx
```diff
@@ -120,6 +146,7 @@ interface TooltipComponentState {
 class TooltipComponent extends React.Component<{}, TooltipComponentState> {
 	private tooltipElement: HTMLDivElement;
 	private tooltipElementInner: HTMLDivElement;
+	private observer: MutationObserver;
```
thgreasi: This will hold no object initially, so let's type it
```suggestion
	private observer: MutationObserver | undefined;
```

src/tooltips.tsx
```diff
@@ -201,7 +245,13 @@ class TooltipComponent extends React.Component<{}, TooltipComponentState> {
 	}
 
 	hide(): void {
-		this.setState({ show: false });
+		this.setState({
```
thgreasi: We should also probably disconnect & de-reference the observer, since we no longer care about the mutations on the target and let the target be GCed if removed from the DOM.
```suggestion
		if (this.observer) {
			this.observer.disconnect();
			this.observer = undefined;
		}
		this.setState({
```
^ GH does a bad diff on suggestions that just prepend some code :(

src/tooltips.tsx
```diff
@@ -13,6 +13,32 @@ const TARGET_OFFSET = 3;
 // The time in ms to show a tooltip after being triggered by a click event
 const CLICK_TIMEOUT = 1000;
 
+const isDetached = (el: Node): boolean => {
+	if (el.parentNode === document) {
```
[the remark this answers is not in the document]
LucianBuzzo: Awesome! I can't believe I missed this :smile:

src/tooltips.tsx
LucianBuzzo: :+1:

src/tooltips.tsx
LucianBuzzo: Good idea

Issue #326, opened 2018-01-25
ProgressBar should have a default color


LucianBuzzo:
By default, the progress bar inner element doesn't have a background color.

thgreasi:
You mean without having to set a primary/etc property?

LucianBuzzo:
Yes indeed

PR #656, opened 2018-11-16
Add a default background color for ProgressBar

sradevski:
Connects-to: #326
Change-type: minor


nazrhom:
@resin-ci retest

LucianBuzzo:
@sradevski You'll need to update the jest snapshots and screenshots I think

sradevski:
> @sradevski You'll need to update the jest snapshots and screenshots I think

@LucianBuzzo I did run the screenshot generation and tests, but nothing changed in the snapshots. I simply added a default value to be `primary`, but that didn't have any visual effect on Storybook, even after removing the `primary` prop from one of the invocations (as `primary` is the default now).

Issue #661, opened 2018-11-19
DropdownButton understands array of child items as one item


karaxuna:
**Describe the bug**
When you have children containing item + array of items, DropdownButton understands it as 2 children.

**To Reproduce**
`
<DropdownButton>
    <div>1</div>
    {[2, 3].map(n => (
        <div key={n}>{n}</div>
    ))}
</DropdownButton>
`

**Expected behavior**
number of items should be 3.

**Screenshots**
<img width="351" alt="tmp" src="https://user-images.githubusercontent.com/2152815/48690035-5ff99e00-ebe6-11e8-8bb5-6b0ecce986ab.png">

PR #662, opened 2018-11-19
DropDownButton should map through children correctly

karaxuna:
Change-type: minor
Connects-to: #661

### DIFF
--- a/__tests__/components/DropDownButton.spec.js
+++ b/__tests__/components/DropDownButton.spec.js
@@ -18,3 +18,25 @@ test('DropDownButton renders correctly', () => {
   let tree = component.toJSON()
   expect(tree).toMatchSnapshot()
 })
+
+test('Opened DropDownButton renders correctly', () => {
+  const component = renderer.create(
+    <Provider>
+      <DropDownButton primary label={<div>DropDown</div>}>
+        <div>Item</div>
+        {[...Array(3).keys()].map((index) => (
+          <div key={index}>Item</div>
+        ))}
+      </DropDownButton>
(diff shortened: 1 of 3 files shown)


karaxuna:
@thgreasi I changed test to:

```javascript
/* globals expect, test */
import React from 'react'
import renderer from 'react-test-renderer'
import Provider from '../../src/components/Provider'
import DropDownButton from '../../src/components/DropDownButton'

test('DropDownButton renders correctly', () => {
  const component = renderer.create(
    <Provider>
      <DropDownButton primary label={<div>DropDown</div>}>
        <div>Item</div>
        {Array(3).map((_, index) => (
          <div key={index}>Item</div>
        ))}
      </DropDownButton>
    </Provider>
  )
  let tree = component.toJSON()
  expect(tree).toMatchSnapshot()
})
```
But tests still pass, even if I write `Item3` in place of `Item` it does not care.

LucianBuzzo:
@karaxuna This is because of the list not being rendered until the button is clicked. If you look at the existing snapshot, there are no menu items in it https://github.com/balena-io-modules/rendition/blob/master/__tests__/components/__snapshots__/DropDownButton.spec.js.snap

thgreasi:
@karaxuna  Let's keep the existing test case and write a new one that simulates a click to the dropdown and checks the children rendered with enzyme (like we do elsewhere) or maybe with https://reactjs.org/docs/test-utils.html .

nazrhom:
@resin-ci retest

thgreasi:
@resin-ci retest

PR #669, opened 2018-11-20
Move remaining JS files to Typescript, second iteration

sradevski:
Moved DropDownButton, ProgressBar, Navbar, Heading, Gauge, and
DeviceStatusGauge to Typescript. Updated storybook to render
bordered items when dropdown list is expanded.

Also contains a bugfix for Navbar not rendering correctly on mobile (caught by Typescript).

Connects-to: #44
Change-type: patch


src/components/DeviceStatusGauge.tsx
```diff
@@ -0,0 +1,73 @@
+import filter = require('lodash/filter');
+import * as React from 'react';
+import { compose } from 'recompose';
+// @ts-ignore
```
LucianBuzzo: What is this for?

LucianBuzzo: @sradevski Would it be simple enough to add a small typings file here https://github.com/balena-io-modules/rendition/tree/master/src/typings ?
I'd prefer to do that, even if we are going to remove this component anyway

sradevski:
@resin-ci retest

Issue #674, opened 2018-11-21
Buttons remain with "focus" styling even when they shouldn't (after a click)


sradevski:
**Describe the bug**
When a button or a dropdown is clicked, they have the "hover" and "focus" styling which is wrong at times. That is only desirable when a user is navigating using the keyboard, or when a dropdown is currently in an expanded situation.

**To Reproduce**
1. Click on any button or dropdown in the storybook using version 4.37.0

See: https://github.com/balena-io-modules/rendition/pull/667/files#diff-f0c64fac293f2a778cfbf16458498428R92

**Screenshots**
After a click:
![screen shot 2018-11-21 at 17 27 23](https://user-images.githubusercontent.com/4820812/48854666-bca9b400-edb2-11e8-9fc2-a13bc1494a06.png)

PR #675, opened 2018-11-21
Button: Fix focus state regression

thgreasi:
Resolves: #674
Change-type: patch


thgreasi:
@resin-ci retest

PR #677, opened 2018-11-26
Prevent `grid-styled` from installing styled-components v4

LucianBuzzo:
This prevents a bug where different major versions of styled-components
would be loaded, causing errors

Change-type: patch


LucianBuzzo:
This is ineffective at stopping styled-components v4 from being installed when rendition is used in another app, as `package-lock.json` is ignored by npm.

sradevski:
@LucianBuzzo would something like [this](https://stackoverflow.com/questions/53162324/installing-specific-version-of-dependencies-for-a-package-inside-that-package-in) work? It seems this might be the solution.

Issue #678, opened 2018-11-26
DropDownButton: Menu's max-height should be customizable


thgreasi:
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/AtRbdZcRlfobItxfl1y6UQ889Z9
See: https://github.com/balena-io-modules/rendition/issues/410
See: https://github.com/balena-io-modules/rendition/pull/411

PR #679, opened 2018-11-27
Add maxHeight prop to DropdownButton

sradevski:
Connects-to: #678
Change-type: minor


src/components/DropDownButton.tsx
```diff
@@ -184,6 +185,11 @@ class DropDownButton extends React.Component<
 						alignRight={alignRight}
 						onClick={this.toggle}
 						minWidth={`${this.state.minWidth}px`}
+						maxHeight={
+							isNumber(this.props.maxHeight)
```
LucianBuzzo: You can use the `px()` function from `utils.ts` here https://github.com/balena-io-modules/rendition/blob/master/src/utils.ts#L40

src/stories/DropDownButton.js
```diff
@@ -36,6 +36,18 @@ storiesOf('Core/DropDownButton', module)
             <div>Item</div>
             <div>Item</div>
           </DropDownButton>
+          <DropDownButton
+            mx={2}
+            maxHeight={120}
+            label={<div>Fixed Height</div>}
```
LucianBuzzo: Fixed height and max height are subtly different, perhaps changing this to `Max Height` is better?

src/components/DropDownButton.tsx
sradevski: Oh cool, thanks!

src/stories/DropDownButton.js
sradevski: True, I will change it.

src/components/DropDownButton.tsx
sradevski: @LucianBuzzo Done, have a look at it when you find the time.

PR #680, opened 2018-11-27
Prevent `grid-styled` from installing styled-components v4

LucianBuzzo:
This prevents a bug where different major versions of styled-components
would be loaded, causing errors

Change-type: patch


LucianBuzzo:
@resin-ci retest

LucianBuzzo:
@resin-ci retest

Issue #401, opened 2018-04-17
Should we remove the hover effects from disabled inputs?


thgreasi:
The mouseover styles of inputs makes them seem clickable.
See: https://github.com/resin-io/resin-ui/pull/1938#discussion_r181057839

thgreasi:
@konmouz what do you think on this>

LucianBuzzo:
Yes definitely

PR #681, opened 2018-11-27
Apply hover style only on enabled inputs

sradevski:
Connects-to: #401
Change-type: minor


sradevski:
@resin-ci retest

Issue #685, opened 2018-12-04
Add column filtering option to Table


sradevski:
There are times when we want to allow the user to choose which columns should be displayed in a table. This should be the same as it is in the balena dashboard.

sradevski:
This is a bit complicated for now in order to match with some internal implementations, so we can tackle this in the next major bump with breaking changes.

PR #686, opened 2018-12-05
Add column selector dropdown to Table

sradevski:
It also enables local storage for the table preferences
(currently only stores the visible columns preferences)

Connects-to: #685
Change-type: minor


sradevski:
The table component will probably need a refactor using the  _Compound Components_ pattern, but the aim of this was to generalize the column selector used in the dashboard.

Issue #691, opened 2018-12-10
Add callback on sort and option to pass default sort options


sradevski:
The sorting is currently not persisted to local storage. Storing it is useful when users wish to keep the same type of sorting consistently without redoing it over and over again.

LucianBuzzo:
@sradevski I think storage solutions are out of scope for this module.

sradevski:
@LucianBuzzo In that case we need to have a callback on sort, as well as provide the option to pass sorting options. Does that sound like a better alternative?

LucianBuzzo:
@sradevski yes definitely

sradevski:
@LucianBuzzo  I am thinking between these two:
1. Allow to pass default sort options that are used only when the table is mounted, but subsequent renders will use the internal table state
2. Allow to pass sort options that will control the sorting throughout the entire table lifetime.

I personally prefer number 1 as it removes the need to pass sorting props around in the users' code. What do you think?

PR #692, opened 2018-12-10
Add callback on sort and option to pass default sort options

sradevski:
Also moved the TableRow in its own file.

Connects-to: #691
Change-type: minor


src/components/Table/TableRow.tsx
```diff
@@ -0,0 +1,91 @@
+import get = require('lodash/get');
+import isEqual = require('lodash/isEqual');
+import isFunction = require('lodash/isFunction');
+import map = require('lodash/map');
+import * as React from 'react';
+import { TableColumn } from 'rendition';
+
+/*
+ * Get the value specified by the `field` value
```
LucianBuzzo: Can you create a separate PR/Commit that extracts `TableRow` into its own file?

src/components/Table/index.tsx
```diff
@@ -103,103 +103,15 @@ const HeaderButton = styled(Button)`
...
-}
-
 export default class Table<T> extends React.Component<
 	TableProps<T>,
 	{
 		allChecked: boolean;
-		reverse: boolean;
+		sort: {
+			reverse: boolean;
+			columnName: null | keyof T;
```
LucianBuzzo: `field` might be a better key than `columnName`, I have no strong opinions on this though.

src/components/Table/index.tsx
```diff
@@ -212,11 +124,15 @@ export default class Table<T> extends React.Component<
 			);
 		}
 
+		const sortState = this.props.defaultSort || {
```
LucianBuzzo: This feels a bit weird, can you not just make the sort controllable through a `sort` property? This `defaultSort` property is limited in functionality.
If you maintained the same constructor logic you could set the state on update with a `componentWillUpdate` lifecycle method, eg:
```js
componentWillUpdate(nextProps) {
  if (!isEqual(nextProps.sort, this.props.sort)) {
    this.setState({ sort: nextProps.sort })
  }
}
```

src/components/Table/index.tsx
sradevski: I feel like `field` is quite generic, but I do agree that it is better to be consistent with the rest of the usage.

src/components/Table/index.tsx
sradevski: It is actually not recommended to do that (and componentWillUpdate is depricated in following versions of React): https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions. I do agree that it gives more control like that, so I am up for doing it as it will be easier to change the behavior later on without breaking the API.

src/components/Table/TableRow.tsx
sradevski: I personally think that refactorings don't necessarily need a separate issue and PR as it is something we are doing continuously (or we should be), unless it is a large change that might affect functionality. I can do it of course, but even though it is a bit cleaner, I feel it is unnecessary additional work. Do you agree?

src/components/Table/index.tsx
LucianBuzzo: Reading that link it appears that the recommendation is to use a completely controlled component, or use `getDerivedStateFromProps` right? They don't seem to explore the idea of *optionally* controlled components, but it seems reasonable that `getDerivedStateFromProps` should do what we are looking for

src/components/Table/TableRow.tsx
LucianBuzzo: @sradevski At the very least, separating the refactor into a different commit is good practice. A separate PR would make this much easier to review, right now it's very hard to see the difference between what was refactored and what was changed. Additionally, the refactor appears to be unrelated to the functionality change here.

src/components/Table/index.tsx
sradevski: `getDerivedStateFromProps` does allow us to do what we want, they just recommend to avoid it if possible as it can create race conditions with internal state management and give you a result you don't expect (as described in their example). For now we can use `componentWillReceiveProps` and once we upgrade to React 16.3+ we can refactor away from the depricated lifecycle methods.

src/components/Table/TableRow.tsx
sradevski: It is really just a copy-paste from the Table component into a new file, but I do agree it is a bit difficult to see when reviewing. I will separate it into its own commit.

PR #693, opened 2018-12-11
Change xterms addon import  path

cyplo:
Changed XTerm library's addon path used, from `dist` to `lib`, as
recommended in XTerm [examples](https://github.com/xtermjs/xterm.js) - this may cause problems when importing -
seen production build errors for reconfix playground. Tested by
importing into reconfix playground via a local path. All other existing rendition  tests
seem happy, also checked the storybook manually to see if everything looks good in the terminal page.

### DIFF
--- a/src/components/Terminal.tsx
+++ b/src/components/Terminal.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { TerminalProps } from 'rendition';
 import styled from 'styled-components';
 import { ITerminalOptions, Terminal as Xterm } from 'xterm';
-import { fit as fitTerm } from 'xterm/dist/addons/fit/fit';
+import { fit as fitTerm } from 'xterm/lib/addons/fit/fit';
 import Theme from '../theme';
 import { Box } from './Grid';
 


cyplo:
Error can be viewed here: https://circleci.com/gh/balena-io-playground/reconfix-play/127

cyplo:
Would be interesting to know how to prevent such things in the future - but no idea how to add a test/build stage for that for rendition @LucianBuzzo - any thoughts here ?

LucianBuzzo:
@cyplo Can you reproduce this anywhere but your circleCI build?

cyplo:
@LucianBuzzo - yes, locally, removed `node_modules` and ran `npm install` and `npm run build` - the and my local build both use newest node v10 - would that make a difference ?

cyplo:
@LucianBuzzo - sure thing, will try making minimal repro, not sure myself why this is happening, is probably related to my recent changes around adding `jest` and `enzyme` to the playground - will play around with this

cyplo:
See #694 for repro @LucianBuzzo

cyplo:
Would be good to maybe port the test from that issue into this PR - what do you think @LucianBuzzo ?

Issue #709, opened 2018-12-23
Banner.js still mentions resin.io


lurch:
src/stories/Banner.js still mentions resin.io instead of balena.io

lurch:
P.S. see also https://github.com/balena-io/docs/issues/994 :wink:

PR #710, opened 2018-12-24
Change Resin.io to balena in Banner

sradevski:
Connects-to: #709
Change-type: patch


sradevski:
@dimitrisnl True, I didn't pay too much attention to it, thanks!

dimitrisnl:
@thgreasi 	Yes, this is the case.

sradevski:
@dimitrisnl @thgreasi So to clarify, It should be uppercase in the sentence (`Balena brings the benefits`), but it should be lowercase in the header?

PR #717, opened 2019-01-11
Import lodash methods independently to reduce bundle size

thgreasi:
Didn't have an impact on the storybook bundle size, since mermaid
imports lodash as a whole. On the other hand, the dashboard bundle
doesn't include mermaid at all, which means that we are already
tree-shaking it away and we will benefit from the lodash change as well.

Also adds linter rules to prevent re-importing as a whole.

Resolves: #145
Change-type: patch
See: https://github.com/knsv/mermaid/blob/a4992963b3873ec7a566ba91b92596d2a360e3b9/src/diagrams/git/gitGraphAst.js#L1
See: https://github.com/knsv/mermaid/blob/a4992963b3873ec7a566ba91b92596d2a360e3b9/src/diagrams/git/gitGraphRenderer.js#L1

### DIFF
--- a/src/stories/Table.js
+++ b/src/stories/Table.js
@@ -1,4 +1,4 @@
-import * as _ from 'lodash'
+import * as cloneDeep from 'lodash/cloneDeep'
 import * as React from 'react'
 import { storiesOf } from '@storybook/react'
 import { action } from '@storybook/addon-actions'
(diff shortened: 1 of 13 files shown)


src/stories/Table.js
```diff
@@ -1,4 +1,4 @@
-import * as _ from 'lodash'
+import * as cloneDeep from 'lodash/cloneDeep'
```
thgreasi: This was really weird :confused: :alien: :satellite: 
Even though in all the rest js files we do imports like
```
import map from 'lodash/map'
```
in here this didn't work and the storybook was failing.
I guess that this is b/c the storybook uses its own transpilation & bundling configuration which differs from what we use to generate the dist.

thgreasi:
@resin-ci retest

thgreasi:
@resin-ci retest

Issue #711, opened 2018-12-25
Make Modal's padding responsive


sradevski:
Currently the horizontal padding is fixed to `50px`, which many times causes overflow on mobile. It is better to reduce it to around `16-20px` for the smallest breakpoint.

PR #720, opened 2019-01-14
Reduce the modal padding for the smallest breakpoint

sradevski:
Connects-to: #711
Change-type: patch

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -1,12 +1,17 @@
 import assign = require('lodash/assign');
 import * as React from 'react';
 import { ModalProps } from 'rendition';
-import styled, { injectGlobal } from 'styled-components';
+import styled, { injectGlobal, withTheme } from 'styled-components';
 import { stopPropagation } from '../utils';
+import { px } from '../utils';
 import Button from './Button';
 import Fixed from './Fixed';
 import { Box, Flex } from './Grid';
 import Txt from './Txt';
+
+interface ThemedModalProps extends ModalProps {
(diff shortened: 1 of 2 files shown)


src/components/Modal.tsx
```diff
@@ -117,6 +116,7 @@ class Modal extends React.Component<ModalProps, any> {
 				<ModalBackdrop z={8888} bg="rgba(0,0,0,0.4)" top right bottom left />
 				<ModalSizer style={props.containerStyle}>
 					<ModalPanel
+						p={['30px 20px', '30px 50px']}
```
LucianBuzzo: It would be great to use the sizings from the default theme ( https://github.com/balena-io-modules/rendition/blob/master/src/theme.ts#L97 ). 
`16px` all the way around looks quite good:

screen shot 2019-01-14 at 09 17 35

LucianBuzzo: @sradevski My bad, I mean to link to the `space` value https://github.com/balena-io-modules/rendition/blob/master/src/theme.ts#L87
You can use the `withTheme` helper from `styled-components` to pass the theme as a property to the component. Checkout the `Button` source for inspiration, its quite straightforward https://github.com/balena-io-modules/rendition/blob/master/src/components/Button.tsx#L245

sradevski: @LucianBuzzo oh yeah, the `space` variable. It's easy to do it, I was just wondering what approach would be more suitable for this PR. In that case, I will update the rest of the values of the Modal as well, if applicable.

PR #722, opened 2019-01-14
Link: fallback to using href as text content

### LINKED ISSUE #721: Link component should use href as text if no children are provided
If an href is provided to the `Link` component, but no child text is present, then the href value should be displayed:
```
<Link href="https://google.com" />
```

LucianBuzzo:
Fixes #721

Change-type: minor

### DIFF
--- a/__tests__/components/Link.spec.js
+++ b/__tests__/components/Link.spec.js
@@ -1,15 +1,29 @@
-/* globals expect, test */
+/* globals expect, describe, it */
+import { mount } from 'enzyme'
 import React from 'react'
 import renderer from 'react-test-renderer'
 import Provider from '../../src/components/Provider'
 import Link from '../../src/components/Link'
 
-test('Link renders correctly', () => {
-  const component = renderer.create(
-    <Provider>
-      <Link>Link</Link>
-    </Provider>
-  )
-  let tree = component.toJSON()
-  expect(tree).toMatchSnapshot()
+describe('Link component', () => {
+  it('Should match the stored snapshot', () => {
+    const component = renderer.create(
+      <Provider>
+        <Link>Link</Link>
+      </Provider>
+    )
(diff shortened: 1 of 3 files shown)


__tests__/components/Link.spec.js
```diff
@@ -1,15 +1,31 @@
...
+    const url = 'https://google.com'
+    const component = mount(
+      <Provider>
+        <Link href={url} />
+      </Provider>
+    )
+
+    console.log(component.html())
+
+    expect(component.find(`a [href="${url}"]`)).toHaveLength(1)
```
sradevski: NIT: I think this doesn't really test the spec, as it simply checks if there is an anchor with a `href` set to `url`. I think it should be something like ```expect(component.find(`a [href="${url}"]`)[0].text()).toEqual(url)```

LucianBuzzo: @sradevski Yep you're totally right! This was a clear mistake, I'll fix it.

__tests__/components/Link.spec.js
```diff
@@ -1,15 +1,31 @@
...
+
+  it('Should display the href prop if a child is not provided', () => {
+    const url = 'https://google.com'
+    const component = mount(
+      <Provider>
+        <Link href={url} />
+      </Provider>
+    )
+
+    console.log(component.html())
```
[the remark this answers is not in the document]
LucianBuzzo: Yep, I'll fix this

sradevski:
@LucianBuzzo I think this one is good to go, would you mind rebasing and merging it? Just so it doesn't get outdated and forgotten :smile:

LucianBuzzo:
@sradevski I'm focusing on https://github.com/balena-io-modules/rendition/pull/688 and then I'll update and merge this one

LucianBuzzo:
@resin-ci retest

Issue #724, opened 2019-01-23
Use styled(Component) as opposed to Component.extend


dimitrisnl:
Currently, it yields the following warning

```
Warning: The "extend" API will be removed in the upcoming v4.0 release. Use styled(StyledComponent) instead. You can find more information here: https://github.com/styled-components/styled-components/issues/1546
```

Example:
https://github.com/balena-io-modules/rendition/blob/39206fbd52d1f0fde7ab44c3caea42d82496cd2b/src/components/Alert.tsx#L86

sradevski:
Done in #800

PR #725, opened 2019-01-23
components: Fix styled components' deprecation warning

dimitrisnl:
Connects-to: #724
Change-type: patch


dimitrisnl:
edit: my HEAD is a bit outdated. Fixing.

dimitrisnl:
Hmm, this is particularly strange. There should be no visual changes. Investigating.

sradevski:
@dimitrisnl I just saw that you updated the screenshot, but it seems something broke. I am assuming the place it broke is where we use `withComponent` to render things as a different tag from the original one. Let me know how things go (and re-add reviewers after a change so we get a notification) 😸

dimitrisnl:
@sradevski 	I'm not very sure what's wrong. There shouldn't be any changes in the snapshots. 

Checking the `withComponent` instances but nothing wrong there. Any ideas?

sradevski:
@dimitrisnl As I assumed, there was a problem with `withComponent` and the styles not being applied correctly to the anchor. I did some changes and pushed it, although didn't do any testing, so you might want to check it out a bit more before merging.

dimitrisnl:
Sure, thanks

sradevski:
@dimitrisnl Looking at the changed files, we probably don't want to pass all the props down to the HTML component (look at how the theme and other props are passed to the native HTML element).

dimitrisnl:
This is getting a bit out of hand. The purpose was to change the deprecated syntax while extending a component. It shouldn't have spawned that many problems.

Will tweak it for couple minutes, and drop it if there is no progress. We shouldn't waste time on such a minor issue.

sradevski:
@dimitrisnl I do agree, but when we decide to move to styled-components v4 we will have to do this anyway. It might be less relevant if we decide to use a different library though.

LucianBuzzo:
@dimitrisnl Now that we've update to styled-components v4 can this PR be closed?

Issue #728, opened 2019-01-24
Fix failing PR checks


JackuB:
This needs fixing

![screen shot 2019-01-24 at 12 41 18](https://user-images.githubusercontent.com/1788727/51675951-7bfa3100-1fd5-11e9-99c9-d907ff670725.png)

Possibly related: #575 as I saw travis failing too on occasion - haven't found anything else mentioning these failed checks.

JackuB:
Confirmed it's caused by node v6 incompatibility

![screen shot 2019-01-24 at 13 08 20](https://user-images.githubusercontent.com/1788727/51677219-2889e200-1fd9-11e9-9a53-7f8630688b7a.png)

JackuB:
![screen shot 2019-02-04 at 15 01 30](https://user-images.githubusercontent.com/1788727/52212894-e0e34000-288d-11e9-8cbf-65dd24ee3687.png)

With [balena-io/ci-images#6](https://github.com/balena-io/ci-images/pull/6) merged, tests passed. Closing

PR #729, opened 2019-01-24
Fix Node 6 tests

JackuB:
connects-to: #728


thgreasi:
@JackuB also add @nazrhom as a reviewer to this, once it's ready

src/asRendition.tsx
```diff
@@ -10,7 +10,12 @@ import { Tooltips } from './tooltips';
 
 const tooltip = new Tooltips();
 
-const prop = oneOfType([number, string, arrayOf(oneOfType([number, string]))]);
+const prop = oneOfType([
+	number,
+	string,
+	arrayOf(number.isRequired),
```
JackuB: Otherwise `arrayOf` is nullable.

JackuB:
So after reading code of our concourse tooling for half a day, it was actually simpler to make tests pass on Node 6 again.

screen shot 2019-01-24 at 21 01 13

But, I need a TypeScript help on how to make this PR better - I assume fixing things with `(X as any)` is not ideal :)

src/components/Table/index.tsx
```diff
@@ -273,7 +272,7 @@ export default class Table<T> extends React.Component<
 		const isChecked = !this.isChecked(item);
 		const checkedItems = isChecked
 			? this.state.checkedItems.concat(item)
-			: reject(this.state.checkedItems, { [rowKey]: identifier });
+			: this.state.checkedItems.filter(value => value[rowKey] !== identifier);
```
JackuB: Lodash's `reject` return was messing the type of `checkedItems`

src/components/Table/index.tsx
```diff
@@ -189,7 +188,7 @@ export default class Table<T> extends React.Component<
 			return data;
 		}
 
-		const column = find(this.props.columns, { field: sort.field });
+		const column: any = find(this.props.columns, { field: sort.field });
```
JackuB: Otherwise I was getting `sortable` is not on `number`, from 

```js
if (isFunction(column.sortable)) {
```

src/components/Table/TableRow.tsx
```diff
@@ -64,7 +64,7 @@ export class TableRow<T> extends React.PureComponent<TableRowProps<T>, {}> {
 				)}
 				{map(columns, column => {
 					const cellAttributes = isFunction(column.cellAttributes)
-						? column.cellAttributes(data, get(data, column.field))
+						? column.cellAttributes(data, get(data, (column as any).field))
```
JackuB: Couldn't get `TableColumn<T>` to work here

sradevski:
@JackuB To be honest, I think we can just drop Node 6 from running in the first place. It doesn't really test much (except if Node 6 can be used for development or not), and its EOL is in April anyhow, so there is no need to (use it) support it. This should also reduce some unnecessary stress on the CI. @LucianBuzzo @thgreasi what do you think?

typings/rendition.d.ts
```diff
@@ -466,7 +466,7 @@ declare module 'rendition' {
 		cellAttributes?:
 			| React.AnchorHTMLAttributes<HTMLAnchorElement>
 			| ((value: any, row: T) => React.AnchorHTMLAttributes<HTMLAnchorElement>);
-		field: keyof T;
+		field: string | number | undefined;
```
thgreasi: Why is this a required change for node 6 to work?
I would prefer to avoid this if we can, since it would mean that passing a field name that's not part of the object will longer result a typing error.
If newer TS keyof semantics about Symbols is messing with it, you could use
```suggestion
		field: keyof T & string;
```

JackuB:
@thgreasi have you removed and reinstalled node modules with node 6? I managed to reproduce the failing tests locally.

thgreasi:
@JackuB I did `nvm uninstall 6 && nvm install  6 && nvm use 6`. Isn't that enough?

thgreasi:
Oh, I also updated npm, maybe that has something to do with it.

JackuB:
@thgreasi Good point there - through trial and error I found the issue is with npm@<5. Haven't found yet what's causing this, but easiest would be to just update npm in our CI images. I'll take a look into that

JackuB:
@cyplo yup PR is ready here https://github.com/balena-io/ci-images/pull/6 - should be discussed on arch call. That should be today afaik.

PR #732, opened 2019-01-30
Do not fail to run the tests after previous crash on the same machine.

cyplo:
In a situation where the previous screenshot test run was interrupted
the next run would fail because of the clash on the docker container
name. This PR makes the tests try to remove the container if it exists
before running the tests .

### DIFF
--- a/scripts/test-screenshots.sh
+++ b/scripts/test-screenshots.sh
@@ -1,22 +1,29 @@
 #!/bin/sh
+set -e
 
 # Exit if docker isn't running
 docker version > /dev/null || exit 1
 
+container_name="rendition-screenshot"
+
 echo "Starting Docker container"
-docker run --name rendition-screenshot -dit buildkite/puppeteer:v1.8.0 || exit 1
+set +e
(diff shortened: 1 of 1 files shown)


cyplo:
Heya, this seems failing on node6 only - typescript errors.
What would be the best course of action here ?
Do we want to support node6 ? 
If not - what about me creating a PR disabling this on CI ?
If yes - let me know if you want me to dig into this.
@LucianBuzzo @sradevski - what do you think ??

cyplo:
See also https://github.com/balena-io-modules/rendition/pull/729 - seems to be the same problem ?

sradevski:
@cyplo The node6 tests are not required, so feel free to merge this, no need to wait for anything. 
Regarding the CI running node6, I would personally prefer just disabling it as it doesn't really test much anyway, but I would like some input from others as well. cc @thgreasi @JackuB @karaxuna @dimitrisnl @LucianBuzzo

cyplo:
Thanks ! :)

PR #738, opened 2019-02-07
Migrate from balena-cdsl to jellyschema

zrzka:
This PR just migrates rendition to `jellyschema` package. `jellyschema` (renamed `balena-cdsl`) provides new API and adds validation support.

You can learn more about `jellyschema` from [these tests](https://github.com/balena-io-modules/jellyschema/tree/v0.11.0/node/tests/src).

Change-type: patch
Signed-off-by: Robert Vojta <robert@balena.io>


cyplo:
@JackuB @sradevski - the node6 tests seem to be still failing - could you take a look ?

sradevski:
@cyplo The error seems to be in the `balena-temen` package, probably because it uses some node8+ feature, but I believe it is not a blocker for this PR.

cyplo:
@zrzka see the comment about temen above ^

JackuB:
Added some details about temen compatibility here https://github.com/balena-io/ci-images/issues/5#issuecomment-461461436

zrzka:
The problem is Web Assembly in our packages. It's not just `balena-temen` package, but it's also `jellyschema`. Other packages like these are coming (`reconfix`, ...). And `reconfix` for example will be used in many places (Etcher, resin-cli, ...), thus, these packages will affect other stuff as well.

I can be wrong, but I think it's safe to say that we need at least Node 8. Newer is better. All these WASM packages are tested with Node 10. I just tried with 6 & 7 and it failed. Works like a charm with 8 and newer.

You can check how these beasts are build here:

* https://github.com/balena-io-modules/jellyschema/tree/v0.11.0/ci
* https://github.com/balena-io-modules/jellyschema/blob/v0.11.0/.nvmrc

Especially this:

* https://github.com/balena-io-modules/jellyschema/blob/v0.11.0/ci/build-wasm.sh
* https://github.com/rustwasm/wasm-pack

All these `.js` files inside our NPM packages are emitted by `wasm-pack` (not us directly). We can modify them, but it will be a nightmare. Even now the build script is crazy and I can't imagine writing our own or patching it with `sed` in a more complicated way. Then we will loose the ability to use newer `wasm-pack` releases.

So, what can I do here. I'm affraid that I can't do anything. Perhaps, one small thing, I can put `8` into `.nvmrc` file in all these repositories to test it with Node 8, not Node 10. Is it worth it? Or are we going to change CI pipeline and drop Node 6? Or ...? Dunno. I'm happy to change / help with whatever you need guys, but I'm affraid that Node 8 it is.

PR #747, opened 2019-02-13
Form: Add alpha support for patternProperties

LucianBuzzo:
This change adds support to the Form component for the JSON schema
"patternProperties" keyword. This is an early implementation and needs
further work to be production ready.
I also took this opportunity to tidy up JSDOM setup, by defining the
document global and other useful values in one place.

Change-type: minor

### DIFF
--- a/__tests__/components/Form.spec.js
+++ b/__tests__/components/Form.spec.js
@@ -405,4 +405,31 @@ describe('Form component', () => {
       component.unmount()
     })
   })
+
+  describe('patternProperties', () => {
+    it('should allow you to add new key value pairs', () => {
+      const schema = {
+        type: 'object',
+        patternProperties: {
+          '^[0-9]+$': {
+            'title': 'Rule',
+            'type': 'string'
+          }
+        }
+      }
+
+      const component = mount(
+        <Provider>
+          <Form
+            schema={schema}
+          />
+        </Provider>
+      )
+
+      const valueInput = component.find('input.rendition-form-pattern-properties__value-field')
+      expect(valueInput.length).toEqual(1)
+      const keyInput = component.find('.rendition-form-pattern-properties__key-field').first()
+      expect(keyInput.length).toEqual(1)
+    })
+  })
 })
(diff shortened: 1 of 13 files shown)


__tests__/components/Form.spec.js
```diff
@@ -405,4 +405,31 @@ describe('Form component', () => {
       component.unmount()
     })
   })
+
+  describe('patternProperties', () => {
```
cyplo: are we concerned with removal of properties in this PR ?

__tests__/components/Form.spec.js
```diff
@@ -405,4 +405,31 @@ describe('Form component', () => {
...
+
+      const component = mount(
+        <Provider>
+          <Form
+            schema={schema}
+          />
+        </Provider>
+      )
+
+      const valueInput = component.find('input.rendition-form-pattern-properties__value-field')
```
cyplo: would having an additional test, that the object produced from the form is in the right shape, help ?

__tests__/components/Form.spec.js
LucianBuzzo: Could you expand on your question? Its a little ambiguous

__tests__/components/Form.spec.js
LucianBuzzo: I spent some time wrestling with Enzyme trying to get tests to run correctly, but kept encountering a very strange bug. In the end I decided to continue with the PR and come back to unit tests as I didn't want to block and alpha release of this functionality.

cyplo: alrighty then :)

__tests__/components/Form.spec.js
cyplo: looking at the test title `should allow you to add new key value pairs` - would an opposite need to be tested ? e.g. a removal of previously added property key and value by the user of the UI ?

LucianBuzzo: @cyplo The functionality for removing key/value pairs exists and works, but isn't tested yet

PR #753, opened 2019-02-15
Form: Allow markdown in description fields

LucianBuzzo:
Change-type: minor

### DIFF
--- a/__tests__/components/Form.spec.js
+++ b/__tests__/components/Form.spec.js
@@ -387,6 +387,36 @@ describe('Form component', () => {
     })
   })
 
+  describe('description keyword', () => {
+    it('should support markdown', () => {
+      const descriptionSchema = {
+        type: 'object',
+        properties: {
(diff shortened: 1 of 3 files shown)


JackuB:
Is there any chance this component could ingest user data? I see we are sanitizing markdown, but still, you can inject images, new elements etc. If so, we should slap a big warning on it or make the markdown an optional switch.

LucianBuzzo:
@JackuB What is your concern? Are you worried this could break layout?

JackuB:
@LucianBuzzo one concern is definitely XSS, the other would be unintended formatting. For example if a project/team/… name could get rendered there, then from my `team_project_env` naming scheme is <code>team<em>project</em>env</code>.

LucianBuzzo:
@JackuB The input is sanitized and we use GitHub flavoured markdown, so your `team_project_env` example would be rendered correctly.

PR #754, opened 2019-02-18
Fix top-level array handling in Form/JellyForm

### LINKED ISSUE #743: JellyForm blows up on custom string formats inside arrays
Given:
```
version: 1
properties:
  - array:
      type: array
      items:
        type: dnsmasq-address
```
When clicking `add item` JellyForm emits an exception:
```
Error: No widget "dnsmasq-address" for type "string"
```

Link to playground [here](https://balena-io-playground.github.io/reconfix-play/index.html?yaml=dmVyc2lvbjogMQpwcm9wZXJ0aWVzOgogIC0gYXJyYXk6CiAgICAgIHR5cGU6IGFycmF5CiAgICAgIGl0ZW1zOgogICAgICAgIHR5cGU6IGRuc21hc3EtYWRkcmVzcw%3D%3D)

cyplo:
When rendering a Form/JellyForm based on JSONSchema where its top-level
type would be an array we tried to put in a default value of `{}` which
crashed rjsf. Using better defaults now (`[]`) in these cases.

* Not sure if this is idiomatic TS/JS code (e.g. `static`)
* Not sure if this change needs to be replicated in Form and JellyForm - I followed the example in the existing code there - it was setting the default of `{}` in both places. Should contructors call each other instead ?

Fixes #743

### DIFF
--- a/src/unstable/components/Form/index.tsx
+++ b/src/unstable/components/Form/index.tsx
@@ -71,8 +71,8 @@ export default class FormHOC extends React.Component<
 		super(props);
 
 		this.state = {
-			value: this.props.value || {},
-			schema: this.parseSchema(this.props.schema),
+			value: props.value || (props.schema.type === 'array' ? [] : {}),
+			schema: this.parseSchema(props.schema),
 		};
 	}
 
--- a/__tests__/components/Form.spec.js
+++ b/__tests__/components/Form.spec.js
@@ -216,6 +216,28 @@ describe('Form component', () => {
     })
   })
 
+  describe('top-level array fields', () => {
+    const arraySchema = {
+      'items': {
+        'type': 'string'
+      },
+      'type': 'array'
+    }
+
+    const callback = sinon.spy()
+
+    it('should not crash on render', () => {
+      mount(
+        <Provider>
+          <Form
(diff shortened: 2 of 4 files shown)


src/unstable/components/Form/index.tsx
```diff
@@ -71,11 +71,20 @@ export default class FormHOC extends React.Component<
...
 
 		this.state = {
-			value: this.props.value || {},
+			value: this.props.value || FormHOC.defaultValueForSchema(props.schema),
+
 			schema: this.parseSchema(this.props.schema),
 		};
 	}
 
+	static defaultValueForSchema(schema: JSONSchema6) {
```
sradevski: Generally speaking, it is bad practice to expose static methods from React components. You can either have this as a utility function in some `utils` file, or, honestly speaking, you can just duplicate it in both places, as `JellyForm` shouldn't be aware of its child `Form` IMO. Also, you can rewrite it as 
```
	static defaultValueForSchema(schema: JSONSchema6) {
   return schema.type === 'array' ? [] : {}
	}
```

cyplo:
how about now @sradevski ?

cyplo:
@resin-ci retest

PR #755, opened 2019-02-25
Don't crash on adding data to top-level array in JellyForm

cyplo:
Makes sure that the form data we pass around is relevant to the
top-level schema type.


cyplo:
@resin-ci retest

cyplo:
@resin-ci retest

PR #756, opened 2019-02-25
Form: Add markdown support for help fields

LucianBuzzo:
Change-type: minor


__tests__/components/__snapshots__/Form.spec.js.snap
```diff
@@ -41,11 +41,9 @@ exports[`Form component should match the stored snapshot 1`] = `
               value="Bulbasaur"
             />
             <div />
-            <div />
```
cyplo: interesting, was it unbalanced before ?

LucianBuzzo: @cyplo This is the JSX equivalent of `<div></div>` which I believe was created by the default help text renderer.

cyplo: ah yes, misread the tag, thought it was `</div>` :D fair enough :)

PR #758, opened 2019-02-26
Form: make objects collapsible by default

LucianBuzzo:
Change-type: minor

### DIFF
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -1,3 +1,5 @@
+import ChevronDown from 'react-icons/lib/fa/chevron-down';
+import ChevronUp from 'react-icons/lib/fa/chevron-up';
 import { px } from 'styled-system';
 
 const primary = '#00AEEF';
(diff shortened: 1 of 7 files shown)


cyplo:
Interesting ! From the client point of view - we'd like the ability to define the `collapsed/-ible` component in the `jsonschema` definition. The definition is for `a thing that gets us data` - would that map directly to the `Form` component in rendition ? Would that be `accordion` + `something that wraps Form and accordion` ? Good question :)

cyplo:
@sradevski @LucianBuzzo - any news on this one ?

LucianBuzzo:
I need to update this to use the `Accordian` from Grommet https://v2.grommet.io/accordion

LucianBuzzo:
collapsibleform

src/unstable/components/Form/templates/ObjectFieldTemplate.tsx
```diff
@@ -1,26 +1,65 @@
...
+	const { DescriptionField, idSchema } = props;
+
+	// The `activeIndex` and `onActive` props are used to make the accordion
+	// a controlled input, allowing the accordion to default to being open
+	const [active, setActive] = React.useState(true);
+
+	const toggleActive = () => setActive(!active);
+
+	const title =
+		props.schema.title && (props.uiSchema['ui:title'] || props.title)
```
sradevski: I would rather have a "ui:collapsible" or a similar option, as we might not always want to have a collapsible form whenever there is a title.

src/theme.ts
```diff
@@ -246,6 +248,18 @@ export default {
 		extend: 'border: none;',
 		color: colors.primary.main,
 	},
+	accordion: {
+		border: {
+			color: colors.quartenary.main,
+		},
+		icons: {
+			// collapse: <ChevronUp />,
```
sradevski: remove the comments.

Issue #761, opened 2019-02-28
Add support for oneOf in filter


karaxuna:
**Is your feature request related to a problem? Please describe.**
I want to filter records by slug, but want to show user friendly name in filter dropdown.

**Describe the solution you'd like**
```
nationality: {
    title: 'Nationality',
    oneOf: [{
        title: 'South African',
        const: 'south_african',
    }, {
        title: 'New Zealand',
        const: 'new_zealand',
    }],
}
```

*Before submitting this issue, please check that this issue is not a duplicate. If there is another issue describing the same problem or feature please add your information to the existing issue's comments.*

sradevski:
This was done AFAIK, closing.

PR #762, opened 2019-02-28
Filter: Add oneOf support

karaxuna:
Change-type: minor
Connects-to: #761


JackuB:
+1 on the missing tests tho

src/components/DataTypes/enum.tsx
```diff
@@ -5,74 +5,16 @@ import * as React from 'react';
...
-
-	return {
-		field,
-		operator,
-		value,
-	};
-};
+import { decodeFilter, OneOfFilter, operators, OperatorSlug } from './oneOf';
 
 export const createFilter = (
```
sradevski: Does this mean there is no more `EnumFilter`? I don't understand why we removed the `EnumFilter` type here and why we are exporting `oneOf` functions from the `enum` file. I think there should be a helper file that contains commonalities between the two.

src/components/DataTypes/oneOf.tsx
```diff
@@ -0,0 +1,138 @@
...
+				const?: any;
+			};
+		};
+	};
+}
+
+export const decodeFilter = (
+	filter: OneOfFilter,
+): {
+	field: string;
```
sradevski: Define these as an interface, it's a bit cleaner.

src/components/DataTypes/enum.tsx
karaxuna: @sradevski I did it because oneOf is more generic form of enum. Enum is oneOf with same const and title. I'll probably revert changes in enum file.

sradevski: @karaxuna If that's the case, then we should somehow merge the two in a clear manner. I still see references to Enum in several places, so we either keep the two separate for this PR and merge later, or do it cleanly now, up to you :smile:

karaxuna:
@LucianBuzzo Added tests. It's all about one functionality, so I don't think I can split it into different PRs.

Issue #763, opened 2019-03-01
Make modal submit on Enter and cancel on Escape keys pressed


sradevski:
It is expected from users that the modal will close by pressing the escape key (and resolve the `cancel` prop if it exists) and submit and close by pressing the enter key.

PR #764, opened 2019-03-01
Make Modal close on Enter and Escape keys press

sradevski:
Connects-to: #763
Change-type: minor

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -80,6 +80,8 @@ class Modal extends React.Component<ThemedModalProps, any> {
 		if (!Modal.mountedCount) {
 			document.body.classList.add(bodyNoOverflowClass);
 		}
+
+		window.document.addEventListener('keydown', this.handleKeyDown);
 		Modal.mountedCount++;
 	}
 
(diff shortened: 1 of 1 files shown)


src/components/Modal.tsx
```diff
@@ -88,8 +90,32 @@ class Modal extends React.Component<ThemedModalProps, any> {
...
 		}
+
+		window.document.removeEventListener('keydown', this.handleKeyDown);
 	}
 
+	handleKeyDown = (e: KeyboardEvent) => {
+		if (this.props.keyDownHandlers && this.props.keyDownHandlers[e.which]) {
+			this.props.keyDownHandlers[e.which](e);
+		} else {
+			console.log(e.defaultPrevented);
```
karaxuna: Forgot to remove this.

LucianBuzzo:
@sradevski Why is there a `keyDownHandlers` prop?

sradevski:
@LucianBuzzo In case you want to override the default behavior (not close the dialog on escape, not submit on enter), or bind other keys with different behavior.

LucianBuzzo:
@sradevski Is this a requirement right now? I don't think that being able to override the default behaviour is a good thing

sradevski:
@LucianBuzzo I just wanted to provide an escape hatch in case there is a specific place where the default behavior is not desired. Anyhow, I agree we can leave this for later on if required and I can remove it from this PR.

sradevski:
@thgreasi I did, but then I did another change that broke input! Nice catch!

sradevski:
@resin-ci retest

Issue #765, opened 2019-03-01
Close dropdown when clicked outside of it


sradevski:
The dropdown should be closed if clicked outside of its context (the dropdown button or its list).

PR #766, opened 2019-03-01
Close dropdown when clicked outside of it

sradevski:
Connects-to: #765
Change-type: minor

### DIFF
--- a/src/components/DropDownButton.tsx
+++ b/src/components/DropDownButton.tsx
@@ -125,6 +125,8 @@ class DropDownButton extends React.Component<
 	DropDownButtonProps,
 	DropDownButtonState
 > {
+	dropdownNode: HTMLDivElement;
+
 	constructor(props: DropDownButtonProps) {
 		super(props);
 		this.state = {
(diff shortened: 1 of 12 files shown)


JackuB:
Wondering what change in this PR is causing this failure `https://ci.balena-dev.com/builds/355466` (all Node versions) https://ci.balena-dev.com/builds/355466

And also that `worker ip-10-1-78-147-1550745925 disappeared while trying to reach it` is not triggering a timeout failure.

src/components/DropDownButton.tsx
```diff
@@ -133,6 +135,22 @@ class DropDownButton extends React.Component<
 		};
 	}
 
+	componentDidMount() {
```
JackuB: So according to this https://reactjs.org/docs/react-component.html#setstate

> `setState()` will always lead to a re-render unless shouldComponentUpdate() returns false. 

And this handler is mounted even for closed dropdowns, if I'm reading it correctly. So if I have 10 closed dropdowns on page, click on something unrelated, it will cause 10 re-renders?

sradevski: @JackuB You are right, it will cause re-renders. Not a big issue in this case, but I agree that it can be improved by attaching the handlers when a component opens, and removing them when it closes.

cyplo:
Heya, it is cool and I would love to have this on th reconfix playground, so thank you for that :)
A question - is there a way we could have a test for this behavior ? E.g. render a dropdown, open it and then close it this way ?

sradevski:
@cyplo It would be possible using a web driver (Cypress, Selenium, etc.), but there is no such support in storybook AFAIK.

cyplo:
I thought maybe it would be possible to have something like we have right now with the `*spec` tests, that are using react test renderer and enzyme, or would that not work ?

src/components/DropDownButton.tsx
```diff
@@ -133,6 +135,24 @@ class DropDownButton extends React.Component<
...
+		window.document.addEventListener('click', this.handleOutsideClick);
```
thgreasi: NIT: How about registering a single listener for all dropdowns or attach the listener only when the dropdown opens?

sradevski: @thgreasi If it is a single listener it will complicate things a bit and break isolation IMO. Regarding attaching the event listener when it opens, I was also thinking if it's better like that or not, but I sort of leaned towards registering listeners once. I am reconsidering it now, as it might be a bit "cleaner".

PR #767, opened 2019-03-04
Fix formula evaluation inside arrays

### LINKED ISSUE #759: Nested formula evaluation for arrays does not seem to work
Using [this example](https://balena-io-playground.github.io/reconfix-play/index.html?yaml=eJyVUk1rwzAMvfdXCHYYjDZl11wKo2wrjK0QxtjRbZVGa2J7ltwQ2I%2Bf3TS0SWEfOkXP1vN7TxGSElO4UyVq9ZLB2uictt4pIaOB1wVWarRHx6FN4XZknbHohJDTEcAENEpt3C42saSle25R0tsOb2yAlXOqmR0hEqy4mwPwmj49LvpgfIE2vZb5DOirOZ8572NdwSNtNqi7kWYMH54FcuNgyg2HdyfBu8Z1NM7Tm%2FTUJbQJuVTWyyGVC%2Ba5gRqhVlpADFRqhyAFMeyJaVViBKVA8IxuBktnVmpVNqCNjEMI1xy%2F4mHuy2TA3cbG4k5JdlUc7KQgzuPgKHiqfKnCoA9Wk5gYfEH29PqwuH8f3K2UtYF8GFeIVkmRQi%2BEiz0Mp47bf6N7gixbzP%2Fu5hcZNeWU9DYfJVjFHH6zH2Uswx1bOMXDkFoxHcX%2F5UwY196RNInl3TfsXuKK) - the `id` field does not appear in the data.

cyplo:
Fixes #759 - a problem where if a formula definition would be inside of an array, it would not get evaluated properly.

### Background 
More info:
We preprocess the data before sending it to temen to evaluate - we put the formulas from the schema in the relevant places inside the data. This was done for objects but not for arrays of objects. The code in this PR basically goes through the data object and makes sure that for every item in every array on any level we invoke the substitution of existing data for the formula. 

While the code works and passes the tests - I'm creating a draft PR because I would like to discuss the approach here first before merging this (also a cool opportunity to test the new github draft PR stuff;) Once this is discussed and any changed to the code are done I will clean up the commits list and make versionist happy.

### Discussion:

I added more tests on the UI level to test this, however I feel that these are of not the best type of a test to use here - I'm testing data transformation logic through the UI, which makes the test more brittle and less clear. 

I was thinking what to do with the above and have two ideas here:
1) In the future - it may be good to move all this `data + schema => data_with_formulas` logic inside reconfix maybe, so that JellyForm only calls these functions ? Or maybe there is an existing functionality that we could reuse here ? cc @zrzka here

2) Extract out the code that does the formula transformations to the data to a separate place and test separately. UI tests would only test UI behavior and unit tests for the pure functions would test the data transformation. Do we want to do this in this PR maybe ? Or a PR that would be merged before this one ? 

In terms of implementation I have some doubts as well:
- Is there a simpler way to do the substitution ? When this code was used to substitute key-value pairs in objects it used lodash's `set` function and it was pretty much just an invocation of the said function in a loop. I tried to find a way to do the same for data in an array using lodash or similar libraries but couldn't find anything, but may have missed that - is there a more idiomatic way to do this ? The current implementation as in this PR feels quite naive.

### DIFF
--- a/__tests__/components/JellyForm.formulas.spec.js
+++ b/__tests__/components/JellyForm.formulas.spec.js
@@ -34,6 +34,7 @@ describe('JellyForm formula runner should', () => {
   describe('substitute formula text from schema', () => {
     it('on the the top level', () => {
       const schema = {
+        type: 'object',
         properties: {
           time: {
             type: 'string',
@@ -49,14 +50,20 @@ describe('JellyForm formula runner should', () => {
       expect(ran).not.toEqual(data)
       expect(ran.time).toBeTruthy()
     })
-    it('when wrapped in object', () => {
+
+    it('when wrapped in an object', () => {
       const schema = {
+        type: 'object',
         properties: {
           wrapper: {
-            time: {
-              type: 'string',
-              $$formula: 'NOW()'
-            }}
+            type: 'object',
+            properties: {
+              time: {
+                type: 'string',
+                $$formula: 'NOW()'
+              }
+            }
+          }
         }
       }
       const data = {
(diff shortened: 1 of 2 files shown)


LucianBuzzo:
@cyplo Ideally, I would like to see all the formula evaluation logic separated out. A good starting point would be to send a PR that extracts the logic to a separate file and adds unit tests for it, Then we can add the array handling logic in a follow up PR.
The interaction between jellyschema and temen parsing (formula evaluation) needs to be extracted from rendition as it is not specific to React and would be needed in other places, for example a CLI implementation.

cyplo:
Let me do the prefactoring then, thanks :)

cyplo:
Heyo, I rebased this on top of master, after #768 got merged. Ported the tests over to the lower level, instead of them being on the UI level.

Let me know if there is some more idiomatic/builtin/library way of doing what I've done here. Thanks !

__tests__/components/JellyForm.formulas.spec.js
```diff
@@ -49,7 +49,8 @@ describe('JellyForm formula runner should', () => {
       expect(ran).not.toEqual(data)
       expect(ran.time).toBeTruthy()
     })
-    it('when wrapped in object', () => {
+
+    it('when wrapped in an object', () => {
       const schema = {
         properties: {
           wrapper: {
```
LucianBuzzo: I know this is pre-existing, but is this supposed to be a valid JSON schema?

cyplo: good point, will go through all of them to validate

cyplo: generally yes, should be valid json schema, that includes our formula extensions

cyplo: I made sure that all the schemas render properly in rjsf

Issue #760, opened 2019-02-27
Remove momentjs to reduce bundle size


dimitrisnl:
We don't use moment extensively. Can try date-fs.

![screen shot 2019-02-27 at 7 33 48 pm](https://user-images.githubusercontent.com/4951004/53510448-abb1c280-3ac6-11e9-8dfb-4dd06087b0aa.png)

![screen shot 2019-02-27 at 7 34 52 pm](https://user-images.githubusercontent.com/4951004/53510585-064b1e80-3ac7-11e9-9d3b-301b3e647612.png)

thoughts @sradevski 	@LucianBuzzo

dimitrisnl:
Or just drop it. We can write a function ourselves if we only want this operation

```js
// Normalize a timestamp to a RFC3339 timestamp, which is required for JSON
// schema.
// https://momentjs.com/docs/#default-format
const normalizeDateTime = memoize((timestamp: string) =>
	moment(timestamp)
		.utc()
		.format(),
);
```

sradevski:
@dimitrisnl It sounds good to me, if we only use it in just one place. No need to bloat rendition unnecessarily. Let's see what Lucian thinks.

LucianBuzzo:
@dimitrisnl @sradevski I'm happy to drop moment for something smaller here.

JackuB:
Moment is still used in Mermaid, so keeping this issue opened.

sradevski:
We are not using moment directly in rendition, and moment will only be imported if users use mermaid. I will close this as it is not really actionable. We can investigate bundle size later on if it proves to be a problem.

PR #771, opened 2019-03-12
Use native Date instead of Momentjs

JackuB:
Change-type: patch
Connects-to: #760 

First tested the previous Moment.js implementation and then changed it to Date()

### DIFF
--- a/src/components/DataTypes/date-time-helpers.tsx
+++ b/src/components/DataTypes/date-time-helpers.tsx
@@ -1,0 +1,31 @@
+import memoize = require('lodash/memoize');
+import * as moment from 'moment';
+
+const pad = (n: number) => (n < 10 ? `0${n}` : n);
+
+export const getDefaultDate = (): string => {
+	const now = new Date();
+	const day = pad(now.getDate());
+	const month = pad(now.getMonth() + 1);
+	return `${now.getFullYear()}-${month}-${day}T10:00`;
+};
+
+// Normalize a timestamp to a RFC3339 timestamp, which is required for JSON
+// schema.
+export const normalizeDateTimeNew = memoize((timestamp: string) => {
+	const d = new Date(timestamp);
+	if (isNaN(d.getTime())) {
+		return 'Invalid date';
+	}
+	return `${d.getUTCFullYear()}-${pad(d.getUTCMonth() + 1)}-${pad(
+		d.getUTCDate(),
+	)}T${pad(d.getUTCHours())}:${pad(d.getUTCMinutes())}:${pad(
+		d.getUTCSeconds(),
+	)}Z`;
+});
+
+export const normalizeDateTime = memoize((timestamp: string) =>
+	moment(timestamp)
+		.utc()
+		.format(),
+);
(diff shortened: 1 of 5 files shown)


src/components/DataTypes/date-time-helpers.tsx
```diff
@@ -0,0 +1,24 @@
+import memoize = require('lodash/memoize');
+
+const pad = (n: number) => (n < 10 ? `0${n}` : n);
+
+export const getDefaultDate = (): string => {
+	const now = new Date();
+	const day = pad(now.getDate());
+	const month = pad(now.getMonth() + 1);
+	return `${now.getFullYear()}-${month}-${day}T10:00`;
```
sradevski: NIT: The original default hour was `00:00`.

src/components/DataTypes/date-time-helpers.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+// Normalize a timestamp to a RFC3339 timestamp, which is required for JSON
+// schema.
+export const normalizeDateTime = memoize((timestamp: string) => {
```
LucianBuzzo: This could be changed to `new Date(timestamp).toISOString()`

src/components/DataTypes/date-time-helpers.tsx
```diff
@@ -0,0 +1,24 @@
+import memoize = require('lodash/memoize');
+
+const pad = (n: number) => (n < 10 ? `0${n}` : n);
+
+export const getDefaultDate = (): string => {
```
LucianBuzzo: This could be changed to 
```
() => {
  const date = new Date()
  date.setHours(0,0,0,0)
  return date.toISOString()
}
```

JackuB: The original format this function was outputting is `2019-03-13T00:00`, this would change it to `2019-03-13T00:00:00Z`. It looks desirable, but that function is untested, so it's hard to tell. I'm okay moving it to the new format, same as the normalize function

src/components/DataTypes/date-time-helpers.tsx
sradevski: Just a note, theoretically RFC3339 and ISO are slightly different, but I don't think it makes a difference in this case.

JackuB: Removing the milliseconds could be enough in our case: https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats

_Should be fine™️_

LucianBuzzo: Trimming the end of the timestamp should be enough in this case

JackuB:
Looks like bundle size is unaffected, because `moment` is part of `mermaid` as well ¯\_(ツ)_/¯

sradevski:
@JackuB I still think it's worth removing a direct dependency as we might get rid of mermaid using tree-shaking (if it is not used). Also, the less dependencies, the better.

JackuB:
@sradevski yea. Anyway, commit updated, feel free to review

Issue #772, opened 2019-03-13
Should allow to use the Button.a component for the Modal's buttons


thgreasi:
**Is your feature request related to a problem? Please describe.**
We recently had designs that grouped links in the footer area of the modal, in an effort to keep the main content area cleaner and move all clickable elements near to each other. The current Modal implementation only allows Button elements and not Button.a elements to be used.

**Describe the solution you'd like**
We should be able to have links in the Modal's footer button area.

See: https://www.flowdock.com/app/rulemotion/r-hardware/threads/K5-XbgzeuiK0OtFYEmPnv8WqFwZ

PR #773, opened 2019-03-13
Modal: Allow using Anchor Buttons in the Modal footer

thgreasi:
Resolves: #772
Change-type: minor


src/components/Modal.tsx
```diff
@@ -69,6 +69,14 @@ const ModalPanel = styled(Box)`
 	box-shadow: 0px 0px 15px 1px rgba(0, 0, 0, 0.4);
 `;
 
+const ModalButton = (props: ButtonProps | ButtonAnchorProps) => {
+	return 'href' in props && props.href ? (
```
sradevski: Why do we need the `'href' in props` part? Seems redundant to me, but I might be missing something.

thgreasi: @sradevski after `'href' in props` TS can infer that `props` are of `ButtonAnchorProps`. W/o it we would need to cast the props before the check `(props as ButtonAnchorProps).href`, otherwise we would get a `'href' does not exist on type 'ButtonProps'` error.

thgreasi:
@sradevski I really couldn't reason why the Overflow screenshot changed. It might just be the bytes that changed, since the visual comparison doesn't show any red diffs.

thgreasi:
@resin-ci retest

PR #774, opened 2019-03-14
Move TabbedBox from unstable to core components as Tabs

xginn8:
HQ: balena-io/balena#1663
Change-type: minor
Signed-off-by: Matthew McGinn <mamcgi@gmail.com>


xginn8:
Demo screenshots attached 
rendition-tab-2
rendition-tab-1

src/components/TabbedBox.tsx
```diff
@@ -0,0 +1,96 @@
...
+	children: JSX.Element[];
+}
+
+interface TabbedBoxState {
+	show: number;
+}
+
+const FlexTab = styled(Flex)`
+	border-style: hidden;
+	border-bottom: black;
```
LucianBuzzo: Can you please change the border color to match the border color we use on textareas and inputs?

xginn8: Done

sradevski:
@xginn8 @LucianBuzzo Any good reason why we should move it to core components? I think what we talked about basing rendition on Grommet will bring us better Tabs out of the box, so I would leave this in the unstable components until then.

LucianBuzzo:
@sradevski This is needed right now to add some much needed functionality to the balenaCloud dashboard. Moving it into core seems fine for now considering the narrow scope of the component functionality.

sradevski:
@LucianBuzzo What I meant is maybe we can just use it from the unstable set of components, but it's also fine if we move it to the core components I guess. We will have a breaking change when moving to grommet anyway, so we can fix it then.

LucianBuzzo:
@sradevski `Tabs` is a much better name - @xginn8 want to update this component to use the simpler name?

xginn8:
> @sradevski `Tabs` is a much better name - @xginn8 want to update this component to use the simpler name?

Updated the name everywhere to `Tabs`, thanks!

xginn8:
@resin-ci retest

PR #775, opened 2019-03-15
Fix how nested modals close on keydown

sradevski:
Originally an Escape keydown would close the
parent modal, resulting in all modals being closed.
This fixes that and it makes modals close in a
stack-like manner.

Change-type: patch


src/components/Modal.tsx
```diff
@@ -79,7 +79,7 @@ const ModalButton = (props: ButtonProps | ButtonAnchorProps) => {
 
 class Modal extends React.Component<ThemedModalProps, any> {
 	static mountedCount = 0;
-
+	ownIndex = 0;
```
thgreasi: Let's add a newline as well :)

PR #776, opened 2019-03-15
Fix up bad TZ advice (London != UTC)

xginn8:
Change-type: patch
Signed-off-by: Matthew McGinn <mamcgi@gmail.com>


cyplo:
Heya, running `npm test` on my local machine produces an error with or without this change on the newest master - is this expected @xginn8 @sradevski ?

JackuB:
Yeah, Node won't allow you to change the TZ after it was once initiated, if you want to mock `Date()` TZ

cyplo:
Yeah, I like it, as then we won't require people to know what to set on their machines and it should just work, if I understood correctly.

Issue #779, opened 2019-03-20
Terminal: The highlighted character is not on the same baseline as the rest of the text


thgreasi:
**Describe the bug**
The highlighted character is not on the same baseline as the rest of the text.

**To Reproduce**
1. In a terminal that support the arrow keys, type in some text and then use the arrow to go backwards onto it

## Expected Behavior

Highlighted text is just highlighted

## Actual Behavior

Highlighted characters are shifted lower on the baseline

**Screenshots**

![Screenshot_2019-03-19_17-19-43](https://user-images.githubusercontent.com/38863/54627407-4882d280-4a6b-11e9-8f7c-278a4e6ccaa8.png)

**Specifications (please complete the following information):**
- Browser: https://whatsmybrowser.org/b/HZDZVSW
- OS: Arch Linux
- Environment: Production

**Additional context**
See: https://github.com/balena-io/resin-ui/issues/2680

PR #780, opened 2019-03-20
Bump xterm to v3.12.0

thgreasi:
it seems that
* we forgot to update the styled during the last upgrade
* even after updating the styles, there was an alignment issue with the character under the terminal's cursor, which gets fixed in v3.12.0

Resolves: #779
Change-type: minor


LucianBuzzo:
Good work, looks like you need to update snapshots and possibly the screenshot as well.

src/stories/Terminal.js
```diff
@@ -83,12 +83,7 @@ class InteractiveTerm extends Terminal {
         const printable =
           !ev.altKey && !ev.altGraphKey && !ev.ctrlKey && !ev.metaKey
         // Ignore arrow keys
-        if (
-          ev.code === 'ArrowUp' ||
-          ev.code === 'ArrowDown' ||
-          ev.code === 'ArrowLeft' ||
-          ev.code === 'ArrowRight'
-        ) {
+        if (ev.code === 'ArrowUp' || ev.code === 'ArrowDown') {
```
sradevski: It seems I cannot use `backspace` in the current storybook, can you check that as well?

thgreasi: @sradevski  Yea that was how it was already and I'm not sure how to fix it.
I guess that it has to do with the fact that we are using a dummy input, instead of a tty which responds back.
Listening to the Backspace button isn't hard, but now that we can move the cursor, it felt like more work that what I was planning for the needs of this bugfix PR.

sradevski: @thgreasi Yeah you can leave it as it is, not a big deal. Also there is [this bug](https://github.com/balena-io/resin-ui/issues/2682) that we need to resolve ASAP, but hopefully it will be resolved upstream soon.

thgreasi:
@LucianBuzzo @sradevski I updated the screenshots and tests now pass. Please take a second look in case you spot anything weird in them.

PR #783, opened 2019-03-20
components: Remove deprecated DeviceStatusGauge

dimitrisnl:
Connects-to: #782
Change-type: major


dimitrisnl:
@sradevski,  Any clues why the tests fail with `Received value does not match stored snapshot "JellyForm component should match the stored snapshot 1".`

..like this has nothing to do with JellyForm

sradevski:
@dimitrisnl AFAIK styled-components generates the classname hash in a sequential manner, so when you removed the component it changed the classname hash of the Form. Just update the snapshot.

README.md
```diff
@@ -1032,6 +1094,7 @@ If you would like to try them out they can be loaded using
 ## Upgrading
 
 [Upgrading from 3.x to 4.x](docs/migrating_3x-4x.md)
+[Upgrading from 4.x to 5.x](docs/migrating_4x-5x.md)
```
sradevski: Add a newline, as currently they appear next to each other.

README.md
```diff
@@ -45,45 +45,107 @@ ReactDOM.render(
...
-- [Select](#select)
-- [Swatches](#swatches)
-- [Table](#table)
-- [Tabs](#tabs)
-- [Terminal](#terminal)
-- [TextWithCopy](#textwithcopy)
-- [Textarea](#textarea)
-- [Tooltips](#tooltips)
-- [Txt](#txt)
+- [Rendition](#rendition)
```
sradevski: How did this change?

dimitrisnl: Ah, damn. It was my prettier.

dimitrisnl:
@sradevski 	

We already use `DeleteButton` in the filters. Should I proceed to update these parts and remove this component?

sradevski:
@dimitrisnl Yes, I believe it is a good time to do so, since the component was already deprecated for quite a bit.

README.md
dimitrisnl: Actually, I'm not so sure really. Must be one of my VSCode extensions.

sradevski: @dimitrisnl I think you should run `build:docs` instead of manually editing the README. There is a README template in `docs` that you should modify, and the README should be automatically generated.

Issue #792, opened 2019-03-25
Autogenerate typings file


LucianBuzzo:
Prerequisite on #791

dimitrisnl:
https://github.com/SitePen/dts-generator

LucianBuzzo:
@dimitrisnl This should be configurable from the typescript compiler options

dimitrisnl:
We have two more JS files in /unstable/components/Form/fields that need converting.

PR #799, opened 2019-03-29
Autogenerate typings

dimitrisnl:
Connects-to: #792
Change-type: minor


dimitrisnl:
@LucianBuzzo 	I'm not really sure how to proceed with the auto-generation. Any pointers?

LucianBuzzo:
@dimitrisnl What issues are you facing?

dimitrisnl:
@LucianBuzzo 

[Reading the documentation](https://www.typescriptlang.org/docs/handbook/compiler-options.html), I would expect that having the `--declaration` flag  while running `npm run build (rimraf dist && tsc && npm run build:docs)` a `.d.ts` file would be generated - which is not the case.

dimitrisnl:
@LucianBuzzo 	Do we want to keep the asRendition wrapper?

LucianBuzzo:
@dimitrisnl Yep we want to keep it (now and in the future) - for this PR, the functionality of the lib should not change at all

dimitrisnl:
Some failing tests, yet there are no visible changes in Alert, Pill & Navbar. Trying to figure it out.

dimitrisnl:
@resin-ci 	retest

src/common-types.d.ts
```diff
@@ -125,6 +125,221 @@ interface Coloring {
...
+interface BoxProps extends DefaultProps, Tooltip {
+	flex?: string | string[];
+	order?: number | string | Array<number | string>;
+}
+
+/* Used by:
+	1. Txt.tsx
+	2. Pill.tsx
+*/
+interface TxtProps extends DefaultProps, Tooltip {
```
sradevski: I am commenting on the `TxtProps`, but most of these "common types" are not actually necessary. What you need to do is export the interface from the component (for example, `export interface TxtProps` in `TxtProps`, and then import it from there in the `Pill` component in this case.

src/components/Txt.tsx
```diff
@@ -1,4 +1,3 @@
-import { Theme, TxtProps } from 'rendition';
```
LucianBuzzo: This seems to be missing the prop interface declaration

LucianBuzzo: Is it inferred automatically?

typings/rendition.d.ts
```diff
@@ -1,242 +1,12 @@
 declare module 'rendition' {
```
LucianBuzzo: Can we completely delete this file?

LucianBuzzo:
When I tested this against the UI it seems that the "common-types" are not picked up automatically and may need to be explicitly imported

src/common-types.d.ts
dimitrisnl: makes sense @sradevski Thanks for the pointer

dimitrisnl: I tried it and it looks good, but my problem lies with the 'extra.d.ts' and 'unstable.d.ts' files, where I can't reference these typings relatevely.

I might have to break those apart.

typings/rendition.d.ts
dimitrisnl: We should, left it for now in order to tackle the 'Form' typings a bit later.

dimitrisnl:
@LucianBuzzo 

Isn't this in direct conflict with having a `rendition.d.ts` file? 
`common-types` is essentially the same but renamed.

Now it gets a little weird, with having the dashboard look for typings this way

```
import { Coloring, RenderableElementWithProps, Theme } from 'rendition';
```

dimitrisnl:
@LucianBuzzo 
Screen Shot 2019-04-04 at 6 29 25 PM

The same thing happens when explicitly importing the types. Notice the error at the top right of the image.

The `.ts` file has no errors during compilation which is baffling

LucianBuzzo:
@dimitrisnl The import should have a path I think, like:
```
import { Colouring } from '../common-types'
```

dimitrisnl:
@LucianBuzzo That would be ideal if the other modules (extra.d.ts, unstable.d.ts) could handle relative paths.

I figured it out, tsconfing wouldn't pick these typings before.

src/components/Filters/index.tsx
```diff
@@ -86,6 +132,20 @@ interface FiltersState {
 	schema: JSONSchema6;
 }
 
+interface FiltersProps extends DefaultProps {
```
sradevski: Maybe this needs to be exported as well?

src/components/Pill.tsx
```diff
@@ -1,10 +1,17 @@
...
 import * as React from 'react';
 import { compose } from 'recompose';
-import { PillProps } from 'rendition';
 import styled, { withTheme } from 'styled-components';
 import asRendition from '../asRendition';
 import { getColor, withProps } from '../utils';
-import Txt from './Txt';
+import Txt, { TxtProps } from './Txt';
+
+export type PillProps = Coloring &
```
sradevski: Maybe make it an interface for consistency purposes.

typings/common-types.d.ts
```diff
@@ -0,0 +1,167 @@
+declare module 'common-types' {
```
LucianBuzzo: Why is this a module instead of a simple `.ts` file that you can import interfaces from?

dimitrisnl: No real reason. I assumed that it made sense for types to be in a .ts' file. Should I update it?

Note that you can't have relative paths in declaration files, so if I wanted to reference something in `unstable.d.ts` I had to have `common-types` the same format.

dimitrisnl:
@LucianBuzzo I followed up with dropping the '.ts' files in favour of simple typescript ones.

dimitrisnl:
Super interesting, some snapshots changed - investigating.

Issue #800, opened 2019-04-01
Upgrade to styled-components v4


LucianBuzzo:
This is will involve updating styled-system and grid-styled components

PR #804, opened 2019-04-11
Upgrade styled-components to v4 and related libraries

sradevski:
There are several breaking changes due to changes in
styled components and styled system. The other breaking
change is that components have to be wrapped in a
Provider now to pass down the theme.

Connects-to: #800
Change-type: major


LucianBuzzo:
@sradevski What are the breaking changes?

src/asRendition.tsx
```diff
@@ -60,32 +55,26 @@ export const withStyledSystem = (child: React.StatelessComponent) => {
...
-
-	(Base.defaultProps as any).theme = theme;
-
-	return (props: any) => {
-		return <Base {...props} />;
-	};
-};
+const enhancers = [withTheme, withTooltip, withStyledSystem];
 
-const Tag = tag(blacklist);
```
LucianBuzzo: How come this has been removed?

sradevski:
@LucianBuzzo It's a draft PR, it wasn't ready for reviews, I just wanted to see if the tests will pass since locally they never finish for some reason. 
The breaking changes are: 

1. If components are not wrapped in a Provider, they will now be unstyled (previously they would have the default theme applied if wrapped in `asRendition`).
2. Removed support for `w` and `f` and replaced with `width` and `fontSize` (as they were depricated and removed in styled system).
3. Removed support for `wrap`, `align`, `justify`, `direction` and replaced with `flexWrap`, `justifyContent`, `alignItems`, and `flexDirection`. Moreover, `flexWrap` changed from boolean to string (following the CSS definition)
4. Styled system spacing props' CSS priority might have changed and it might affect spacing unexpectedly.

LucianBuzzo:
@sradevski Ok cool, thanks for letting me know. Since the `Flex` component is so widely used, could we provide a backwards compatible interface for know? This would allow us to more easily implement it in the balenaCloud dashboard

src/asRendition.tsx
sradevski: It isn't properly implemented (doesn't forward `ref`), so I temporarily removed it and will add a new hoc (it's a 3-liner anyway, no need to depend on another package)

sradevski:
@LucianBuzzo I was torn between keeping it backward-compatible and making it a breaking change, and what I thought was it is a pretty easy "search-and-replace" breaking change so it might be better to just go for it. Otherwise, the code will stick around for much longer and we will never move to the new styled-system interface.

LucianBuzzo:
@sradevski You make a good argument, lets go for it!

sradevski:
I also need to update all of the READMEs for some components, but I will do that in a follow-up commit.

src/components/BadgeSelect.tsx
```diff
@@ -99,7 +103,7 @@ class BadgeSelect extends React.Component<BadgeSelectProps, BadgeSelectState> {
 
 				{this.props.items.map(item => (
 					<ButtonWrapper key={item} onClick={() => this.setSelected(item)}>
-						<Badge text={item} />
+						{<Badge text={item} />}
```
dimitrisnl: Why is this needed?

src/components/Container.tsx
```diff
@@ -1,34 +1,49 @@
...
-	fontSize,
-	responsiveStyle,
-	space,
+	maxWidth,
+	MaxWidthProps,
 	textAlign,
+	TextAlignProps,
 } from 'styled-system';
-import { DefaultProps, Theme } from '../common-types';
+import { Theme } from '~/common-types';
```
dimitrisnl: @sradevski Can we use `../` instead? This one fails in the dashboard. There is also in some other places (Img, Pager, etc)

sradevski: @dimitrisnl Yes, this was auto-imported in VSCode, I will update it everywhere, thanks for the info.

src/components/BadgeSelect.tsx
sradevski: Not needed, probably just remained after doing something else.

Issue #805, opened 2019-04-13
Terminal crashes when resizing and content has "tab" characters in it


sradevski:
Directly related to: https://github.com/xtermjs/xterm.js/issues/1932#issuecomment-482636851

PR #806, opened 2019-04-13
Upgrade xterm to fix crashing on tab character

sradevski:
The issue was fixed by https://github.com/xtermjs/xterm.js/releases/tag/3.12.1

Connects-to: #805
Change-type: patch

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17926,7 +17926,6 @@
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/strip-css-comments/-/strip-css-comments-3.0.0.tgz",
       "integrity": "sha1-elYl7/iisibPiUehElTaluE9rok=",
-      "dev": true,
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -17926,7 +17926,6 @@
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/strip-css-comments/-/strip-css-comments-3.0.0.tgz",
       "integrity": "sha1-elYl7/iisibPiUehElTaluE9rok=",
-      "dev": true,
```
thgreasi: I'm not sure why this got removed, but the package.json seems fine and we don't publish the lock anyway, so I will just ignore it...

sradevski: @thgreasi `scope-css` requires `strip-css-comments`, so that is probably why. But yes, it is irrelevant.

Issue #815, opened 2019-04-29
Explicitly use 'em' values in breakpoints


dimitrisnl:
Styled-systems as of (2.0) converts integer values to pixels, so we have to state that we use 'em' values.

PR #816, opened 2019-04-29
layout: Fix default theme's breakpoints

dimitrisnl:
Fixed: #815
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Container.tsx
+++ b/src/components/Container.tsx
@@ -21,7 +21,7 @@ const Container = (props: ThemedContainerProps) => {
 		<ContainerBase
 			{...props}
 			maxWidth={props.theme.breakpoints.map(
-				(bp, i) => `${bp - props.theme.space[i]}em`,
+				(bp, i) => `calc(${bp} - ${props.theme.space[i]}em)`,
 			)}
 		/>
 	);
--- a/__tests__/components/__snapshots__/Banner.spec.js.snap
+++ b/__tests__/components/__snapshots__/Banner.spec.js.snap
@@ -8,7 +8,7 @@ exports[`Button renders correctly 1`] = `
     className="sc-bdVaJa lgVSbN"
   >
     <div
-      className="sc-htoDjs ckGGnw sc-gzVnrw hnoozl"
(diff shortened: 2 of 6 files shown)


src/components/Container.tsx
```diff
@@ -21,7 +21,7 @@ const Container = (props: ThemedContainerProps) => {
 		<ContainerBase
 			{...props}
 			maxWidth={props.theme.breakpoints.map(
-				(bp, i) => `${bp - props.theme.space[i]}em`,
+				(bp, i) => `calc(${bp} - ${props.theme.space[i]}em)`,
```
dimitrisnl: This is a magic number that I don't really understand.
It has been there since the inception of the component https://github.com/balena-io-modules/rendition/blob/0c7c35edf25b9047553058b4d6dd00c6cad2f3e3/src/components/Container.js

sradevski: I am assuming this was done to have some padding (and not occupy the entire width), but it is quite confusing, I agree. I think we also need to add `px` to the spacing parameter, otherwise `calc` won't work as expected, but I am not sure what the desired maxWidth is anyhow.

dimitrisnl: As it stands it replicates the existing behaviour, but indeed we need to reconsider why we do it this way

Any thoughts @LucianBuzzo ?

LucianBuzzo: There is no need for a `calc` here IMO, can we just make this as simple as possible?

dimitrisnl: We need calc because the breakpoint value is as follows now `48em`. I would either have to replace the `em` part and make it an integer or just use calc @LucianBuzzo @sradevski

dimitrisnl: @sradevski Am I thinking it wrong? I would like to merge this asap but I would love to refactor it if needed.

sradevski: @dimitrisnl The previous and the new logic you wrote should be the same, so I think its fine. Whether that was intended, and why it is like that is a a separate conversation.

Issue #820, opened 2019-04-30
Outline buttons should have white background as opposed to transparent


dimitrisnl:
Imo, having a transparent background is a special case that we should specifically mark using the `bg` props

PR #821, opened 2019-04-30
Use solid white background for outline buttons

dimitrisnl:
Fixes: #820
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


dimitrisnl:
I expect that everyone that checks the Outline button, thinks that the background is of solid white color. In colored backgrounds, the outline button blends in and it's very hard to notice.

src/components/Button.tsx
```diff
@@ -34,6 +34,7 @@ const ColouredButton = styled(ButtonBase)`
 `;
 
 const Outline = styled(ButtonBase)`
+	background: #fff;
```
sradevski: I would only fall back to `white` if the `bg` prop is not passed to it maybe? Not a strong opinion on this.

dimitrisnl: I agree @sradevski , that makes sense

sradevski:
@dimitrisnl Just update the snapshots, I see they failed again. Also adding an outlined button to storybook with a bg might be nice.

dimitrisnl:
Screen Shot 2019-05-01 at 10 23 23 AM

That was a great suggestion. Seems like outline button's color, mimics the bg property somehow. 

```js
          <Button
            m={2}
            outline
            bg="#000"
            >Outline</Button>
        </Box>
        <Box m={3}>
          <Button
            m={2}
            primary
            bg='red'
            >Outline</Button>
        </Box>
```

sradevski:
@dimitrisnl when I think about it, shouldn't the outline button actually have transparent background? If you want a button with a white color, you should just use a normal button with a white background.

dimitrisnl:
I don't have any strong opinion, but we don't have any.

What I want, are some simple white background buttons, with green, red, yellow, etc colors to match the products. When using outline they are hidden in the colored background.

sradevski:
@dimitrisnl But in that case, maybe it is better to just use `bg="fff"` on a normal button. Any way is fine, it is just that it feels that an "Outline" button should just have an outline and be clickable, and not have a solid background.

dimitrisnl:
I see. It just seems strange to be that we don't that the simplest form a button, a plain white one.

PR #827, opened 2019-05-02
Update default theme color palette

dimitrisnl:
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


dimitrisnl:
## On the left the current palette and on the right the new one

Screen Shot 2019-05-02 at 3 49 15 PM
Screen Shot 2019-05-02 at 3 48 46 PM
Screen Shot 2019-05-02 at 3 48 32 PM
Screen Shot 2019-05-02 at 3 48 19 PM

dimitrisnl:
I would really love to add a new button state, of pure white background, dark text, and white border. We have none and it's the most common case.

package.json
```diff
@@ -119,6 +119,7 @@
     "lodash": "^4.17.11",
     "marked": "^0.4.0",
     "mermaid": "^8.0.0",
+    "polished": "^3.2.0",
```
LucianBuzzo: We're already using [`color`](https://www.npmjs.com/package/color) which has `darken` and `lighten` methods, can we just use that instead of adding another dep?

dimitrisnl: @LucianBuzzo Oh, I missed that.
I went for polished as it was from the Styled component guys

Will use it. 👍

dimitrisnl:
Very strange, `Color` produces different values (!?)

Screen Shot 2019-05-02 at 4 25 44 PM

dimitrisnl:
Another thing that I noticed is that `Color` returns hsl values and they have rounding errors https://github.com/Qix-/color/issues/127 which explains the weird layout of Swatches

LucianBuzzo:
@dimitrisnl You can use the `color.hex()` method

LucianBuzzo:
If `color` is too much of a headache, ditch it entirely and switch to using `polished`.

PR #843, opened 2019-05-14
Fix files that are missing from build output

### LINKED ISSUE #842: Can't import Mermaid component
https://codesandbox.io/s/l0nlp9yzm?fontsize=14

Importing the mermaid component gives the error `Could not find module`

LucianBuzzo:
Fixes #842


LucianBuzzo:
@balena-ci retest

thgreasi:
Crossposting from FD
> @lucianbuzzo, any context on why we removed the BadgeSelect & FilterForm components? This sounds like a major change. Isn't it?

LucianBuzzo:
@thgreasi, The `FilterForm` component was internal to the `Form` component and was never exposed, and the `BadgeSelect` component was actually removed from the export in a previous major version bump but the core file never got deleted.

nazrhom:
@balena-ci retest

PR #855, opened 2019-05-24
Markdown update

LucianBuzzo:
- Don't memoize markdown-to-html function
- Allow inline html in markdown strings


__tests__/components/Markdown.spec.js
```diff
@@ -0,0 +1,275 @@
...
+import React from 'react'
+import renderer from 'react-test-renderer'
+import {
+  Provider
+} from '../../dist'
+import {
+  Markdown
+} from '../../dist/extra/Markdown'
+
+describe('Markdown component', () => {
```
sradevski: What was the reason to add markdown tests of our own? `marked` is much better tested and it feels redundant for us to also test the same things.

src/extra/Markdown.tsx
```diff
@@ -13,10 +13,39 @@ const markedOptions = {
...
 	breaks: true,
 	headerIds: false,
-	sanitize: true,
+	// Input text is sanitized using `sanitize-html` prior to being transformed by
+	// `marked`
+	sanitize: false,
 };
 
-const render = memoize((text: string = '') => marked(text, markedOptions));
+const sanitizerOptions = {
```
sradevski: Maybe we can add a test or two to check if the input is properly sanitized.

__tests__/components/Markdown.spec.js
LucianBuzzo: @sradevski In short, its the same reasons we test anything.
Different combinations of sanitizers and marked options can lead to different results and rendering outputs. Additionally, by testing the output at the rendition level we can ensure that we maintain expected behaviour and a consistent API, even if we change the underlying implementation.

src/extra/Markdown.tsx
LucianBuzzo: Good idea

PR #856, opened 2019-05-27
button: Increase height to 38px

dimitrisnl:
https://github.com/balena-io-modules/rendition/issues/847

The previous implementation was relying on space units.


dimitrisnl:
`npx run jest -- -u` doesn't produce any new snapshots, yet the docker build fails. That's a bit strange.

src/components/Button.tsx
```diff
@@ -25,7 +25,7 @@ const overrideHoverEffect = css`
 const ButtonBase = styled(Button)`
 	font-family: ${props => props.theme.titleFont};
 	font-weight: ${props => normal(props)};
-	height: ${props => px(props.theme.space[4])};
+	height: 38px;
```
sradevski: Wouldn't it be nice if we can configure this through theming? I am OK with fixed value for now as well, just a thought.

sradevski:
@dimitrisnl @LucianBuzzo  I was thinking it might be better to set the button's `white-space: nowrap` in this PR, since it has fixed height anyway. Any objections to that?

dimitrisnl:
I agree, it will help with these weird cases

src/components/Button.tsx
dimitrisnl:
```js
  button: {
      border: {
          width: string;
          radius: string;
      };
      primary: {};
      disabled: {
          opacity: number;
      };
      minWidth: string;
      maxWidth: string;
      padding: {
          vertical: string;
          horizontal: string;
      };
  };
```

Grommet doesn't give the option in the theme.

sradevski: Yeah, I meant we can extend the theme definition, but ignore it for now.

dimitrisnl:
@balena-ci 	retest

src/components/Button.tsx
LucianBuzzo: @dimitrisnl @sradevski Lets fix this now, and extend the theme definition with a `height` value

sradevski:
@dimitrisnl Maybe also include #822 in this PR, but as a separate commit.

src/components/Button.tsx
sradevski: @dimitrisnl can you please add `height` to the theme in this PR? Cheers

dimitrisnl: Ah missed this. Will do 👍

Issue #845, opened 2019-05-16
Improve button styles


dimitrisnl:
- Make the 'Primary' button (light blue shade) the default button.
- Introduce a 'Light' button variant as seen in Zeplin designs
- Ensure that the disabled button doesn't have any hover effects
- Hovered button has the same look as a disabled on. Should probably apply a darker shade as opposed to a light one.

dimitrisnl:
@LucianBuzzo @sradevski

dimitrisnl:
https://github.com/balena-io-modules/rendition/pull/844

sradevski:
The only thing left to do is change the `default` button. How about making `secondary outline` as the default button for now, as it is the closest to the current `default` one and we don't have to make a bunch of changes across projects?

dimitrisnl:
I agree @sradevski

PR #858, opened 2019-05-28
Change button hover style, Add a light button variant

sradevski:
All button styles have been updated to match the new styleguide.

This PR doesn't include an `outline` light button variant as it is not used anywhere as of now.
We can add it as soon as it is required so as not to add unnecessary features.

Currently the default button hasn't changed in order not to create a major version just for it. One thing I was thinking is making `outline secondary` as the default button as it has almost the same look as the current default one, so we don't need to make changes in many places at this point.

Connects-to: #845
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


sradevski:
@balena-ci retest

PR #865, opened 2019-06-06
Add RadioButton and RadioButtonGroup components

sradevski:
The "disabled" effect doesn't match our styleguide yet, but its implementation is a bit more challenging due to how Grommet applies it and it will require a PR upstream. We also need to revise the styleguide as I personally don't like how the disabled effect is applied in all cases, so we can leave it for another PR. Opened an issue to track this [here](https://github.com/balena-io-modules/rendition/issues/868)

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


LucianBuzzo:
@sradevski Why did this get called `RadioButton` instead of `Radio` ?

sradevski:
@LucianBuzzo just so we don't diverge unnecessarily from Grommet, as I don't see a benefit from it being called `Radio` instead.

PR #879, opened 2019-06-19
Add build step when updating jest snapshots

sradevski:
I have been annoyed multiple times where I update everything, and the build fails when I open the PR just because I forgot to build the project before updating the snapshots. I think this wasn't necessary before [this PR](https://github.com/balena-io-modules/rendition/commit/f848e43ec48e0570090b17e52f390b449002dd38).

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "build:docs": "node ./scripts/generateDocs.js",
     "build": "npm run tsify:xterm-css && npm run build:cjs && npm run build:esm5 && npm run build:docs",
     "generate-screenshots": "./scripts/generate-screenshots.sh",
+    "update-snapshots": "npm run build && npm run jest -- -u",
     "build:storybook": "build-storybook -o .out",
     "catch-uncommitted": "catch-uncommitted --catch-no-git --skip-node-versionbot-changes",
     "test:fast": "npm run lint && TZ='Etc/UTC' npm run jest",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "test": "npm run build && npm run test:fast && npm run catch-uncommitted",
     "test:local-visual": "./scripts/test-screenshots.sh",
     "test:visual": "npm run ci:screenshot -- -o .reg/current && reg-cli .reg/current __screenshots__ .reg/diff -R .reg/report.html",
-    "jest": "jest __tests__ --maxWorkers=4",
+    "jest": "npm run build && jest __tests__ --maxWorkers=4",
```
LucianBuzzo: I don't love this - this means that the `test:fast` command now requires the whole library to be built. If anything we should probably add a distinct separate command for updating screenshots. Additionally, the `test` command already incorporates the `build` step

sradevski: @LucianBuzzo I don't think there are cases when you would want to run tests on a stale build, but I am fine either way. I will add another `update-snapshots` command so we always update snapshots based on a fresh build.

package.json
```diff
@@ -27,6 +27,7 @@
     "build:docs": "node ./scripts/generateDocs.js",
     "build": "npm run tsify:xterm-css && npm run build:cjs && npm run build:esm5 && npm run build:docs",
     "generate-screenshots": "./scripts/generate-screenshots.sh",
+    "update-snapshots": "npm run build && npm run jest -- -u",
```
thgreasi: it would be nice if we had the same script prefix for screenshot & snapshot update/generation.

LucianBuzzo: Like this?
```suggestion
    "generate-snapshots": "npm run build && npm run jest -- -u",
```

sradevski: I was trying to be closer to the real action that is happening, but I agree, it is better to be consistent.

PR #882, opened 2019-06-24
Allow specifying the filter components rendering order

sradevski:
There are a couple of files that changed due to running prettify. We need the flexibility to specify the order in order to make the new dashboard designs implementable. This also makes the search input component flex to the entire available width, so it appears better on larger screens. We can discuss whether this is preferable or not.

Note: The screenshots change all the time due to some small glitches, I am working on fixing it in a separate branch.

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


dimitrisnl:
@sradevski 	Can you share the branch you mentioned?

LucianBuzzo:
@sradevski Is there a reason for not simply changing the render order? I feel like we're adding a lot of additional complexity that isn't necessarily needed.

sradevski:
@LucianBuzzo If we decide that we want the same order everywhere, I agree that it is better to simply change the ordering, I was just trying to preserve the existing behavior to isolate the change.

LucianBuzzo:
@sradevski Lets keep it the same everywhere

sradevski:
@LucianBuzzo @dimitrisnl I will pause this until the UI convergence to decide on the ordering of the filters.

sradevski:
This is not necessary as of now, closing

PR #905, opened 2019-07-12
Replace our Input with the Grommet one

### LINKED ISSUE #848: Recreate Input component with Grommet

sradevski:
This actually has no breaking changes as the underlying component is still a native `input`. The only issue is the typings which will be fixed separately in grommet.

The main visual difference of the Input component is it now takes 100% of the parent's width, which wasn't the case before.

Resolves #848 
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Select.tsx
```diff
@@ -14,7 +14,8 @@ const StyledGrommetSelect = styled(GrommetSelect)<InternalSelectProps<any>>`
 	font-family: ${props => props.theme.font};
 	font-size: ${props => px(props.theme.fontSizes[1])};
 	font-weight: normal;
-	height: ${props => (props.emphasized ? '46px' : '36px')};
+	height: ${props =>
+		px((props.emphasized ? props.theme.space[5] : props.theme.space[4]) - 2)};
```
LucianBuzzo: Why is the select component 2px smaller than the input component? Surely they should be the same height?

sradevski: @LucianBuzzo that is so they are actually the same size. The problem is the select component's button has 1px borders that make the actual button `40px`. if you change the height of the button to `38px`, then the alignment with the drop is messed up. There isn't a clean way to set the height otherwise. I should probably add a comment though.

sradevski: Nevermind, found a better way to do it.

LucianBuzzo: We should be using `box-sizing: border-box`, which should constrain the border within the defined height - is this not the case?

sradevski: It is using `box-sizing: border-box`, but instead of setting the height on the input, it should have been the button.

dimitrisnl:
Screen Shot 2019-07-25 at 1 16 31 PM

The monospace version gets its font family overridden by Nunito

sradevski:
@dimitrisnl This was due to a missing semicolon in the stylings, fixed

Issue #906, opened 2019-07-12
Make breakpoints font-size & space agnostic


dimitrisnl:
The current implementation

1. Depends on `em` values, so any change in the global font-size affects the breakpoint values
2. Container component uses [a weird formula](https://github.com/balena-io-modules/rendition/blob/master/src/components/Container.tsx#L24) that takes into account the space units. That way we have another variable that affects the breakpoints.

PR #907, opened 2019-07-12
Update breakpoint handling

dimitrisnl:
We're not enforcing the use of `em` values. Instead we're applying whatever the user providers.

As for the default theme, we're using the breakpoints/container that Bootstrap 4 uses.

Connects-to: #906
Change-type: major

---
##### Contributor checklist

##### Reviewer Guidelines
---


LucianBuzzo:
@dimitrisnl NIT: Your PR description above is really confusing, because it seems to simultaneously refer to both the current state of `master` and the new state that these changes bring

src/theme.ts
```diff
@@ -95,7 +95,8 @@ const colors = {
 const defaultControlHeight = 36;
 const emphasizedControlHeight = 48;
 
-export const breakpoints = ['32em', '48em', '64em', '80em'];
+export const breakpoints = ['576px', '768px', '992px', '1200px'];
+export const containers = ['540px', '720px', '960px', '1140px'];
```
LucianBuzzo: Why do we have a separate list of `containers` breakpoints here? it seems completely redundant - can we not just make them the same?

dimitrisnl: But that's what we were doing before. We would abstract an arbitrary number from the breakpoint to calculate the width.

That's what I'm going for:

Screen Shot 2019-07-12 at 4 20 20 PM

LucianBuzzo: Make the breakpoints numbers, and then apply a conversion to reduce the number in the container component.

sradevski: Styled system expects a unit, so you can just use `calc` as before.

dimitrisnl: Why not give the option to the user the change the container size?

sradevski: The container size is meant to be coupled with the breakpoints, plus some margin, right? I think in order to answer this we need to know what exactly the container is meant to do.

dimitrisnl: I took these breakpoints/container sizes from bootstrap as a reference. 

Considering there isn't a clear formula for the max-width calculation, I decided to include them in a `container` array.

My use-case is to emulate the previous behavior of the container, without having the side effects we do now.

If we want to use a specific magic number for the calculation, I don't have any objection. Current we extract the values from the `theme.space` array which makes no sense.

src/components/Container.tsx
```diff
@@ -17,14 +17,7 @@ const ContainerBase = styled.div<ThemedContainerProps>`
...
 const Container = (props: ThemedContainerProps) => {
-	return (
-		<ContainerBase
-			{...props}
-			maxWidth={props.theme.breakpoints.map(
-				(bp, i) => `calc(${bp} - ${props.theme.space[i]}em)`,
-			)}
-		/>
-	);
+	return <ContainerBase {...props} maxWidth={props.theme.containers} />;
```
sradevski: How about like this:
```suggestion
	return <ContainerBase {...props} px={[0, 1, 2, 3]} maxWidth={props.theme.breakpoints} />;
```

dimitrisnl: @sradevski 	The thing is that the common libraries don't apply the breakpoint as the max-width. Bootstrap does that with `container-fluid` instead.

LucianBuzzo:
@dimitrisnl I experimented with this PR, and it seems like these changes still don't give the desired bootstrap-style functionality. I made some changes that I think work quite well, you could give them a try and see what you think:
```
diff --git a/src/components/Container.tsx b/src/components/Container.tsx
index b8ee870..25ee5f6 100644
--- a/src/components/Container.tsx
+++ b/src/components/Container.tsx
@@ -9,6 +9,7 @@ import {
        TextAlignProps,
 } from 'styled-system';
 import asRendition from '../asRendition';
+import { px } from '../utils'
 import { DefaultProps, RenditionSystemProps, Theme } from '../common-types';
 
 const ContainerBase = styled.div<ThemedContainerProps>`
@@ -17,7 +18,12 @@ const ContainerBase = styled.div<ThemedContainerProps>`
 `;
 
 const Container = (props: ThemedContainerProps) => {
-       return <ContainerBase {...props} maxWidth={props.theme.containers} />;
+       const gutter = props.theme.space[3];
+       const widths = props.theme.breakpoints.map((point) => {
+               return `calc(${point} - ${px(gutter)})`
+       });
+
+       return <ContainerBase {...props} maxWidth={['100%', ...widths]} />;
 };
 
 Container.displayName = 'Container';
```

dimitrisnl:
@LucianBuzzo 	I think that works great. Another reason I wanted the gutter/max-width available was so that I could hardcode the current values that the website uses. As it stands I'll have to do responsive fixes in most (if not all pages) which wasn't my intention.

dimitrisnl:
@sradevski 	@LucianBuzzo 	

Let me recap

The website atm uses these values

```
@media screen and (min-width: 64em) -> max-width: 74em;
@media screen and (min-width: 48em) -> max-width: 56em;
@media screen and (min-width: 32em) -> max-width: 44em;
max-width: 32em


1024px 1184px
768px 896px
512px 704px
512px
```

They come directly from the previous implementation.
Due to the changes in the global font size, these values are now messed up.
I want to proceed with a solution that will not break the website.
I believe giving the option to conditionally set the container size is the best.

The previous implementation was removing arbitrary space values from the max-width
https://github.com/balena-io-modules/rendition/pull/907/files#diff-f2e94b1f317257f5bdc621189300ac50L24

Of course, this is no ideal and it makes no sense to me, but I have to respect that if I don't want to fix the responsiveness in all of the website pages. So having a fixed value doesn't work for now.

dimitrisnl:
Closed in favour of https://github.com/balena-io-modules/rendition/issues/918

PR #925, opened 2019-07-31
Recreate textarea

dimitrisnl:
---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Textarea.tsx
```diff
@@ -1,27 +1,37 @@
...
+	}
+	if (props.valid) {
+		return props.theme.colors.success.main;
+	}
+};
+
+const StyledGrommetTextArea = styled(GrommetTextArea)`
+	${monospace};
+
+	padding: 15px 20px;
```
sradevski: Let's have this as one of the spacing array from the theme (I guess 16px both axes)

dimitrisnl: Ah, ok. Just went with the designs.

sradevski: @dimitrisnl actually I also see that for the Input and Select component we use `14px 20px`, so maybe go with that for consistency.

PR #926, opened 2019-07-31
table: Display pager only when needed

dimitrisnl:
1. Doesn't display pager when there are no elements
2. Doesn't display pager when the elements fit into a single page

Fixes: #895
Change-type: minor

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Table/index.tsx
+++ b/src/components/Table/index.tsx
@@ -404,9 +404,11 @@ class Table<T> extends React.Component<TableProps<T>, TableState<T>> {
 
 		const sortedData = this.sortData(items).slice(lowerBound, upperBound);
 
+		const shouldShowPaper = !!usePager && totalItems > 0;
+
 		return (
 			<>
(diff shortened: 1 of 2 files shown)


src/components/Table/index.tsx
```diff
@@ -404,9 +404,12 @@ class Table<T> extends React.Component<TableProps<T>, TableState<T>> {
 
 		const sortedData = this.sortData(items).slice(lowerBound, upperBound);
 
+		const shouldShowPaper =
+			!!usePager && totalItems > _itemsPerpage && totalItems > 0;
```
sradevski: It might be better to show the pager when there are items, so as they know how many items are in the table, and also to have a more consistent experience. I think we can hide it only where there are no items at all, as then it doesn't make much sense.

PR #938, opened 2019-08-09
docs: fix up misleading Checkbox link

xginn8:
Change-type: patch
Signed-off-by: Matthew McGinn <matthew@balena.io>


xginn8:
heh "misleading" link, see what i did there :})

PR #939, opened 2019-08-09
docs: fix up misleading Swatches link

xginn8:
Change-type: patch
Signed-off-by: Matthew McGinn <matthew@balena.io>


xginn8:
i did it again :)

dimitrisnl:
Let's make sure to deploy storybook, since these changes won't be automatically available

xginn8:
@dimitrisnl how shall we deploy storybook? is that something i can self-serve?

dimitrisnl:
Hmm, it's good. I'll keep an eye on it and deploy it myself. Essentially we run the deploy script locally as there is something wrong with our pipeline atm.

PR #960, opened 2019-09-06
Colocate tests with components

sradevski:
This PR moves all tests and snapshots next to the components the test correlates to. It also separates the `Box` and `Flex` components into their own folders for consistency sake.
It seems previously the eslint runs didn't format the `.spec` files as expected, so that change is also included as a consequence (so most of the diff changes come from there). 

---
##### Contributor checklist

##### Reviewer Guidelines
---


package.json
```diff
@@ -35,9 +35,9 @@
     "test": "npm run build && npm run test:fast && npm run catch-uncommitted",
     "test:local-visual": "./scripts/test-screenshots.sh",
     "test:visual": "npm run ci:screenshot -- -o .reg/current && reg-cli .reg/current __screenshots__ .reg/diff -R .reg/report.html",
-    "jest": "jest __tests__ --maxWorkers=4",
-    "eslint": "eslint src __tests__ .storybook",
-    "eslint-fix": "eslint src __tests__ .storybook --fix",
+    "jest": "jest --maxWorkers=4",
+    "eslint": "eslint src **/*.spec.js .storybook",
```
LucianBuzzo: if spec files are in the src directory I don't think you need the wildcard expansion

sradevski: Good point, I just replaced the __tests_ part without thinking about it.

sradevski:
@balena-ci retest

Issue #944, opened 2019-08-14
ComponentWIllReceiveProps is now deprecated.


dimitrisnl:
As of 16.9.0 (August 8, 2019) both this and [some other lifecycle](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html) methods will still work but will throw some warnings.

In the next major release, we have to append the `UNSAFE_` prefix, if we want to keep the same functionality. Ideally, we should refactor any instance that uses these methods.

dimitrisnl:
https://github.com/balena-io-modules/rendition/search?q=componentWillReceiveProps&unscoped_q=componentWillReceiveProps

dimitrisnl:
runnin `npx react-codemod rename-unsafe-lifecycles` will do the patch fix too.

alexandrtovmach:
@dimitrisnl Please take a look to my PR that's refactor deprecated hooks #968

alexandrtovmach:
Fixed by #968 . Issue should be closed

PR #968, opened 2019-09-13
fix: Refactored deprecated react lifecycle hooks

alexandrtovmach:
Connects-to: #944 
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Filters/index.tsx
+++ b/src/components/Filters/index.tsx
@@ -89,23 +89,23 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
 		this.state.edit.push(this.getCleanEditModel());
 	}
 
-	componentWillReceiveProps(nextProps: FiltersProps) {
+	componentDidUpdate(prevProps: FiltersProps) {
 		const newState: any = {};
 
 		// If the schema prop updates, also update the internal 'flat' schema
-		if (!isEqual(nextProps.schema, this.props.schema)) {
-			newState.schema = SchemaSieve.flattenSchema(nextProps.schema);
+		if (!isEqual(prevProps.schema, this.props.schema)) {
+			newState.schema = SchemaSieve.flattenSchema(this.props.schema);
 		}
 
-		if (!isEqual(nextProps.filters, this.props.filters)) {
-			const filters = nextProps.filters || [];
+		if (!isEqual(prevProps.filters, this.props.filters)) {
+			const filters = this.props.filters || [];
 			newState.filters = filters.map(filter =>
 				SchemaSieve.flattenSchema(filter),
 			);
 		}
 
-		if (!isEqual(nextProps.views, this.props.views)) {
-			const views = nextProps.views || [];
+		if (!isEqual(prevProps.views, this.props.views)) {
+			const views = this.props.views || [];
 			newState.views = this.flattenViews(views);
 		}
 
--- a/src/components/Table/index.tsx
+++ b/src/components/Table/index.tsx
@@ -3,6 +3,7 @@ import filter from 'lodash/filter';
 import find from 'lodash/find';
 import includes from 'lodash/includes';
 import isArray from 'lodash/isArray';
+import isEqual from 'lodash/isEqual';
 import isFunction from 'lodash/isFunction';
 import isPlainObject from 'lodash/isPlainObject';
 import map from 'lodash/map';
@@ -161,9 +162,11 @@ class Table<T> extends React.Component<TableProps<T>, TableState<T>> {
 		};
 	}
 
-	componentWillReceiveProps(newProps: TableProps<T>) {
-		if (newProps.sort !== this.props.sort) {
-			this.setState({ sort: newProps.sort } as TableState<T>);
+	componentDidUpdate(prevProps: TableProps<T>) {
+		if (this.props.sort && !isEqual(prevProps.sort, this.props.sort)) {
+			this.setState({
+				sort: this.props.sort,
+			});
 		}
 	}
 
@@ -221,7 +224,7 @@ class Table<T> extends React.Component<TableProps<T>, TableState<T>> {
 
 	sortData(data: T[]): T[] {
 		const { sort } = this.state;
-		if (sort.field === null) {
+		if (!sort || sort.field === null) {
 			return data;
 		}
 
--- a/src/unstable/components/Form/index.tsx
+++ b/src/unstable/components/Form/index.tsx
@@ -93,16 +93,16 @@ export default class FormHOC extends React.Component<
 		widgets[name] = value;
 	}
 
-	componentWillReceiveProps(nextProps: FormProps) {
-		if (!isEqual(this.props.value, nextProps.value)) {
+	componentDidUpdate(prevProps: FormProps) {
+		if (!isEqual(this.props.value, prevProps.value)) {
 			this.setState({
-				value: nextProps.value,
+				value: this.props.value,
 			});
 		}
 
-		if (!isEqual(this.props.schema, nextProps.schema)) {
+		if (!isEqual(this.props.schema, prevProps.schema)) {
 			this.setState({
-				schema: this.parseSchema(nextProps.schema),
+				schema: this.parseSchema(this.props.schema),
 			});
 		}
 	}
--- a/src/unstable/components/JellyForm/index.tsx
+++ b/src/unstable/components/JellyForm/index.tsx
@@ -48,20 +48,20 @@ class JellyForm extends React.Component<JellyFormProps, JellyFormState> {
 		};
 	}
 
-	componentWillReceiveProps(nextProps: JellyFormProps) {
-		if (!isEqual(this.props.value, nextProps.value)) {
+	componentDidUpdate(prevProps: JellyFormProps) {
+		if (!isEqual(this.props.value, prevProps.value)) {
 			this.setState({
-				value: runFormulas(this.state.schema, nextProps.value),
+				value: runFormulas(this.state.schema, this.props.value),
 			});
 		}
 
 		if (
-			!isEqual(this.props.schema, nextProps.schema) ||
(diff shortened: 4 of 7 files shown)


alexandrtovmach:
@sradevski @LucianBuzzo @dimitrisnl 

Could you take a look?

src/components/Filters/index.tsx
```diff
@@ -89,23 +89,23 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
 		this.state.edit.push(this.getCleanEditModel());
 	}
 
-	componentWillReceiveProps(nextProps: FiltersProps) {
+	componentDidUpdate(prevProps: FiltersProps) {
```
LucianBuzzo: Can you change this to use `getDerivedStateFromProps`? `componentDidUpdate` triggers after a render, so if state is set from this lifecycle hook, then we end up triggering another render unnecessarily.

src/components/Table/index.tsx
```diff
@@ -138,9 +138,9 @@ class Table<T> extends React.Component<TableProps<T>, TableState<T>> {
 		};
 	}
 
-	componentWillReceiveProps(newProps: TableProps<T>) {
-		if (newProps.sort !== this.props.sort) {
-			this.setState({ sort: newProps.sort } as TableState<T>);
+	componentDidUpdate(prevProps: TableProps<T>) {
```
LucianBuzzo: As above, can you please use `getDerivedStateFromProps` here?

src/unstable/components/Form/index.tsx
```diff
@@ -91,16 +91,16 @@ export default class FormHOC extends React.Component<
 		widgets[name] = value;
 	}
 
-	componentWillReceiveProps(nextProps: FormProps) {
-		if (!isEqual(this.props.value, nextProps.value)) {
+	componentDidUpdate(prevProps: FormProps) {
```
LucianBuzzo: As above, can you please use `getDerivedStateFromProps` here?

src/unstable/components/JellyForm/index.tsx
```diff
@@ -48,20 +48,20 @@ class JellyForm extends React.Component<JellyFormProps, JellyFormState> {
 		};
 	}
 
-	componentWillReceiveProps(nextProps: JellyFormProps) {
-		if (!isEqual(this.props.value, nextProps.value)) {
+	componentDidUpdate(prevProps: JellyFormProps) {
```
LucianBuzzo: This is another place that can use `getDerivedStateFromProps`

alexandrtovmach:
@LucianBuzzo Hey, sorry for delay. I just now started looking on your comments and I think it's not a right places to use `getDerivedStateFromProps`.

From docs:
> This method exists for rare use cases where the state depends on changes in props over time.

Sounds like it works for us, but ― it works only in cases where we store previous props in component state.

alexandrtovmach:
React blog example:
```
class EmailInput extends Component {
  state = {
    email: this.props.defaultEmail,
    prevPropsUserID: this.props.userID
  };

  static getDerivedStateFromProps(props, state) {
    // Any time the current user changes,
    // Reset any parts of state that are tied to that user.
    // In this simple example, that's just the email.
    if (props.userID !== state.prevPropsUserID) {
      return {
        prevPropsUserID: props.userID,
        email: props.defaultEmail
      };
    }
    return null;
  }

  // ...
}
```

LucianBuzzo:
@alexandrtovmach Regarding your quote from the docs

> This method exists for rare use cases where the state depends on changes in props over time.

This is precisely what is happening in these components. The state is being derived from changes to props over time and we want to make sure that variance to the component props correctly affect the internal state of the component.

alexandrtovmach:
@LucianBuzzo Yes, but we also need to have previous props stored somewhere to compare it with a new one. If we are going to use `componentDidUpdate` we can get it as an argument of the function, but with `getDerivedStateFromProps` we need to store it somewhere...

alexandrtovmach:
@LucianBuzzo So, what is the final decision?)

LucianBuzzo:
Hi @alexandrtovmach apologies for the lack of activity, I've been on annual leave for the last 10 days and enjoying being away from my computer :smile: I'll take another look through this PR in the next few days :+1:

LucianBuzzo:
Hey @alexandrtovmach I went back through the PR and I think your approach is the best way for now. I incorrectly thought that `this.props` is available inside `getDerivedStateFromProps`. I'm concerned that we might be triggering unnecessary render cycles in a few places, however I'm happy to address this separately, if and when the render cycles become an issue.

LucianBuzzo:
@alexandrtovmach Could you please rebase your PR and add the `Connects-to` and `Change-type: patch` fields to the git commit body?

alexandrtovmach:
@LucianBuzzo Done

sradevski:
@LucianBuzzo don't merge this just yet, I want to check if the re-renders will affect the Table performance in a significant way.

src/components/Table/index.tsx
sradevski: This one can easily be done using `getDerivedStateFromProps`, something like:
```
static getDerivedStateFromProps(nextProps, prevState){
        if (nextProps.sort !== prevState.list) {
            return { sort: nextProps.sort };
        }
        
      return null;
    }
```

This also means we can add a check if state is undefined and set the default value that is currently set in the constructor.

alexandrtovmach: `getDerivedStateFromProps` havent `prevProps`, and that's why we can't use it

alexandrtovmach:
@sradevski Sorry for delay. I'm trying to use `getDerivedStateFromProps` for _src/components/Table/index.tsx_, but faced with issue about typings: https://github.com/Microsoft/TypeScript/issues/24018

I also have problems with tests. Could you please take a look

src/components/Table/index.tsx
```diff
@@ -155,9 +161,15 @@ class Table<T> extends React.Component<TableProps<T>, TableState<T>> {
 		};
 	}
 
-	componentWillReceiveProps(newProps: TableProps<T>) {
-		if (newProps.sort !== this.props.sort) {
-			this.setState({ sort: newProps.sort } as TableState<T>);
+	// trick for static method with generic: https://github.com/Microsoft/TypeScript/issues/24018
+	static getDerivedStateFromProps<S extends TableField>(
```
sradevski: All you need to do is change this to:
```
	static getDerivedStateFromProps<T>(
		nextProps: TableProps<T>,
		prevState: TableProps<T>,
	) {
		if (nextProps.sort !== prevState.sort) {
			return { sort: {...nextProps.sort} };
		} else {
			return null;
		}
	}
```

and get rid of the additional type you defined above.

alexandrtovmach:
Tests above are failed for me:

> Table component › columns property › column.sortable property › should sort by the specified sortable column on click
> Table component › columns property › column.sortable property › should reverse the sort order when clicking sortable column twice
> Table component › columns property › column.sortable property › should reverse the sort order when clicking sortable column twice
> Table component › columns property › column.sortable property › should allow a custom sorting function
> Form component › value property › should set the value if the value is changed after the component mounts
> JellyForm component › value property › should set the value if the value is changed after the component mounts

alexandrtovmach:
@sradevski Could you try on your machnie?

src/components/Table/index.tsx
```diff
@@ -155,9 +156,14 @@ class Table<T> extends React.Component<TableProps<T>, TableState<T>> {
 		};
 	}
 
-	componentWillReceiveProps(newProps: TableProps<T>) {
-		if (newProps.sort !== this.props.sort) {
-			this.setState({ sort: newProps.sort } as TableState<T>);
+	static getDerivedStateFromProps<T>(
```
sradevski: I tried this out and it seems we cannot use `getDerivedStateFromProps` properly, because it is also called when you call `setState` and we cannot know which one we should set as the sort value (there isn't a single source of truth). After all, it seems the best approach for now is to use 
```
	componentDidUpdate(prevProps: TableProps<T>) {
		if(this.props.sort && prevProps.sort !== this.props.sort){
			this.setState({sort: this.props.sort});
		}
	}
```
instead, I shouldn't have insisted on using `getDerivedStateFromProps` without trying it out properly.

alexandrtovmach: As I said from the beginning =)

alexandrtovmach:
@sradevski Thank you for help

Tests passed and hooks updated, seems like this PR ready =)

alexandrtovmach:
@sradevski done

sradevski:
@alexandrtovmach It seems you haven't pulled the latest changes from master and rebased on top of it instead of merging master into your branch, can you please do that and push again? Thanks

sradevski:
@balena-ci retest

sradevski:
@alexandrtovmach it seems the footer for the first commit is not valid (you need to have a newline between the title and the footer), you can check other PRs for reference if you want. Also it would be nice to squash the merge commit (you can use `git rebase -i HEAD~2`) and then just squash the merge commit. I know this is all a bit annoying, but it is required in order for our automation to work correctly.

alexandrtovmach:
@sradevski Sorry for delay, I was on vacation. Please check it now

sradevski:
@alexandrtovmach I wouldn't expect any visual changes for this PR, can you rerun the screenshot and snapshot generation please?

alexandrtovmach:
@sradevski seems like merge artifacts

sradevski:
@balena-ci retest

sradevski:
@alexandrtovmach thanks for the PR, merging now!

alexandrtovmach:
https://gph.is/g/aeYmVka

alexandrtovmach:
@sradevski thank you for a lot effort and help

LucianBuzzo:
Thanks for the contribution @alexandrtovmach !

PR #984, opened 2019-09-30
button: Increase styles specificity

dimitrisnl:
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -62,85 +62,107 @@ const ButtonBase = styled(Button)`
 `;
 
 const ColouredButton = styled(ButtonBase)`
-	color: ${props =>
-		isLight(getColor(props, 'color', 'main'))
-			? props.theme.colors.text.main
-			: 'white'};
-	${props => {
-		const color = getColor(props, 'color', 'dark');
-		return getHoverEffectOverride(
-			color,
-			isLight(color) ? props.theme.colors.text.main : 'white',
-		);
-	}};
+	& {
+		color: ${props =>
+			isLight(getColor(props, 'color', 'main'))
+				? props.theme.colors.text.main
+				: 'white'};
+		${props => {
+			const color = getColor(props, 'color', 'dark');
+			return getHoverEffectOverride(
+				color,
+				isLight(color) ? props.theme.colors.text.main : 'white',
+			);
+		}};
+	}
 `;
 
 const Outline = styled(ButtonBase)<{ color?: string }>`
-	color: ${props => props.theme.colors.text.main};
-	border-color: ${props => props.color || props.theme.colors.text.main};
-	svg {
-		color: ${props => props.theme.colors.tertiary.main} !important;
+	& {
+		color: ${props => props.theme.colors.text.main};
+		border-color: ${props => props.color || props.theme.colors.text.main};
+		svg {
+			color: ${props => props.theme.colors.tertiary.main} !important;
+		}
+		${props =>
+			getHoverEffectOverride(
+				props.color || props.theme.colors.text.main,
+				isLight(props.color) ? props.theme.colors.text.main : 'white',
+			)};
 	}
-	${props =>
-		getHoverEffectOverride(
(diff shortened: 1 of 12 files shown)


dimitrisnl:
@sradevski This finally fixes the issue I have with the website

dimitrisnl:
Can't figure out why there is such a mismatch with generated snapshots & screens

src/components/Button/index.tsx
```diff
@@ -36,22 +36,24 @@ const getHoverEffectOverride = (
...
-	white-space: nowrap;
-	svg {
-		color: inherit !important;
-		font-size: 0.925em;
-	}
-	/* These rules cause consistent styling when the button is rendered as a link */
-	display: inline-flex;
-	align-items: center;
-	justify-content: center;
+	&& {
```
sradevski: Can you just add a comment why we are adding this?

dimitrisnl:
@sradevski I didn't notice the screens before. Do they look good now?

dimitrisnl:
As for your follow-up, no. I noticed the same behaviour in the `Banner` component.

dimitrisnl:
I'm going to try with having a single ampersand in a separate commit.
One issue I noticed, is that it overrides the styles that come via className.

dimitrisnl:
@balena-ci retest

dimitrisnl:
@sradevski any clues on why would store-npm fail here? It really makes me wonder.

sradevski:
@dimitrisnl if the commit hash is the same, it will fail (maybe it attempted publishing under the same commit hash twice). Making some small change and pushing again will fix the problem. Also, for local testing, I would recommend using `npm pack` instead.

dimitrisnl:
@sradevski It fails due to a TableRow component. I pushed new commits and noticed the same behaviour.

sradevski:
@dimitrisnl It seems the typings changed somewhere (React or Grommet). Can you override the `defaultValue` typings in the `Checkbox` component to match whatever it asks for?

dimitrisnl:
It doesn't make sense to have `number` as a default value for a checkbox. Unless I'm reading it wrong.

sradevski:
@dimitrisnl for a checkbox it doesn't make sense, but there is no `HTMLCheckboxElement` type, only `HTMLInputElement` unfortunately, so it applies to any kind of input.

src/components/Checkbox/index.tsx
```diff
@@ -60,9 +60,10 @@ const Checkbox = ({ className, ...otherProps }: InternalCheckboxProps) => {
 };
 
 export interface InternalCheckboxProps
-	extends DefaultProps,
+	extends Omit<DefaultProps, 'defaultValue'>,
```
sradevski: Sorry about this, but what I think should have happened, is in `TableRow`, `checkboxAttributes` should be of `CheckboxProps` type instead, which will resolve this issue and make it more accurate.

dimitrisnl: Hm, still doesn't work. Wondering how did the last PR get through.

sradevski: @dimitrisnl You also need to update `Table.index`, as the `checkboxAttributes` comes from there.

dimitrisnl:
Great! The single `&` does the trick. I noticed some changes in the local snapshots that might help us understand the issue further.

Will document it, update the screenshots/snapshots and push again.

dimitrisnl:
@sradevski If you're good, I can merge.

sradevski:
@dimitrisnl Some of the screenshots are a bit weird (check the Tags or Steps one for example), so something changed for sure. It seems it is only the `plain` button that is affected as far as I could tell. Maybe we relied on the CSS not being applied correctly until now? Let's get those fixed first, otherwise a lot of things will break.

dimitrisnl:
I see @sradevski Thanks for spotting that, I'll address them.
In the meantime I'll update the deployed storybook version.

PR #1009, opened 2019-10-31
<Button/> README.md typo fix: quartenary

m-podlesny:
Change-type: patch

---
##### Contributor checklist

**Couldn't run the .sh script in my environment**
**Couldn't run the .sh script in my environment**

##### Reviewer Guidelines
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,8 @@ For an interactive demo of all components, see https://balena-io-modules.github.
 - [Tabs](#tabs)
 - [Tag](#tag)
 - [Terminal](#terminal)
-- [TextWithCopy](#textwithcopy)
 - [Textarea](#textarea)
+- [TextWithCopy](#textwithcopy)
 - [Tooltips](#tooltips)
 - [Txt](#txt)
 - [Markdown](#markdown)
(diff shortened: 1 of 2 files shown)


mfazer:
Is that change impacts on screenshots or snapshots?
I'm sure about `<Button/>` component only.
Someone have to check and test other components too (the same typo many times).

sradevski (dismissed):
Hi there, thanks for the PR! 
The only thing that needs to be fixed is you need to squash your commits and add a change type and signed off messages in the footer of the commit, as described here: https://github.com/balena-io-modules/rendition/blob/master/docs/COMMIT-GUIDELINES.md

Also, it would be great if you can rename the rest of the typos in the README, as they all should be `quartenary`.

LucianBuzzo:
AFAIK `quarternary` is the correct spelling, and the theme prop is misspelt - we need to fix this, instead of adjusting the docs to the incorrect spelling

corysimmons:
You're both wrong unless I'm missing something? `quaternary` seems to be the correct spelling.

image

This should be changed throughout all the components and docs. It would be a major bump as well so any projects depending on it would need updated.

Happy to do this if the major bump doesn't bother anyone.

LucianBuzzo:
@corysimmons d'oh! You're right of course. I think the major version bump is fine

LucianBuzzo (approved):

PR #1011, opened 2019-11-05
Reduce the luminocity threshold in the isLight method

dimitrisnl:
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


dimitrisnl:
Also, if we had the marketing website palette included we would be able to catch this.

dimitrisnl:
What do you mean @sradevski ? - wasn't the case that it was not acceptable to have dark text for such light variants?

sradevski:
@dimitrisnl I think this is just because the brand colors from the marketing palette are quite light, and adjusting the luminosity threshold is just to make it work well for our use-case, and it is not necessarily a general best-practice. Anyway, I think it is fine personally, and we can always revert it if necessary, so just go for it.

PR #1023, opened 2019-11-28
Fix regression in Form recreating itself on every change

sradevski:
Fixes a regression introduced by https://github.com/balena-io-modules/rendition/pull/968 where the Form would recreate itself, making every edit of a value move the cursor to the end of the content.

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/unstable/components/Form/index.tsx
+++ b/src/unstable/components/Form/index.tsx
@@ -74,19 +74,29 @@ const FormWrapper = styled(Box)`
 	}
 `;
 
-export default class FormHOC extends React.Component<
-	FormProps,
-	{ value: any; schema: JSONSchema6 }
-> {
+const parseSchema = (schema: JSONSchema6) => {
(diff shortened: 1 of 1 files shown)


src/unstable/components/Form/index.tsx
```diff
@@ -74,44 +74,43 @@ const FormWrapper = styled(Box)`
...
+		if (!state) {
+			return { schema: parseSchema(props.schema), value: props.value };
+		}
 
-		this.state = {
-			value: props.value || (props.schema.type === 'array' ? [] : {}),
-			schema: this.parseSchema(props.schema),
-		};
+		if (!isEqual(state.value, props.value)) {
+			return {
```
sradevski: I would actually prefer for this to be a only controlled component without a local state, but this will do for now.

PR #1028, opened 2019-12-18
Filters: make full text search work with "type" array values

LucianBuzzo:
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Filters/SchemaSieve.spec.js
+++ b/src/components/Filters/SchemaSieve.spec.js
@@ -995,6 +995,25 @@ describe('SchemaSieve', () => {
 
       expect(ajv.compile(filter)).not.toThrow()
     })
+
+    it.only('should work when "type" is an array', () => {
+      const schema = {
+        type: 'object',
+        properties: {
+          foo: { type: ['string', 'null'] },
+          bar: { type: 'string' }
+        }
+      }
+
+      const value = {
+        foo: 'test',
+        bar: 'baz'
+      }
+
+      const filter = sieve.createFullTextSearchFilter(schema, 'test')
+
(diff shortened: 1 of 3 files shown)


src/components/Filters/SchemaSieve.ts
```diff
@@ -50,7 +51,10 @@ export const createFullTextSearchFilter = (
 	const stringKeys = reduce(
 		schema.properties,
 		(carry, item: JSONSchema6, key) => {
-			if (item.type === 'string') {
+			if (
+				item.type === 'string' ||
+				(isArray(item.type) && includes(item.type, 'string'))
```
sradevski: Use `item.type.includes` instead (and optionally `Array.isArray`).

src/components/Filters/SchemaSieve.spec.js
```diff
@@ -995,6 +995,25 @@ describe('SchemaSieve', () => {
 
       expect(ajv.compile(filter)).not.toThrow()
     })
+
+    it.only('should work when "type" is an array', () => {
+      const schema = {
+        type: 'object',
+        properties: {
+          foo: { type: ['string', 'null'] },
```
sradevski: Does this mean that `foo` would be an array of strings or nulls? In which case would it have null values?

LucianBuzzo: @sradevski this means that `foo` can be of type `string` *or* `null`. A nice explanation can be found here https://json-schema.org/understanding-json-schema/reference/type.html#type

Issue #1017, opened 2019-11-22
Tags: hidden fields are applied to Tag Filter


gffrd:
Spawned out of Dashboard issue [#3189](https://github.com/balena-io/balenacloud-dashboard/issues/3189) (FYI @thgreasi)

----

##  Expected Behavior
When I create a Tag Filter, I expect it to only filter on the fields I see when creating the filter.

## Actual Behavior
The content of fields hidden when selecting a different operator (going from TAG IS to TAG NAME CONTAINS, for example) is included in the filter.

## Steps to Reproduce the Problem
(ideally, apply a few tag name+values to devices, then use them in creation of filters so you can see the effect)
1. From the Application Device list, add a new Filter for a Tag
2. Ensure Operator is set as "IS," and enter a Name and Value
3. Change Operator from IS to VALUE CONTAINS (Name field disappears)
4. Save filter
5. Look at applied Filter and results—it will be for NAME:VALUE, instead of implied VALUE only, and include only results for a complete match, and not just the intended Value match (ignoring Name).

## Specifications
- Browser: whatsmybrowser.org/b/UW9983P
- OS: MacOS 10.14.6
- Environment: production, staging

## Screenshots
![Screen Recording 2019-11-20 at 12 06 58 PM](https://user-images.githubusercontent.com/7716996/69285366-9aa16800-0ba5-11ea-93aa-8b1217a6bf6a.gif)
(Non-GIF capture [here](https://www.dropbox.com/s/jilf67ceapgtdal/Dash-TagFilter-HiddenField-2_3.mp4?dl=0))

## References
N/A

PR #1030, opened 2019-12-19
fix(Filters): hidden fields are applied to Tag Filter

iamsolankiamit:
Fixes: https://github.com/balena-io-modules/rendition/issues/1017

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/DataTypes/object.tsx
+++ b/src/components/DataTypes/object.tsx
@@ -5,6 +5,7 @@ import findKey from 'lodash/findKey';
 import includes from 'lodash/includes';
 import isString from 'lodash/isString';
 import mapValues from 'lodash/mapValues';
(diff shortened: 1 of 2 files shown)


src/components/DataTypes/object.tsx
```diff
@@ -194,6 +195,15 @@ export const createFilter = (
 	schema: JSONSchema6,
 ): JSONSchema6 => {
 	const { title } = schema;
+
+	if (operator.includes('key')) {
```
sradevski: I have a bit of a hard time understanding the solution here, as it feels quite hacky to me. If you can't find a better approach, at the very least, I would extract this into a function that clearly says what this does.

PR #1031, opened 2019-12-26
Add HighlightedName component

corysimmons:
See: https://app.zeplin.io/project/5cd17644e6012f68816cca3b/screen/5d62783b38345e9b5a66e786
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


README.md
```diff
@@ -432,6 +433,17 @@ example `<Heading.h1>`, `<Heading.h2>`, `<Heading.h3>` etc.
...
 
+### HighlightedName
+
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/HighlightedName/story.js)
+
+#### Props
+
+| Name   | Type   | Default   | Required   | Description   |
+| ------ | ------ | --------- | ---------- | ------------- |
+| `shade`        | `number` | - | - | The index for one of the pre-determined shades for the highlighted name |
```
sradevski: I think I wasn't clear when I was talking about the component, what I meant is the `Highlighted Name` component should autogenerate the color from its content, and for a start, we can only have the `children` as the only prop for the component. What I meant is, the color autogeneration feature should be very similar to what we have in the current `Badge` component (not the new `Badge` that I submitted a PR for). Let me know if anything else is unclear

corysimmons:
Can't run `npm run generate-screenshots` without https://github.com/balena-io-modules/rendition/pull/1032 getting merged first, otherwise it will clump those out-of-sync screenshots in with this component's screenshots. Once 1032 gets merged, I'll add another commit to this PR.

sradevski:
@corysimmons IMO until we find the cause for the flakiness of those tests, or have a better comparison method (maybe use Percy?) it is fine to include the flaky screenshots. Not ideal, but they shouldn't be a blocker.

corysimmons:
I suspect someone just didn't regenerate them when they were supposed to. Rather than try to track down the commit where it broke, I was just going to re-sync it, then if it gets out of sync again, it will be immediately evident where it was introduced (since I'll be generating screenshots multiple times a day).

I'll get to work tracking down where it was introduced though.

sradevski:
@corysimmons If you didn't generate them, the build should fail and it shouldn't allow you to merge, so I think it is flakiness that is the issue. Anyway, we will see if the issue reoccurs and we can investigate then.

corysimmons:
@sradevski Okay it generates the background colors using the same helper that's in `Core/Badge`. Can you please re-review when you get a chance?

corysimmons:
You were right. Screenshots immediately got out of sync again.

README.md
```diff
@@ -432,6 +433,27 @@ example `<Heading.h1>`, `<Heading.h2>`, `<Heading.h3>` etc.
...
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/HighlightedName/story.js)
+
+#### Props
+
+| Name          | Type      | Default | Required | Description                                |
+| ------------- | --------- | ------- | -------- | ------------------------------------------ |
+| `primary`     | `boolean` | -       | -        | If true, use the `primary` theme color     |
+| `secondary`   | `boolean` | -       | -        | If true, use the `secondary` theme color   |
+| `tertiary`    | `boolean` | -       | -        | If true, use the `tertiary` theme color    |
+| `quaternary`  | `boolean` | -       | -        | If true, use the `quaternary` theme color  |
```
sradevski: I would actually remove all props except for `children`. It is better to start with a minimum API, and add on top of it if truly necessary (since we already support `bg` and `color`).

src/components/HighlightedName/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+const backgroundColor = memoize((text: string): string => {
+	return colorHash.hex(text.replace(/\s/g, ''));
+});
+
+const BaseHighlightedName = styled(Txt.span)`
+	font-family: ${props => props.theme.titleFont};
+	display: inline-block;
+	border-radius: 2px;
+	line-height: 1;
+	color: ${props => props.color};
```
sradevski: I don't think this is necessary, as passing `color` to the `Txt` component should be enough

src/components/HighlightedName/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+`;
+
+const HighlightedName = ({
+	children,
+	className,
+	theme,
+	...props
```
sradevski: You can also add the `color` here, and change `props.color` below to just `color`

src/components/HighlightedName/index.tsx
corysimmons: When `isLight` returns true, it returns `#fff`, but the `#fff` as the `color` prop. But the color prop never gets fed to the styled component. I think this is broken in the old Badge component.

src/components/HighlightedName/index.tsx
corysimmons: I agree, but it seemed like the convention was to do `props.foo` for most stuff rather than destructure so I went with that. I can (and would prefer to) destructure everything though.

corysimmons:
@sradevski Ready for review again.

LucianBuzzo:
LGTM, It is really odd that the new screenshots are just a few pixels longer though

PR #1045, opened 2020-01-17
Add a "position" prop to the Modal

sradevski:
Allowing the modal to be top-aligned is better
when the modal can expand vertically. If center-aligned
the modal expands in both directions, resulting in a bad
jumping UX. Setting the modal as top-aligned eliminates that
as the modal will only expand in the -y direction

Any comments or objections?

Change-type: minor

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


LucianBuzzo:
@sradevski what are your thoughts on making the modal *always* top align?

sradevski:
@LucianBuzzo I find it center-aligned to look much better, and is better UX-wise as you would expect the main content (which would be the Modal) to be in the middle of your screen. The only time when top-aligned makes sense is when the modal content grows dynamically. Let's also talk with the designers and see what they think.

LucianBuzzo:
@sradevski it seems to me that there are specific situations where you should always choose to top align the modal, otherwise the modal should always be center aligned. Is it possible to have the modal automatically use the best layout? Otherwise we are sort of setting a trap!

sradevski:
@LucianBuzzo yeah it is, and I don't really like it, and I was expecting some suggestions TBH 😄 . I was just thinking a bit more about this, and I think I found a great solution. The modal jumping on resize can also be eliminated if we animate the modal resize, which will appear much smoother. I will try it out tomorrow and see how it goes.

sradevski:
It turns out, animating auto-calculated height is not supported, so it will need a bit more work. I will close this PR in favor of this issue: https://github.com/balena-io-modules/rendition/issues/1049

PR #1046, opened 2020-01-17
Improve DropDownButton Component

StefKors:
This PR improves the DropDownButton Component with the following things:

- Support for bg prop
- Dropdown items now have hover styles

Change-type: patch
Signed-off-by: Stef Kors <stef@balena.io>

---
##### Contributor checklist

##### Reviewer Guidelines
---


StefKors:
@balena-ci retest

StefKors:
@balena-ci retest

StefKors:
@balena-ci retest

src/components/DropDownButton/index.tsx
```diff
@@ -68,17 +68,23 @@ const Item = styled.div<{ border: boolean }>`
 	border-top: ${props =>
 		props.border ? `1px solid ${props.theme.colors.gray.main}` : '0'};
 	border-radius: ${props => px(props.theme.radius)};
+	background: white;
```
grahammcculloch: Should this be a theme color?

src/components/DropDownButton/index.tsx
```diff
@@ -68,17 +68,23 @@ const Item = styled.div<{ border: boolean }>`
...
+		cursor: pointer;
+	}
 `;
 
 const JoinedButton = styled(Button)`
 	margin: 0;
 	width: 100%;
 `;
-
+// {..._.without('bg', props)}
```
grahammcculloch: This line needs deleting rather than commenting out I think?

src/components/DropDownButton/index.tsx
```diff
@@ -203,12 +209,12 @@ class DropDownButton extends React.Component<
...
 			<Wrapper
 				className={className}
-				{...props}
+				style={{ background: 'none' }}
+				{...omit(props, 'bg')}
 				ref={node => (this.dropdownNode = node)}
 			>
 				{joined ? (
 					<Toggle
-						{...props}
```
grahammcculloch: I haven't looked into the code-base for this but are you sure that this component is never used by someone who is relying on a prop being passed through here? This _looks_ like a breaking change to me (albeit probably a very insignificant one).

StefKors:
@sradevski I've added the custom background prop for use in the CRM opportunities view. This is because we don't have a wide spectrum of colors available and I wanted to have differences between the different stages:

Screenshot 2020-01-27 at 09 23 00

Screenshot 2020-01-27 at 09 24 57

I would love to solve this using the theme colours, though that might also not be appropriate. Any suggestions?

src/components/DropDownButton/index.tsx
StefKors: @grahammcculloch Above here on line 41 the background is also set directly. shall I also adjust that one?

grahammcculloch: @StefKors ahh - having taken the time to look through the Rendition code-base now I see that 'white' is used all over the place! So this would be part of a larger clean-up/refactor problem and I would ignore my comment above for this PR!

src/components/DropDownButton/index.tsx
StefKors: agree, I'll clean this up

sradevski:
@StefKors It looks like using `Select` with `Badge` would be more appropriate, at least from the looks of it. What happens if you click on the left part of the dropdown? Intuitively, that shouldn't do anything, so maybe we need to rethink the design, rather than modify the component.

src/components/DropDownButton/index.tsx
StefKors: I agree, I'll return this prop. I think it got removed during testing.

StefKors:
@sradevski Agree if staying with the dropdown button it should be a joined version.
A `select` with a `badge` would be an option. I would use it like this then: 

Screenshot 2020-01-27 at 10 40 51

sradevski:
@StefKors yeah that makes much more sense, also UX wise it is clear what it does (the dropdown has a different purpose, which honestly is a bit vague and we need to rethink it)

StefKors:
@sradevski cool, I'll use that and I'll close this PR. As it seems just modifying the dropdown color when Rendition updates are near is unnecessary.

Issue #1048, opened 2020-01-21
Add a Spinner component


sradevski:
This is a must-have component that we use all over the place.

Ideally, the component can be used with both with and without `children`.
- If used without `children`, just render a spinner wherever the component is rendered
- If used with `children`, create an overlay over the `children`, and show the spinner in the center.

We might want to have 2 sizes, but that is open to discussion.

PR #1051, opened 2020-01-24
Add a spinner component

sradevski:
I'd like to try it out in the dashboard before merging, but I would really like to hear opinions on the API and if you would do something differently. The icon is based on this design: https://app.zeplin.io/project/5be48160a58b413fdb05729a/screen/5e21aeaff9536869fa02df8f

Connects-to: #1048
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Spinner/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Spinner/story.js)
+
+## Props
+
+| Name          | Type      | Default   | Required   | Description                                          |
+| ------ | ------ | --------- | ---------- | ------------- |
+| `children`     | `React.ReactNode` |         | -          | If the spinner has children, it will show an opaque overlay over the children when it is spinning |
+| `spinning`     | `boolean`         | true    | -          | If passed, it will control whether the spinner is shown or not |
+| `emphasized`   | `boolean`         | false   | -          | If true, it will render a large spinner |
```
iamsolankiamit: We should try to keep props similar or intuitive. So instead of emphasised, we can provide a `size` prop which can have some levels `sm | md | lg`, or a `number`.

src/components/Spinner/README.md
```diff
@@ -0,0 +1,12 @@
+# Spinner
+
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Spinner/story.js)
+
+## Props
+
+| Name          | Type      | Default   | Required   | Description                                          |
+| ------ | ------ | --------- | ---------- | ------------- |
+| `children`     | `React.ReactNode` |         | -          | If the spinner has children, it will show an opaque overlay over the children when it is spinning |
+| `spinning`     | `boolean`         | true    | -          | If passed, it will control whether the spinner is shown or not |
```
iamsolankiamit: this can be renamed as hidden. `spinning: false` sounds like it is rendered but it won't be spinning.

grahammcculloch: ...or `active`? I'm not too familiar with Rendition standards but 'positive' boolean properties are usually a bit more intuitive than 'negative' boolean ones (with the notable exception of 'disabled'). 

To be honest I read the description and assumed `spinning` meant that, if set, the spinner would be spinning (rotating) _in addition to_ the label and any other spinner UI.

src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from 'react';
+import styled, { keyframes } from 'styled-components';
+import { DefaultProps, RenditionSystemProps } from '../../common-types';
+
+import asRendition from '../../asRendition';
+import { px } from '../../utils';
+import { Flex } from '../Flex';
+import Txt from '../Txt';
+
+const rotate360 = keyframes`
```
grahammcculloch: Is there a common file for 'standard' definitions like this?

src/components/Spinner/README.md
grahammcculloch: ...or `large` (boolean, default false)?

sradevski: @iamsolankiamit @grahammcculloch The reason I went with `emphasized` is we used that terminology in other components (`Input`, `Select`, and `ProgressBar` so far), so it is just for consistency. About the multiple values, that's not something we want to support. We want to be very opinionated and not allow users to specify whatever value they want, so a boolean works better. We might add it to the theme if necessary, but I would rather do it only when we see a real need for it, and not just for the sake of adding it.

src/components/Spinner/README.md
sradevski: @iamsolankiamit @grahammcculloch I just borrowed the naming from a ant design, but I agree that it can be made better. I think `show` is the most descriptive in this case, what you think?

src/components/Spinner/index.tsx
sradevski: @grahammcculloch not yet, we haven't put much focus on animations so far. I will be cleaning up rendition in the following months, so some of the common parts will be refactored (not necessary yet).

src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		return (
+			<SpinnerBase label={label} emphasized={emphasized} {...otherProps} />
+		);
+	}
+
+	return (
+		<Container {...otherProps}>
+			{spinning ? (
```
LucianBuzzo: This can use a `&&` operator instead of a ternary operator

sradevski:
@grahammcculloch @LucianBuzzo can you guys re-check this PR? thanks

PR #1052, opened 2020-01-28
Fix filter using is/is_not operator for objects

iamsolankiamit:
Change-type: Patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


LucianBuzzo:
@iamsolankiamit What does this fix? Does there need to be a regression test added?

src/components/Filters/spec.js
```diff
@@ -133,6 +164,34 @@ describe('Filters component', () => {
       expect(component.find(FiltersSummary)).toHaveLength(0)
       component.unmount()
     })
+
+    it('should filter using "is" operator for both key and value properties of an object', () => {
```
LucianBuzzo: Contrary to the test title, this test doesn't actually check that the filtering behaves as expected, it just checks that the label is present - can you add a test for schemasieve that asserts that the filtering behaves as expected?

src/components/Filters/spec.js
```diff
@@ -133,6 +193,42 @@ describe('Filters component', () => {
...
+        </Provider>
+      )
+      expect(
+        component
+          .find('button')
+          .at(4)
+          .text()
+      ).toBe('Tag is rarity : 10')
+
+      const filteredByTagIs = SchemaSieve.filter(tagIsFilter, pokedex)
```
LucianBuzzo: There is a separate spec file for the filtering mechanism here https://github.com/balena-io-modules/rendition/blob/master/src/components/Filters/SchemaSieve.spec.js can you move this test into that file?

src/components/Filters/spec.js
```diff
@@ -133,6 +193,42 @@ describe('Filters component', () => {
...
+        </Provider>
+      )
+      expect(
+        component
+          .find('button')
+          .at(4)
+          .text()
+      ).toBe('Tag is not rarity : 10')
+
+      const filteredByTagIsNot = SchemaSieve.filter(tagIsNotFilter, pokedex)
```
LucianBuzzo: Could you move this as well please?

src/components/Filters/spec.js
LucianBuzzo: One last nitpick, these test names don't match what the test is doing, could you please change it to something like
```suggestion
    it('should  show the correct label when filtering using "is" operator for both key and value properties of an object', () => {
```

src/components/Filters/spec.js
```diff
@@ -133,6 +164,34 @@ describe('Filters component', () => {
...
+      )
+      expect(
+        component
+          .find('button')
+          .at(4)
+          .text()
+      ).toBe('Tag is rarity : 10')
+    })
+
+    it('should filter using "is_not" operator for both key and value properties of an object', () => {
```
LucianBuzzo: Same as above: could you make the test name reflect what is being tested?

src/components/Filters/spec.js
```diff
@@ -133,6 +164,34 @@ describe('Filters component', () => {
       expect(component.find(FiltersSummary)).toHaveLength(0)
       component.unmount()
     })
+
+    it('should  show the correct label when filtering using "is" operator for both key and value properties of an object', () => {
```
LucianBuzzo: Double whitespace
```suggestion
    it('should show the correct label when filtering using "is" operator for both key and value properties of an object', () => {
```

src/components/Filters/spec.js
```diff
@@ -133,6 +164,34 @@ describe('Filters component', () => {
...
+      )
+      expect(
+        component
+          .find('button')
+          .at(4)
+          .text()
+      ).toBe('Tag is rarity : 10')
+    })
+
+    it('should  show the correct label when filtering using "is_not" operator for both key and value properties of an object', () => {
```
LucianBuzzo:
```suggestion
    it('should show the correct label when filtering using "is_not" operator for both key and value properties of an object', () => {
```

Issue #1054, opened 2020-02-03
Find an alternative to animate.css module


thundron:
It goes fine through compilation but then errors at runtime whilst being underused in the code base.
We should look into a simpler solution (probably just plain css animations), given that we only use it for notifications as of now

PR #1056, opened 2020-02-04
Replace animate.css with keyframes and classes

thundron:
Fixes: #1054
Change-type: patch
Signed-off-by: Lorenzo Alberto Maria Ambrosi <lorenzothunder.ambrosi@gmail.com>

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/animations.ts
+++ b/src/animations.ts
@@ -1,0 +1,35 @@
+import { css, keyframes } from 'styled-components';
+
+const fadeIn = keyframes`
+	from {
+		opacity: 0;
+	}
+
+	to {
+		opacity: 1;
+	}
+`;
+
+const fadeOut = keyframes`
+	from {
+		opacity: 1;
+	}
+
+	to {
+		opacity: 0;
+	}
+`;
+
+export const animations = css`
+	.fadeIn {
+		animation: ${fadeIn};
+	}
+
+	.fadeOut {
+		animation: ${fadeOut};
+	}
+
+	.animated.faster {
+		animation-duration: 500ms;
+	}
+`;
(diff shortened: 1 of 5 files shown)


src/animations.ts
```diff
@@ -0,0 +1,35 @@
+import { css } from 'styled-components';
+
+export const animations = css`
```
sradevski: You can also use `keyframes`, I believe (same below): https://styled-components.com/docs/basics#animations
```suggestion
export const animations = keyframes`
```

thundron:
@balena-ci retest

src/components/Notifications/index.tsx
```diff
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import ReactNotification, { store } from 'react-notifications-component';
-import 'react-notifications-component/dist/theme.css';
```
sradevski: Are the styles below everything that is in the theme?

thundron: No they're just for the storybook as I was doing some tests on Etcher too. WRT porting them all: should I make a file inside the Notification component with the theme? Feels better than a `styles.ts` in the src folder or writing them inside index.ts of the component, which would make it quite uncomfortable to read

thundron: No, I was just testing stuff in the storybook + Etcher, including them in a separate commit for semplicity

sradevski:
@thundron Let's add https://github.com/euheny/babel-plugin-transform-css-import-require-to-string instead of copy-pasting CSS. We actually have https://github.com/balena-io-modules/rendition/blob/master/scripts/tsifyCss.js to do similar things for the terminal and the MD editor, but we can remove that script as well and use the babel transform instead.

thundron:
Sure, on it

PR #1061, opened 2020-02-12
Replace animate.css with simple animations

thundron:
Include react-notifications-component's theme.css in a string as it causes some trouble when using rendition without bundling it

Connects-to: #1054 
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


thundron:
I have some trouble with the `npm run generate-screenshots` step, it hangs and then fires a timeout error

src/components/Notifications/index.tsx
```diff
@@ -74,6 +75,10 @@ const NotificationContainer = ({
 };
 
 export const NotificationsContainer = styled(ReactNotification)`
+	${animations}
+
```
sradevski: Let's remove the whitespace here.

PR #1063, opened 2020-02-13
Make Tooltips Accessible

iamsolankiamit:
Change-type: Minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Badge/index.tsx
```diff
@@ -46,7 +46,12 @@ const BaseBadge = styled(Txt.span)`
 	line-height: 1;
 `;
 
-const Badge = ({ children, className, shade }: ThemedBadgeProps) => {
+const Badge = ({
```
LucianBuzzo: This refactor should be in a separate commit, or even better, a separate PR.

src/tooltips.tsx
```diff
@@ -13,25 +13,6 @@ const TARGET_OFFSET = 3;
 // The time in ms to show a tooltip after being triggered by a click event
 const CLICK_TIMEOUT = 1000;
 
-const createOnRemoveObserver = (
```
LucianBuzzo: This refactor should be in a separate commit, or even better, a separate PR.

iamsolankiamit: @LucianBuzzo This refactor is required, in the previous implementation we only rendered one tooltip component and updated it's content based on the hovered/clicked component. for accessibility changes to work, we need to render all the tooltips in separate component and connect them to their references using id and aria-describedby.

src/components/Badge/index.tsx
iamsolankiamit: Sure, I can move it into a separate commit.

src/tooltips.tsx
LucianBuzzo: @iamsolankiamit My comment stands, if it is a pre-requisite for making the tooltips accessible, then it should be made in a sperate commit and labeled as such

iamsolankiamit:
closing this in favour of #1068

PR #1066, opened 2020-02-17
Fix Tooltip display for badges

iamsolankiamit:
Change-type: Patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Badge/index.tsx
+++ b/src/components/Badge/index.tsx
@@ -39,7 +39,12 @@ const BaseBadge = styled(Txt.span)`
 	line-height: 1;
 `;
 
-const Badge = ({ children, className, shade }: ThemedBadgeProps) => {
+const Badge = ({
+	children,
+	className,
+	shade,
+	...otherProps
+}: ThemedBadgeProps) => {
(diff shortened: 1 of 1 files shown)


src/components/Badge/index.tsx
```diff
@@ -55,6 +60,7 @@ const Badge = ({ children, className, shade }: ThemedBadgeProps) => {
 			px="12px"
 			fontSize={0}
 			bg={shadeHex}
+			{...otherProps}
```
sradevski: I am a bit hesitant to pass down everything, as sometimes we pass down things that we didn't expect to pass, leading to all sorts of bugs and inconsistencies. What exactly do we need to pass for the tooltips to work? Also, if we do this, it should be at the top, so all the custom props we pass override whatever the user passes (so the badge stays consistent and users can't mess with it).

iamsolankiamit: @sradevski we need `onMouseLeave`, `onMouseEnter`, `onClick`(for tooltip which open on click), and `aria-describedby` for the tooltip to work. I can only take them if you agree with that.

PR #1068, opened 2020-02-19
Replace tooltip with Rc-Tooltip

iamsolankiamit:
Change-type: Minor

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/tooltip.tsx
```diff
@@ -0,0 +1,151 @@
...
+import merge from 'lodash/merge';
+import * as React from 'react';
+import styled from 'styled-components';
+import { TooltipPlacement } from './common-types';
+import RCTooltip from 'rc-tooltip';
+
+// The time in ms to show a tooltip after being triggered by a click event
+const CLICK_TIMEOUT = 1000;
+
+const Arrow = styled.div`
```
sradevski: Are these from the arrow from the previous tooltip? Let's just use some library or whatever is default from rc-tooltip, and improve it later. I would really prefer to keep the codebase size to a minimum, especially for some trivial elements such as an arrow.

src/tooltip.tsx
```diff
@@ -0,0 +1,151 @@
...
+				trigger?: 'hover' | 'click';
+				placement?: TooltipPlacement;
+				containerStyle?: React.CSSProperties;
+				innerStyle?: React.CSSProperties;
+				arrowStyle?: React.CSSProperties;
+		  };
+	children: React.ReactElement;
+}
+
+const overlayStyles = {
```
sradevski: Maybe not all of these are necessary? We don't have to include everything from the previous tooltip, we just need to make the new one work 😄

src/tooltip.tsx
iamsolankiamit: @sradevski, sure inwhich case we will have to import rc-tooltips css too. I didn't do it, since we wanted to have our own design.

src/tooltip.tsx
iamsolankiamit: Umm Sure, will remove which seems un-ncessary

src/tooltip.tsx
sradevski: @iamsolankiamit we might want to check with Alexandra on an updated design for the tooltips, but let's just start with the defaults first. Let's make this functional first, and afterwards please chat with Alexandra on providing you with a design that would match the rest of the designs better.

LucianBuzzo:
@iamsolankiamit Are you still working on this? It would be cool to see it merged

package.json
```diff
@@ -136,6 +136,7 @@
     "marked": "^0.8.0",
     "mermaid": "8.4.0",
     "prop-types": "^15.7.2",
+    "rc-tooltip": "^4.0.0-alpha.3",
```
LucianBuzzo: Is there any indication of how stable their alpha version is?

iamsolankiamit:
@LucianBuzzo yes, I've been focusing on other issues. I'll fix this by tomorrow.

iamsolankiamit:
@sradevski I have a PR open on rc-trigger to handle the disabled hover cases. Once that is merged and rc-tooltip is updated, the disabled bug will be resolved.

sradevski:
@iamsolankiamit amazing stuff, ping me once it's all ready and merged.

sradevski:
For reference, here is the rc-trigger PR: https://github.com/react-component/trigger/pull/168

LucianBuzzo (approved):

JSReds:
not satisfying all use cases

PR #1069, opened 2020-02-19
Add usebreakpoint hook

JSReds:
See: [Flowdock discussion](https://www.flowdock.com/app/rulemotion/resin-frontend/threads/ATGc-ie1nUzCHGSbmrx8HMYIQoH)
Change-type: patch

##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Provider/BreakpointProvider.tsx
+++ b/src/components/Provider/BreakpointProvider.tsx
@@ -1,0 +1,48 @@
+import debounce from 'lodash/debounce';
+import * as React from 'react';
+import { BreakpointContext } from '../../contexts/BreakpointContext';
+export interface BreakpointProviderProps {
+	breakpoints: number[];
+	onBreakpointChange?: (breakpoint: number) => void;
+}
+
+export const BreakpointProvider: React.FC<BreakpointProviderProps> = ({
+	breakpoints,
+	children,
+}) => {
+	const [currentBreakpoint, setCurrentBreakpoint] = React.useState<number>(0);
+
+	React.useEffect(() => {
+		if (!breakpoints || !breakpoints.length) {
+			return;
+		}
+
+		const handleResize = () => {
+			const clientWidth = window.innerWidth;
+			const breakpointIndex = breakpoints.findIndex(
+				breakpoint => clientWidth < breakpoint,
+			);
+			const currentBreakpoint =
+				breakpointIndex >= 0 ? breakpointIndex : breakpoints.length - 1;
+			setCurrentBreakpoint(currentBreakpoint);
+		};
+
+		const debouncedResize = debounce(handleResize, 80);
+
+		handleResize();
+
+		window.addEventListener('resize', debouncedResize, { passive: false });
+
+		return () => {
+			window.removeEventListener('resize', debouncedResize);
+		};
+	}, [breakpoints]);
+
+	return (
+		<BreakpointContext.Provider value={{ currentBreakpoint }}>
+			{children}
+		</BreakpointContext.Provider>
+	);
+};
+
+BreakpointProvider.displayName = 'BreakpointProvider';
--- a/src/components/Provider/index.tsx
+++ b/src/components/Provider/index.tsx
@@ -5,6 +5,7 @@ import styled from 'styled-components';
 import { DefaultProps, Theme } from '../../common-types';
 import defaultTheme from '../../theme';
 import { px } from '../../utils';
+import { BreakpointProvider } from './BreakpointProvider';
 
 const Base = styled(Grommet)`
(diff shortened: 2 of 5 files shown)


src/components/Provider/BreakpointProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const BreakpointProvider: React.FC<BreakpointProviderProps> = ({
+	breakpoints,
+	children,
+}) => {
+	const [currentBreakpoint, setCurrentBreakpoint] = React.useState(
+		0 as Breakpoint,
+	);
+
+	const handleResize = () => {
```
sradevski: Let's move this inside the effect

src/components/Provider/BreakpointProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+		const clientWidth = window.innerWidth;
+		if (clientWidth < breakpoints[0]) {
```
sradevski: I am thinking, we probably want this to be more dynamic, as the length of the `breakpoints` array can change, and the hook should be able to handle it seamlessly.

src/components/Provider/BreakpointProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+			setCurrentBreakpoint(1);
+		} else if (clientWidth < breakpoints[2]) {
+			setCurrentBreakpoint(2);
+		} else if (clientWidth >= breakpoints[2]) {
+			setCurrentBreakpoint(3);
+		}
+	};
+
+	React.useEffect(() => {
+		const debouncedResize = debounce(handleResize, 80);
```
sradevski: Once we move handleResize in here and make it dynamic, before any of this is called, we should validate that `breakpoints` exists and that the length is not 0 (and maybe set some maximum as well).

src/hooks/useBreakpoint.ts
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { BreakpointContext } from '../contexts/BreakpointContext';
+
+export const useBreakpoint = <T extends any>(values: T[]): T => {
+	if (values.length > 4) {
```
sradevski: Same as in the provider, we should be more flexible in accepting a range of array sizes and dynamically handle it (for example between 1 and 6)

src/contexts/BreakpointContext.ts
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+
+export type Breakpoint = 0 | 1 | 2 | 3;
```
sradevski: Same like the rest, we can probably make it a bit more dynamic, so maybe a plain number is fine.

src/components/Provider/BreakpointProvider.tsx
```diff
@@ -1,44 +1,35 @@
...
-			setCurrentBreakpoint(2);
-		} else if (clientWidth >= breakpoints[2]) {
-			setCurrentBreakpoint(3);
-		}
-	};
+	const [currentBreakpoint, setCurrentBreakpoint] = React.useState<number>(0);
 
 	React.useEffect(() => {
-		const debouncedResize = debounce(handleResize, 80);
+		if (breakpoints && !!breakpoints.length) {
```
sradevski: Let's reduce nesting by checking the opposite condition and returning early (`if (!breakpoints..) return`)

JSReds:
@balena-ci retest

src/components/Provider/index.tsx
```diff
@@ -22,7 +23,11 @@ const Base = styled(Grommet)`
 `;
 
 const Provider = ({ theme, ...props }: ThemedProvider) => {
-	return <Base theme={merge(defaultTheme, theme)} {...props} />;
+	return (
+		<BreakpointProvider breakpoints={defaultTheme.breakpoints}>
```
sradevski: We should pass the breakpoints from the merged theme here, not the default one. You would extract `merge(defaultTheme, theme)` in a variable, and use that to pass the breakpoints.

src/hooks/useBreakpoint.ts
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { BreakpointContext } from '../contexts/BreakpointContext';
+import { breakpoints } from '../theme';
+
+export const useBreakpoint = <T extends any>(values: T[]): T => {
```
sradevski: Let's add a comment what this is and how it should be used.

src/components/Provider/BreakpointProvider.tsx
sradevski: This doesn't seem to have been resolved, let's return early here.

PR #1071, opened 2020-02-26
fix: build and bump typescript version

JSReds:
fix build null index error and bump typescript version

Change-type: patch
See: https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/unstable/components/Form/widgets/PasswordWidget.tsx
+++ b/src/unstable/components/Form/widgets/PasswordWidget.tsx
@@ -39,7 +39,9 @@ export interface PasswordStrengthProps {
 }
 
 const PasswordStrength = ({ password }: PasswordStrengthProps) => {
-	const [strengthScore, setStrengthScore] = React.useState();
+	const [strengthScore, setStrengthScore] = React.useState<
+		number | undefined
+	>();
 
 	React.useEffect(() => {
 		// @ts-ignore If you wish to show a stength meter, you need to load and set `zxcvbn` to a window variable by yourself.
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "ts-jest": "^24.3.0",
     "ts-loader": "^5.3.0",
     "ts-node": "^8.0.2",
-    "typescript": "^3.5.2",
+    "typescript": "^3.8.2",
     "webpack": "^4.41.5",
     "zxcvbn": "^4.4.2"
   },
--- a/src/extra/Markdown/specsuite.json
+++ b/src/extra/Markdown/specsuite.json
@@ -307,137 +307,137 @@
   {
     "name": "xss: 36",
     "source": "ABC<div style=\"x\\x3Aexpression(javascript:alert(1)\">DEF",
-    "expected": "<p>ABC</p><div>DEF</div><p></p>"
+    "expected": "<p>ABC</p><div>DEF<p></p>\n</div>"
   },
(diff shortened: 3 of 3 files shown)


src/unstable/components/Form/widgets/PasswordWidget.tsx
```diff
@@ -39,7 +44,8 @@ export interface PasswordStrengthProps {
 }
 
 const PasswordStrength = ({ password }: PasswordStrengthProps) => {
-	const [strengthScore, setStrengthScore] = React.useState();
```
sradevski: I don't think this change is necessary to fix the typings, let's keep it as it was.

src/unstable/components/Form/widgets/PasswordWidget.tsx
```diff
@@ -39,7 +44,8 @@ export interface PasswordStrengthProps {
...
+	const [strengthTitle, setStrengthTitle] = React.useState<string>();
+	const [strengthStyle, setStrengthStyle] = React.useState<StrengthStyle>();
```
sradevski: We don't need to keep things in state that can be derived (title can be derived from score, and so can styles).

src/unstable/components/Form/widgets/PasswordWidget.tsx
JSReds: the problem was in the typing of the strengthScore, since it starts with an undefined value and as explained in the link attached in the PR, undefined cannot be used as index type.
The other solution that we can apply is to set strengthScore to NaN as a starting value, but it seemed like a risky workaround.
Were you thinking of another way?

https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types

sradevski: @JSReds From what I can see it can also start as `0`

src/unstable/components/Form/widgets/PasswordWidget.tsx
```diff
@@ -48,25 +54,24 @@ const PasswordStrength = ({ password }: PasswordStrengthProps) => {
 		if (zxcvbn && password) {
 			try {
 				const { score } = zxcvbn(password);
-				setStrengthScore(score);
+				setStrengthTitle(STRENGTH_TITLES[score]);
```
LucianBuzzo: Is this refactor required due to the Typescript bump?

src/unstable/components/Form/widgets/PasswordWidget.tsx
JSReds: @sradevski totally agree but starting with 0 will skip the `|| strengthScore === undefined` check. Should we change the control or delete it and always render the component with 0 width ?

maybe I would opt to change the rule to `!strengthScore` and delete index 0 `{width: 0}`, from `STRENGTH_STYLE` as it will never be used. what do you think about it  ?

src/unstable/components/Form/widgets/PasswordWidget.tsx
JSReds: yes @LucianBuzzo, but @sradevski helped me find a much better solution

JSReds:
closing because github reopen it if it detect a push on the same branch just after it merged it.
I made a big mistake

PR #1072, opened 2020-02-28
aria-label on alert close button

iamsolankiamit:
Change-type: Patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/internal/DismissableContainer/index.tsx
+++ b/src/internal/DismissableContainer/index.tsx
@@ -54,6 +54,7 @@ export interface DismissableContainerProps extends FlexProps {
 	// This is passed so that we can vertically center the dismiss button with the first line of the content, regardless of the height of the content.
 	baselineHeight?: number;
 	solid?: boolean;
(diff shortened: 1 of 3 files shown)


src/internal/DismissableContainer/index.tsx
```diff
@@ -86,6 +86,8 @@ export const DismissableContainer = withTheme(
 					{children}
 					{onDismiss && (
 						<DismissButton
+							title="Dismiss Alert"
```
sradevski: This is not always an alert, it can be the onboarding steps, or something else. It's better if each component that uses the `DismissableContainer` pass their own title.

iamsolankiamit: Ahh, I missed this, I'll update this.

iamsolankiamit: I've added a default title of 'Dismiss' if nothing is passed.

PR #1073, opened 2020-03-02
Compact filters

JSReds:
Add compact prop to filters component

See: [Flowdock discussion](https://www.flowdock.com/app/rulemotion/resin-frontend/threads/ATGc-ie1nUzCHGSbmrx8HMYIQoH)
Change-type: patch
Depends-on: #1069
Signed-off-by: Andrea Rosci andrear@balena.io

this change has been made to allow a better view of the filters component on small screens resolutions. The `compact` property has been added and can be used as all `StyledSystemProps` in the following way:
```compact=[true, true, false, false]``` 
where each position in the array corresponds to a rendition breakpoint, thus giving the possibility to obtain different results on different resolutions

filters-compact

---
##### Contributor checklist

##### Reviewer Guidelines
---

### CODE REVIEW

File: src/components/Filters/ViewsMenu.tsx [hunk 1]
```diff
@@ -91,7 +92,16 @@ class ViewsMenu extends React.Component<ViewsMenuProps, ViewsMenuState> {
 	}
 
 	public render() {
-		const { views, renderMode, hasMultipleScopes } = this.props;
```
sradevski: This looks like it's an unrelated change, let's keep it as it was for this PR.

src/components/Filters/ViewsMenu.tsx
```diff
@@ -103,64 +113,68 @@ class ViewsMenu extends React.Component<ViewsMenuProps, ViewsMenuState> {
...
-					noListFormat
-					icon={<FontAwesomeIcon icon={faChartPie} />}
-					label="Views"
-					{...this.props.buttonProps}
-				>
-					<Box py={1}>
-						{!hasViews && (
-							<Box py={2} px={3}>
-								{"You haven't created any views yet"}
+				<UseBreakpointWrapper values={compact}>
```
sradevski: We don't need a wrapper component. Since `useBreakpoint` is a hook, we should use it as such.
We can either:
1. Refactor this component to make it a functional one, and use the hook as usual.
2. Extract the render body (or the one that makes sense) to another functional component, and use the hook there as usual.

I would prefer 1. as it is a very simple class component and should be pretty trivial to refactor it.

sradevski: Actually now that I look at the code, the hook shouldn't even live here, the parent should use the hook, and then just pass the result from the hook (something like `shouldShowLabel`), instead of the entire `compact` array.

src/components/Filters/index.tsx
```diff
@@ -362,23 +363,29 @@ class Filters extends React.Component<FiltersProps, FiltersState> {
...
-							primary
-							icon={<FontAwesomeIcon icon={faFilter} />}
-							onClick={() =>
-								this.setState({ showModal: true, editingFilter: null })
-							}
-							{...this.props.addFilterButtonProps}
-						>
-							Add filter
-						</Button>
+						<UseBreakpointWrapper values={hookValues}>
```
sradevski: See my comment above, let's do that for this component. I realize that this one is much bigger, so option 2. might be better. The fact that this component is pretty big means it needs to be refactored anyway, so lets do it in 2 commits, 1 that refactors it, and the second one that adds the compact prop

src/components/Filters/spec.js
```diff
@@ -192,6 +192,26 @@ describe('Filters component', () => {
           .text()
       ).toBe('Tag is not rarity : 10')
     })
+
+    it('should remove buttons labels if compact porperty is set to true', () => {
```
sradevski:
```suggestion
    it('should remove buttons labels if compact property is set to true', () => {
```

src/hooks/useBreakpoint.ts
```diff
@@ -22,3 +22,11 @@ export const useBreakpoint = <T extends any>(values: T[]): T => {
 
 	return newValues[currentBreakpoint];
 };
+
+export const UseBreakpointWrapper = ({
```
sradevski: Let's remove this, as it will just make things confusing.

JSReds:
> I have a few general comments here:
> 
> 1. When adding new functionality, please provide an explanation for what is being done and why, or link to an issue that contains this information
> 2. If you are changing the visual output of a component, please provide a screenshot (or even better a GIF) of the change in the PR description


sorry @LucianBuzzo, I was sure I linked the conversation on flowdock. I added a small description, the link to the discussion on flowdock and a GIF

LucianBuzzo:
@JSReds thanks a bunch! With the description and GIF you added the intention behind the pull request is much clearer :+1:

src/components/Filters/spec.js
```diff
@@ -192,6 +192,26 @@ describe('Filters component', () => {
           .text()
       ).toBe('Tag is not rarity : 10')
     })
+
+    it('should remove buttons labels if compact property is set to true', () => {
+      const component = mount(
+        <Provider>
+          <Filters compact={[true, true, true, true]} schema={schema} />
```
sradevski: NIT: We could do `true` here (I am thinking we should modify `useBreakpoint` to also accept a single value, and just use `Array.isArray` to check if it is array or not, and if it not, just convert it to one). Might be unnecessary for now though :)

src/components/Filters/ViewsMenu.tsx
```diff
@@ -176,6 +181,7 @@ export interface ViewsMenuProps {
 	setFilters: (filters: JSONSchema6[]) => void;
 	deleteView: (view: FiltersView) => any;
 	renderMode?: FilterRenderMode | FilterRenderMode[];
+	shouldHideLabel: boolean;
```
sradevski: Normally we use positive names, such as `shouldShowLabel`, but it's fine like this as I guess it matches the values from the `compact` array better.

src/components/Filters/index.tsx
```diff
@@ -1,63 +1,26 @@
...
 import castArray from 'lodash/castArray';
 import cloneDeep from 'lodash/cloneDeep';
 import findIndex from 'lodash/findIndex';
 import includes from 'lodash/includes';
-import isEmpty from 'lodash/isEmpty';
-import isEqual from 'lodash/isEqual';
-import map from 'lodash/map';
 import reject from 'lodash/reject';
-import * as React from 'react';
+import React, { useState } from 'react';
```
sradevski: I think this might cause typing issues, so let's just keep it as `* as React` and use `React.useState`

src/components/Filters/FilterModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+		setNewEdit(currentEdit);
+	};
+	return (
+		<Modal
+			title="Filter by"
+			cancel={onClose}
+			done={() => addFilter(newEdit)}
+			action="Save"
+		>
+			{map(newEdit, ({ field, operator, value }, index) => {
```
sradevski: Let's switch to the native `map` here.

src/components/Filters/index.tsx
```diff
@@ -66,482 +29,281 @@ const FilterWrapper = styled(Box)`
...
+		schema: flatSchema,
+		edit: [],
+	});
+
+	React.useEffect(() => {
+		filtersState.edit.push(SchemaSieve.getCleanEditModel(filtersState.schema));
+	}, []);
+
+	React.useEffect(() => {
+		setFiltersState({
```
sradevski: In order not to add `filtersState` as a dependency and recalculate this any time the state changes, we need to do `setState(filtersState => ...)` instead, otherwise we won't be using the altest filters state.

src/components/Filters/index.tsx
```diff
@@ -66,482 +29,281 @@ const FilterWrapper = styled(Box)`
...
+	}, [schema]);
 
-	public setEditOperator(operator: string, index: number) {
-		const currentEdit = this.state.edit.slice();
-		const item = currentEdit[index];
-		currentEdit.splice(index, 1, { ...item, operator });
-		this.setState({
-			edit: currentEdit,
+	React.useEffect(() => {
+		setFiltersState({
```
sradevski: Same here as above.

src/components/Filters/index.tsx
```diff
@@ -66,482 +29,281 @@ const FilterWrapper = styled(Box)`
...
+	}, [filters.length]);
 
-	public setEditValue(value: string, index: number) {
-		const currentEdit = this.state.edit.slice();
-		const item = currentEdit[index];
-		currentEdit.splice(index, 1, { ...item, value });
-		this.setState({
-			edit: currentEdit,
+	React.useEffect(() => {
+		setFiltersState({
```
sradevski: And here :)

src/components/Filters/index.tsx
```diff
@@ -66,482 +29,281 @@ const FilterWrapper = styled(Box)`
...
-		currentEdit.splice(index, 1, { ...item, operator });
-		this.setState({
-			edit: currentEdit,
+	React.useEffect(() => {
+		setFiltersState({
+			...filtersState,
+			filters: filters.map(filter => SchemaSieve.flattenSchema(filter)),
 		});
-	}
+	}, [filters.length]);
```
sradevski: I think we can edit the filters, so the length will stay the same, but the filters will change, so we need the whole array as a dependency

src/components/Filters/index.tsx
```diff
@@ -66,482 +29,281 @@ const FilterWrapper = styled(Box)`
...
-		currentEdit.splice(index, 1, { ...item, value });
-		this.setState({
-			edit: currentEdit,
+	React.useEffect(() => {
+		setFiltersState({
+			...filtersState,
+			views: flattenViews(views),
 		});
-	}
+	}, [views.length]);
```
sradevski: Same here

src/components/Filters/index.tsx
```diff
@@ -66,482 +29,281 @@ const FilterWrapper = styled(Box)`
...
 			return;
 		}
 
-		this.props.onFiltersUpdate(
-			this.state.filters.map(filter => SchemaSieve.unflattenSchema(filter)),
+		onFiltersUpdate(
+			filtersState.filters.map(filter => SchemaSieve.unflattenSchema(filter)),
 		);
-	}
+	}, [filtersState.filters.length]);
```
sradevski: Same here

src/components/Filters/index.tsx
```diff
@@ -66,482 +29,281 @@ const FilterWrapper = styled(Box)`
...
+
+	React.useEffect(() => onViewsUpdate && onViewsUpdate(filtersState.views), [
+		filtersState.views.length,
```
sradevski: Same here

src/components/Filters/spec.js
JSReds: I also thought about it and I think it is absolutely to do, so we make the same behaviour of the `StyledSystemProps`. But I thought of doing it in another PR, what do you think? @sradevski

src/components/Filters/ViewsMenu.tsx
JSReds: absolutely agree, usually I use positive names too. In this case I used it this way to match the `compact` prop in a better way

src/components/Filters/spec.js
sradevski: @JSReds yeah for sure, since it is an extension of the interface, users won't need to do any changes, so we can do it in a separate PR.

src/components/Filters/index.tsx
```diff
@@ -66,482 +30,293 @@ const FilterWrapper = styled(Box)`
...
-		this.setState({
-			edit: currentEdit,
-		});
-	}
+	React.useEffect(() => {
+		setFiltersState(prevState => ({
+			...prevState,
+			filters: filters.map(filter => SchemaSieve.flattenSchema(filter)),
+		}));
+	}, [useDeepCompare(filters)]);
```
sradevski: Now that I look at the code, why do we even need this effect? We already have all the CRUD methods for filters and views that set the state, so these hooks feel unnecessary?

src/components/Filters/index.tsx
```diff
@@ -66,482 +30,293 @@ const FilterWrapper = styled(Box)`
...
 
-		let filters: JSONSchema6[];
+		let currentFilters: JSONSchema6[];
 
 		if (!!$id) {
-			const matchIndex = findIndex(currentFilters, { $id });
-			currentFilters.splice(matchIndex, 1, filter);
-			filters = currentFilters.slice();
+			const matchIndex = findIndex(filtersState.filters, { $id });
+			filtersState.filters.splice(matchIndex, 1, filter);
```
sradevski: This mutates the filters, which means the component won't rerender. Let's use immutable functions instead.

src/components/Filters/index.tsx
```diff
@@ -66,482 +30,293 @@ const FilterWrapper = styled(Box)`
...
+	};
 
-	public saveView(name: string, scope: string | null) {
+	const saveView = (name: string, scope: string | null) => {
 		const view: FiltersView = {
 			id: randomString(),
 			name,
 			scope,
-			filters: cloneDeep(this.state.filters),
+			filters: cloneDeep(filtersState.filters),
```
sradevski: Why do we clone the filters here? Ideally, we should never do deep clones, and deep comparisons, and just recreate objects as deep as necessary when they change.

JSReds:
> It's great that the Filters/index.tsx file got smaller and easier to reason about. I still find it hard to reason about, however, and it feels like there are many hacks here and there just to make things work. Let's clean it up and start thinking from first principles - think how React works, ignore how it was implemented before :)
> 
> Another thing, maybe we want to override the dropdown button icon to the `View` one, otherwise of we have a set of actions dropdown next to it on mobile, it will look like two equal dropdowns are enxt to each other. We can do this as a separate commit though, once all of the other issues are resolved :)

I agree @sradevski, I wanted to avoid too many changes in this PR and make it easier to read the changes. I thought to do a refactor in another PR. But at this point I did it in this one :)

src/components/DropDownButton/index.tsx
```diff
@@ -108,11 +108,13 @@ const Toggle = ({
...
 				icon={
-					open ? (
-						<FontAwesomeIcon icon={faChevronUp} />
-					) : (
-						<FontAwesomeIcon icon={faChevronDown} />
-					)
+					<Flex>
+						{props.icon}
+						<FontAwesomeIcon
+							style={{ marginLeft: props.icon ? '4px' : undefined }}
```
JSReds: @sradevski the alternative to this inline style could be to put an empty string between the two icons and update jest tests.

src/components/DropDownButton/index.tsx
```diff
@@ -108,11 +108,13 @@ const Toggle = ({
 			<JoinedButton
 				{...props}
 				icon={
-					open ? (
-						<FontAwesomeIcon icon={faChevronUp} />
-					) : (
-						<FontAwesomeIcon icon={faChevronDown} />
-					)
+					<Flex>
```
sradevski: @JSReds Let's not modify the DropDownButton as it is unnecessary. Instead, in the filters we switch between `<icon> Views...` and `<icon>` as labels Alternatively, we can check `if (label || icon)` in the DropDown instead of checking only for the `label`.

src/components/DropDownButton/index.tsx
```diff
@@ -199,7 +201,7 @@ class DropDownButton extends React.Component<
 			className,
 			...props
 		} = this.props;
-
+		console.log(this.props);
```
sradevski: Let's remove this :)

src/components/DropDownButton/index.tsx
JSReds: @sradevski that was my first try but in both cases the result will be something like this:

Screenshot 2020-03-26 at 14 48 15

should we leave it like that and play with `viewsMenuButtonProps: {width: n}` from the Filters component?

src/components/DropDownButton/index.tsx
JSReds: I would say so :cold_sweat:

src/components/DropDownButton/index.tsx
sradevski: @JSReds that's fine, the only issue here is the large padding AFAIK, but that is themable and in the dashboard is lower than in rendition.

src/components/Notifications/defaultStyle.ts
```diff
@@ -7,41 +7,22 @@ export default css`
 	.notification-container-bottom-center,
 	.notification-container-bottom-left,
 	.notification-container-bottom-right,
-	.notification-container-center,
```
sradevski: Make sure you install the latest npm modules, as this might be happening because you have older version (or maybe it is a newer version, not sure, just try and see :))

src/components/Filters/story.js
```diff
@@ -241,6 +241,16 @@ storiesOf('Next/Filters', module)
       </Box>
     )
   })
+  .add('Compact', () => {
+    const props = {
```
sradevski: Why isn't this inline with the component?

JSReds: changed to inline as recommended

src/components/Notifications/defaultStyle.ts
JSReds: I actually had an old version, thanks for the info :)

PR #1084, opened 2020-03-26
Update the spinner typings to accept a custom label component

sradevski:
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


README.md
```diff
@@ -676,7 +676,7 @@ You can refer to the [grommet select page](https://v2.grommet.io/select) for all
 | `children`     | `React.ReactNode` |         | -          | If the spinner has children, it will show an opaque overlay over the children when it is spinning |
 | `show`         | `boolean`         | true    | -          | If passed, it will control whether the spinner is shown or not |
 | `emphasized`   | `boolean`         | false   | -          | If true, it will render a large spinner |
-| `label`        | `string`          |         | -          | A label that will be rendered next to the spinner. Renders on right-hand side for standard spinner, and below spinner for emphasized |
+| `label`        | `string` | `React.ReactNode`|            | -          | A label that will be rendered next to the spinner. Renders on right-hand side for standard spinner, and below spinner for emphasized |
```
iamsolankiamit: the `|` between `string` and `React.ReactNode` is messing with the table layout. we can either use `&#124;` or `\|`

sradevski: It was my bad, they all had to be as a code snippet

iamsolankiamit: wrapping it with `` ` `` (backtick) won't fix it

for ex. The table  has `a|b` and it doesn't works

| a |  ab     | c |
|---|---------|---|
| a |  `a|b`  | c |

and the table below has `a\|b` and it does works

| a |   ab   | c |
|---|--------|---|
| a |  a\|b  | c |

sradevski:
@balena-ci retest

README.md
sradevski: @iamsolankiamit In other places we also use the code snippet pattern and it renders as expected in VSCode for example. I agree that what you suggest is better though, I'll update now.

PR #1087, opened 2020-03-30
Bump grommet to 2.11.3

sradevski:
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Checkbox/index.tsx
+++ b/src/components/Checkbox/index.tsx
@@ -21,9 +21,15 @@ export const getHoverStyle = (props: any) => `
 	& label:hover input:not([disabled]) + div {
 		border-color: ${props.theme.colors.tertiary.main}!important;
 	}
-	& label:hover input[checked]:not([disabled]) + div {
-		border-color: ${props.theme.colors.primary.dark}!important;
-		background: ${props.theme.colors.primary.dark};
+	${
+		props.isChecked
+			? `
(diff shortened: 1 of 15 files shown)


src/components/Checkbox/index.tsx
```diff
@@ -21,9 +21,15 @@ export const getHoverStyle = (props: any) => `
 	& label:hover input:not([disabled]) + div {
 		border-color: ${props.theme.colors.tertiary.main}!important;
 	}
-	& label:hover input[checked]:not([disabled]) + div {
-		border-color: ${props.theme.colors.primary.dark}!important;
-		background: ${props.theme.colors.primary.dark};
+	${
```
sradevski: Grommet stopped applying the `checked` attribute on both radio buttons and checkboxes, so we pass the checked prop instead. This wouldn't work for uncontrolled components, but we always have the components as controlled so it doesn't matter.

src/components/Checkbox/spec.js.snap
```diff
@@ -178,6 +178,7 @@ exports[`Checkbox renders correctly 1`] = `
   className="c0 c1"
 >
   <div
+    checked={true}
```
iamsolankiamit: let's rename the prop to isChecked so that it doesn't gets forwarded to dom

PR #1088, opened 2020-03-30
Make spinner use tertiary color

sradevski:
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


iamsolankiamit:
I think we should have some screenshot testing for these.

sradevski:
@iamsolankiamit updated

iamsolankiamit:
the emphasised with label looks like it is off center, is that on purpose?

sradevski:
@iamsolankiamit The reason that was centered was the spinner div took the entire width before, which was unintended as in many cases we want the spinner to be left-aligned (as discussed before). If you wish to center the spinner in a page, then you either pass the page as a child, or you center it manually.

PR #1089, opened 2020-03-31
Fix: filters search bar

JSReds:
Add useEffect right dependecy to update search bar results

See: #1090 
See: [flowdock link](https://www.flowdock.com/app/rulemotion/resin-frontend/threads/RQD66JxIGeGR0AioZO9V5QHNElJ)
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Filters/index.tsx
```diff
@@ -85,14 +85,20 @@ const Filters = ({
...
+	React.useEffect(() => {
+		const newFilters = searchString
+			? SchemaSieve.upsertFullTextSearch(
+					schemaState,
+					filtersState,
+					searchString,
+			  )
+			: SchemaSieve.removeFullTextSearch(filtersState);
+
+		setFiltersState(newFilters);
```
sradevski, line 99: If this changes the filtersState, then the effect above should fire. Like this the parent callback will be called twice, which is not really desired.

src/components/Filters/index.tsx
```diff
@@ -116,6 +122,12 @@ const Filters = ({
 		}
 	}, [schema]);
 
+	const updateParentFilters = (updatedFilters: JSONSchema6[]) =>
```
sradevski, line 125: Let's move this above the effects that use it.

src/components/Filters/index.tsx
sradevski: I just tried and it doesn't fire, but I don't get it why. This whole component needs some refactoring now that it became even more confusing. Can you please look into simplifying this once the bug is fixed? Thanks

PR #1092, opened 2020-04-01
revert: filters refactor

JSReds:
revert filters refactor due to a regression on search bar

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Filters/index.tsx
```diff
@@ -28,229 +29,370 @@ const FilterWrapper = styled(Box)`
...
+			value,
+		};
+	}
+
+	public getOperators(field: string) {
+		const schema = this.state.schema;
+		return SchemaSieve.getOperators(schema, field);
+	}
+
+	public setEditField(field: string, index: number) {
```
thgreasi, line 146: I think that's not used anywhere.

src/components/Filters/index.tsx
```diff
@@ -28,229 +29,370 @@ const FilterWrapper = styled(Box)`
...
+		const currentEdit = this.state.edit.slice();
+		currentEdit.splice(index, 1, this.getCleanEditModel(field));
+		this.setState({
+			edit: currentEdit,
+		});
+	}
+
+	public setEditOperator(operator: string, index: number) {
```
thgreasi, line 154: I think that's not used anywhere.

src/components/Filters/index.tsx
```diff
@@ -28,229 +29,370 @@ const FilterWrapper = styled(Box)`
...
+		const currentEdit = this.state.edit.slice();
+		const item = currentEdit[index];
+		currentEdit.splice(index, 1, { ...item, operator });
+		this.setState({
+			edit: currentEdit,
+		});
+	}
+
+	public setEditValue(value: string, index: number) {
```
thgreasi, line 163: I think that's not used anywhere.

thgreasi:
Can we also add the `compact` prop on a separate commit as per:
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/RQD66JxIGeGR0AioZO9V5QHNElJ

src/components/Button/index.tsx
```diff
@@ -223,9 +224,14 @@ const Base = (props: ThemedButtonProps) => {
 		plain,
 		active,
 		light,
+		compact,
 		...restProps
 	} = props;
 
+	// Add a method to hide the button label. Here we can just pass an array of 
+	// boolean to decide on each breakpoint whether to show the label or not
+	const shouldCompact = useBreakpoint(compact || [false]);
```
sradevski, line 233:
```suggestion
	const shouldShowLabel = useBreakpoint(compact || [false]);
```

src/components/DropDownButton/index.tsx
```diff
@@ -84,17 +85,18 @@ const Toggle = ({
 	handler,
 	label,
 	joined,
+	compact,
 	...props
 }: InternalDropDownButtonProps & {
 	open: boolean;
 	handler: React.MouseEventHandler<HTMLElement>;
 }) => {
+	const shouldCompact = useBreakpoint(compact || [false]);
```
sradevski, line 94:
```suggestion
	const shouldShowLabel = useBreakpoint(compact || [false]);
```

src/components/DropDownButton/index.tsx
```diff
@@ -84,17 +85,18 @@ const Toggle = ({
...
 	if (joined) {
-		if (label || props.icon) {
+		if (shouldCompact) {
```
sradevski, line 96: Make this `label || shouldShowLabel`, and then have `{shouldShowLabel && <Box .../>}` instead of having two separate conditions with almost the same code.

src/components/Filters/ViewsMenu.tsx
```diff
@@ -181,7 +177,7 @@ export interface ViewsMenuProps {
 	setFilters: (filters: JSONSchema6[]) => void;
 	deleteView: (view: FiltersView) => any;
 	renderMode?: FilterRenderMode | FilterRenderMode[];
-	shouldHideLabel: boolean;
+	compact: boolean[];
```
sradevski, line 180: Make this optional

src/components/Filters/index.tsx
```diff
@@ -28,229 +29,344 @@ const FilterWrapper = styled(Box)`
...
+							disabled={this.props.disabled}
+							views={this.state.views || []}
+							schema={this.props.schema}
+							hasMultipleScopes={
+								this.props.viewScopes && this.props.viewScopes.length > 1
 							}
+							setFilters={filters => this.setFilters(filters)}
+							deleteView={view => this.deleteView(view)}
+							renderMode={this.props.renderMode}
+							compact={this.props.compact || [false]}
```
sradevski, line 325:
```suggestion
							compact={this.props.compact}
```
Since every sub-component should initialize its own default values.

src/components/Filters/index.tsx
```diff
@@ -28,229 +29,344 @@ const FilterWrapper = styled(Box)`
...
+						<Button
+							mr={30}
+							disabled={this.props.disabled}
+							primary
+							icon={<FontAwesomeIcon icon={faFilter} />}
+							onClick={() =>
+								this.setState({ showModal: true, editingFilter: null })
+							}
+							label="Add filter"
+							compact={this.props.compact || [false]}
```
sradevski, line 294:
```suggestion
							compact={this.props.compact}
```
Same as below.

src/components/Button/index.tsx
JSReds: @sradevski in doing so we would have a contradiction in the result since `compact true` will remove the label and we would have `shouldShowLabel true` which should also remove the label.

sradevski: @JSReds that's right, I was thinking of `shouldHideLabel`, but I am fine with it as it is as well, not a big deal.

src/components/DropDownButton/index.tsx
```diff
@@ -84,17 +85,19 @@ const Toggle = ({
...
 	open: boolean;
 	handler: React.MouseEventHandler<HTMLElement>;
 }) => {
+	const shouldCompact = useBreakpoint(compact || [false]);
 	if (joined) {
-		if (label || props.icon) {
+		if (label || shouldCompact) {
 			return (
 				<JoinedButton {...props} onClick={handler}>
+					{!shouldCompact && <Box mr={2}>{label}</Box>}
```
thgreasi, line 99: Why did we moved this outside the `<Flex>`?
I feel that if we keep that in there there will be zero visual diff.
If we really want to move it, then let's do it in a different isolated PR.

src/components/DropDownButton/index.tsx
```diff
@@ -84,17 +85,19 @@ const Toggle = ({
...
+	compact,
 	...props
 }: InternalDropDownButtonProps & {
 	open: boolean;
 	handler: React.MouseEventHandler<HTMLElement>;
 }) => {
+	const shouldCompact = useBreakpoint(compact || [false]);
 	if (joined) {
-		if (label || props.icon) {
+		if (label || shouldCompact) {
```
thgreasi, line 96: I'm not sure I get this change :thinking:
I would expect this to still be `label || props.icon`.
Can I get some context?

JSReds: @thgreasi this change had been made when the compact property was added to the filters. Now since the compact functionality is managed directly by the buttons we can remove that check as it was originally and manage it with the compact property

Issue #1077, opened 2020-03-13
Add support for 'active step' in Steps component


grahammcculloch:
**Is your feature request related to a problem? Please describe.**
See [Jellyfish issue 3318](https://github.com/product-os/jellyfish/issues/3318). 

**Describe the solution you'd like**
![image](https://user-images.githubusercontent.com/2925657/76590815-f448ec00-6520-11ea-8266-d05cc0542817.png)
In the above screenshot its really not clear that the user is on Step 1. It would be nice if we could pass in an `activeStepIndex` property to Steps, that would highlight the active step in a prominent way.

**Describe alternatives you've considered**
I considered adding my own styling but I can't pass in my own class name to the `Step` components and because Rendition uses styled-components, I can't target any of the existing classes for styling.

grahammcculloch:
Here's my work in progress (previous/next buttons are just for the benefit of the storybook story) ....
![Active Steps](https://user-images.githubusercontent.com/2925657/76606503-ef962f00-6544-11ea-8129-82da5d08b06c.gif)

@sradevski thoughts?

sradevski:
@grahammcculloch This might be a design question more than anything else. I would guess it makes sense only the first incomplete step to be highlighted, is that reasonable assumption? I think we can do the following:
1. Allow numbering of steps (a number in the circle).
2. When a step is done, put a green checkmark icon in the place of the number
3. When the step is to be done next, fill the number circle with some color (blue, for example)
4. When the step is neither done, nor next in line, only show an outlined circle with the number inside using a grayish opaque color

What do you think? I'd suggest you ping Alexandra in flowdock and discuss about which approach would work best.

grahammcculloch:
Fixed with #1094

PR #1094, opened 2020-04-02
feat: Add support for ordered "Steps"

grahammcculloch:
Connects-to: #1077
Change-type: minor

---

The `Steps` component has been updated to support the concept of having **ordered** steps. This mode of operation is designed to be used when the user must complete a sequence of steps _in order_ and be aware of which step they are currently on.

Ordered Steps

The existing `Steps` functionality is unchanged. The only minor change is making the step link/text colors more consistent.

##### Contributor checklist

##### Reviewer Guidelines
---


grahammcculloch:
Not sure why there is a failing `List` snapshot test - I haven't touched any code that affects that as far as I know. cc @sradevski

src/components/Steps/index.tsx
```diff
@@ -99,9 +228,30 @@ const FramelessSteps = ({ children, ...otherProps }: any) => {
...
+		}: InternalStepsProps,
 		ref,
 	) => {
+		if (ordered) {
+			if (typeof activeStepIndex !== 'number') {
+				throw new Error(
+					'You must specify the activeStepIndex for ordered Steps',
+				);
+			} else {
+				const steps = React.Children.toArray(children);
```
sradevski, line 249: There is also `React.Children.count(children)` which is performant.

src/components/Steps/index.tsx
```diff
@@ -1,72 +1,181 @@
...
 `;
 
-const HeaderContainer = styled(Flex)`
-	min-height: ${props => px(props.theme.fontSizes[5])};
+const StepTxt = styled(Txt)<{ active: boolean }>`
+	color: ${props => (props.active ? '' : props.theme.colors.tertiary.light)};
 `;
 
-export const Step = ({ status, children, onClick }: StepProps) => {
+const StepLink = styled(Link)`
```
sradevski, line 83: Why should this be overridden? IMO all links should look the same, and if we think the current one doesn't look as expected, we need to update the Link component itself.

src/components/Steps/index.tsx
```diff
@@ -1,72 +1,181 @@
...
-			)}
+			<StepIcon
+				ordered={Boolean(ordered)}
+				status={status}
+				active={Boolean(active)}
+				index={index}
+			/>
 			{onClick ? (
-				<Link ml={status === 'none' ? 0 : 2} onClick={onClick}>
+				<StepLink color="tertiary.light" onClick={onClick}>
```
sradevski, line 139: Same reasoning here, I don't think we should be overriding the color. If it is like this in the designs, we need to chat with Alexandra and update the Link component instead.

src/components/Steps/index.tsx
```diff
@@ -1,72 +1,181 @@
...
+const isStepClickable = (
+	ordered: boolean,
+	steps: any[],
+	index: number,
+	activeStepIndex: number,
+) => {
+	if (!ordered) {
+		return true;
+	}
+	const allPreviousStepsComplete = every(steps.slice(0, index), {
```
sradevski, line 164: We can use `findIndex` where status is not completed, and compare that with the component index instead, so we don't have to recreate an array twice (here and below) (maybe the lodash method, the native one doesn't have the best support yet).

grahammcculloch:
@sradevski thanks for your detailed review. I've applied all the changes you suggested in [this commit](https://github.com/balena-io-modules/rendition/pull/1094/commits/a940c7d448613fe3a106e4bf93227c51afcfdc96). 

Also note my [earlier comment](https://github.com/balena-io-modules/rendition/pull/1094#issuecomment-618188750) about the failing `List` snapsot test...

grahammcculloch:
@sradevski 
> I think you can remove node_modules, reinstall them and run the snapshot generation again, it's probably a dependency that got changed (since we don't have a lockfile) that caused the diff.

So I've tried the following:
```
git checkout master
rm -rf node_modules/
npm install
npm run jest -- src/components/List/spec.js
```

And I still get the error saying the snapshot has changed (same as in the failing CI step in this PR).

I've tried clearing the jest cache (`npm run jest -- --clearCache`) but that doesn't fix it.

Also on a clean master branch (as above) if I run `npm run build` I get the following error:
```
node_modules/htmlparser2/lib/WritableStream.d.ts:12:15 - error TS2749: 'StringDecoder' refers to a value, but is being used as a type here.

12     _decoder: StringDecoder;
````

FYI this error goes away if I update the `@types/node` package dependency to `^12.0.0`. But doesn't explain why its not working on my machine with a clean checkout of master.

I'm running node v12.16.1, npm v6.13.4 on Ubuntu (Windows WSL).

sradevski:
@grahammcculloch can you run `npm run generate-snapshots` and push? As I said, since we don't lock on specific versions, some dependency might have changed and you get that. Once you update the snapshots it should work as expected.

sradevski:
@balena-ci retest

sradevski:
@grahammcculloch I am not sure why but one screenshot fails, which really makes me rethink the decision to not have a lockfile. Anyhow for now try running `npm run generate-screenshots` and see if that solves the issue.

PR #1099, opened 2020-04-17
Apply text color from theme globally

sradevski:
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/extra/MarkdownEditor/defaultStyle.ts
```diff
@@ -411,7 +411,7 @@ export default css`
 		border-bottom-right-radius: 4px;
 		padding: 10px;
 		font: inherit;
-		z-index: 0;
+		z-index: 1;
```
iamsolankiamit, line 414: was the editor updated?

sradevski: @iamsolankiamit we don't have a package lock (since it's ignored by clients), so things can change, like in this case.

PR #1104, opened 2020-04-30
fix: internal nullable prop

JSReds:
set index as a nullable internal prop on Step component

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Steps/index.tsx
+++ b/src/components/Steps/index.tsx
@@ -105,7 +105,7 @@ const StepIcon = ({ ordered, status, active, index }: StepIconProps) => {
 	);
 };
 
-export const Step = ({
+const StepBase = ({
 	ordered,
 	active,
 	index,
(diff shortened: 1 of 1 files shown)


JSReds:
@sradevski from what I can see all the components currently work like this, in fact if we see for example the `Alert` component the `plaintext` property is shared with the user who uses it even if it is declared within the `InternalAlertProps` iterator. Indeed we should review them for a moment since as you say they should not be visible to the user, probably not in this PR.
As for the `Index` and `ordered` props, at the moment they are not really internal props since they are not calculated internally. Perhaps for the time being I would simply move it in the `StepProps`. what do you think ?

JSReds:
@sradevski yes it is, now the index prop is no longer accessible

JSReds:
@balena-ci rebase

PR #1106, opened 2020-05-07
Update to balena lint

Page-:
Depends on https://github.com/balena-io/ci-images/pull/51

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,8 +40,8 @@
     "jest:watch": "jest --maxWorkers=4 --watch",
     "eslint": "eslint src .storybook",
     "eslint-fix": "eslint src .storybook --fix",
-    "tslint": "resin-lint --typescript src",
-    "tslint-fix": "resin-lint --typescript --fix src",
+    "tslint": "balena-lint --typescript src",
+    "tslint-fix": "balena-lint --typescript --fix src",
     "lint": "npm run eslint && npm run tslint",
     "precommit": "lint-staged",
     "prepack": "npm run build",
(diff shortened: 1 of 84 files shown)


package.json
```diff
@@ -88,12 +89,11 @@
     "jsdom": "^11.6.2",
     "lint-staged": "^4.1.1",
     "match-media-mock": "^0.1.1",
-    "prettier": "^1.15.3",
+    "prettier": "^2.0.5",
```
sradevski, line 92: Doesn't balena/lint include prettier as well?

Page-: It does which is why I updated this to match, the problem is this repo uses eslint for js files with quite a specialized config, so whilst I could switch to balena-lint for the js I don't know enough about the special stuff this repo uses to judge if it's safe and so we need to keep prettier to do `prettier && eslint`

sradevski: @Page- we should define react-specific linting rules in @balena/lint and remove everything custom here. I have an issue in the dashboard for it already, but haven't found the time to do it, so for now let's just merge this as it is.

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

PR #1107, opened 2020-05-07
Lodash to native

Page-:
This removes some unnecessary lodash usage which should reduce the bundle size and in some cases improve typings/performance


Page-:
@balena-ci retest

PR #1108, opened 2020-05-08
Add custom renderer for marked

StefKors:
**DRAFT**

This PR adds the possibility of using a custom renderer for marked. With this, we can for example modify the URL behavior.

Change-type: patch
Signed-off-by: Stef Kors <stef@balena.io>

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/extra/Markdown/index.tsx
+++ b/src/extra/Markdown/index.tsx
@@ -9,6 +9,7 @@ export { defaultSanitizerOptions } from '../utils';
 type MarkdownProps = TxtProps & {
 	children: string;
 	sanitizerOptions?: sanitizeHtml.IOptions;
+	renderer?: () => void;
 };
 
 /*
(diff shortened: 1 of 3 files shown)


grahammcculloch:
@StefKors I've added a commit that tidies up a few things - hope it makes sense. Feel free to push back on the change I made that removes the `marked` argument.

Still to do:
1. Add a unit test
1. (Re)generate screenshots using `npm run generate-screenshots`
1. Update documentation in `src/extra/Markdown/README.md` and re-generate documentation using `npm run build:docs`

StefKors:
@grahammcculloch does this mean that when a user wants to use a custom renderer that they need to install `Marked` themselves?

I would love to remove the marked argument, but I think it's bad practice to have users manually install a package when it's already included in Rendition. This could also give some difficult version issues.

Is there another format/structure we can use here? I would prefer not to add extra arguments.

---
The other changes look good! thanks for the help with the typings, it explains a lot 👍

StefKors:
Moving Jellyfish logic to Rendition is favor of exposing the `marked` renderer.

PR #1110, opened 2020-05-13
Add header height in Theme for Modal component

thundron:
This gives us more freedom on modals top placement

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


thundron:
@balena-ci retest

thundron:
> The whole point about not exposing this was to limit the flexibility so we can't easily mess it up. If we decide that it's absolutely necessary, I'd say we only expose `center` and `top` and nothing else. We can also take a top margin (which is maybe already supported?) to account for navbars, but that should be it.

As far as I can see on Grommet's documentation, we can't freely modify the `margin` outside of fixed values (xxsmall, xsmall, small, etc.) while specifying the position is actually what we need e.g. for Etcher - as you can see [here](https://app.zeplin.io/project/5ae18c241fbe2df5bf06e671/screen/5dd28247a07ca959917c5981) the modal we use *starts from the bottom* and that way we can also easily expand it to fit just below the header, whereas with a margin it would be moved from the *center* which, given the above structure, doesn't really make sense to me.
I'll experiment a bit more with the margin to see if we can have our own property along Grommet's, but sounds a bit like a workaround

sradevski:
@thundron I don't think the Modal is bottom-aligned, it just happens that its height goes from under the navbar to the bottom. I was suggesting that it's fine to expose the position argument, but only with `top` and `center`, and top can even have a predefined top margin to account for headers (normally 64px, although we can expose it as a prop). Let me know what you think :)

thundron:
@sradevski Waiting for NPM to work again to push the changes (I need the `balena-lint` module as it's not installed with `npm i` in the project)
I added a `marginTop` property applied then to the Layer's `style` prop which shouldn't affect the default `margin="small"` we use from Grommet

sradevski:
@thundron we can even make it a fixed value and not expose the prop, and only expose it if we see the default value doesn't work well for all use-cases. It's easier to add props, rather than to remove them, so it's better to start by being conservative by default.

src/components/Modal/index.tsx
```diff
@@ -125,6 +125,9 @@ class Modal extends React.Component<ThemedModalProps, any> {
 				responsive={false}
 				margin="small"
 				animate={false}
+				style={{
+					marginTop: theme.header.height,
+				}}
```
JSReds, line 130: I think you can also use the prop margin without style and get the same result: 
``` margin={{
    "horizontal": "small",
    "top": theme.header.height,
    "bottom": "small",
  }}```

Comment (thundron) [reply src/components/Modal/index.tsx:130]: You can only use the breakpoints you define using Grommet's API (e.g. re-assign [these values](https://github.com/grommet/grommet/blob/1ee98897864208d6b3ec0462cc9bb5f8c994484a/src/js/utils/index.d.ts#L41)) in the `margin` property

Comment (JSReds) [reply src/components/Modal/index.tsx:130]: From the file you shared I can see that it should actually accept both 
`margin?: MarginType;`
`export type MarginType = EdgeType;`
´´´
type EdgeType = "none" | EdgeSizeType | {bottom?: EdgeSizeType | string,end?: EdgeSizeType | string,horizontal?: EdgeSizeType | string,left?: EdgeSizeType | string,right?: EdgeSizeType | string,start?: EdgeSizeType | string,top?: EdgeSizeType | string,vertical?: EdgeSizeType | string} | string
```
where `top: EdgeSizeType | string`should accept `theme.header.height` which should be a string

thundron: It should, seeing [this issue](https://github.com/grommet/grommet/issues/2456), but it doesn't for some reason

thundron: or rather the value is accepted but totally ignored

JSReds: It is overwritten by the default value of the `position` prop, which is `center` and puts the modal in an absolute position with `top: 50%`, a value that is overwriting the setted margin: {top: ...}.
The `margin` prop in this case is just changing the `top` `bottom` `left` `right` value of the modal position and not the css margin value. To make it work you should set `position="top"` as a Layer prop or otherwise remove the `margin` prop since in this case it is not doing much.

src/components/Modal/story.js
```diff
@@ -349,3 +350,26 @@ storiesOf('Next/Modal', module)
   .add('Nested Modals', () => {
     return <NestedModalDemo />
   })
+  .add('With Top Margin', () => {
+    return (
+      <Modal
```
sradevski, line 355: Let's wrap this in a provider and just pass `theme={{height: 2em}}` without importing the whole theme.

src/components/Modal/index.tsx
```diff
@@ -125,6 +125,9 @@ class Modal extends React.Component<ThemedModalProps, any> {
 				responsive={false}
 				margin="small"
 				animate={false}
+				style={{
+					marginTop: theme.header.height,
```
sradevski, line 129: Let's apply it only when the position is `top`, otherwise it's fine to use style, I don't mind it.

src/components/Modal/index.tsx
```diff
@@ -175,6 +179,7 @@ class Modal extends React.Component<ThemedModalProps, any> {
 export interface ModalProps extends DefaultProps {
 	title?: string;
 	width?: ResponsiveStyle;
+	position?: 'center' | 'top';
```
sradevski, line 182: Let's add it to the README as well :)

thundron: I thought we had automated docs, added.
Is there a place where theme properties are documented so I can add the header.height too?

PR #1111, opened 2020-05-14
Basic Header implementation

iamsolankiamit:
Change-type: minor

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Header/NavDropDown.tsx
+++ b/src/components/Header/NavDropDown.tsx
@@ -1,0 +1,133 @@
+import React, { MouseEvent } from 'react';
+import styled from 'styled-components';
+import { Box, BoxProps } from '../Box';
+import { Divider } from '../Divider';
+import { Link, LinkProps } from '../Link';
+import { Txt } from '../Txt';
+
+import { faChevronDown } from '@fortawesome/free-solid-svg-icons/faChevronDown';
+import { faChevronUp } from '@fortawesome/free-solid-svg-icons/faChevronUp';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import asRendition from '../../asRendition';
+import { invertBreakpoint } from '../../utils';
+
+const Panel = styled(Box)`
+	min-width: 169px;
+	border-radius: 2px;
+	box-shadow: -10px 9px 21px 0 rgba(152, 173, 227, 0.08);
+	border: solid 1px #e8ebf2;
+	background-color: #ffffff;
+	margin-top: 12px;
+`;
+
+const PanelWrapper = styled(Box)`
+	transform: translateX(-20px);
+	position: absolute;
+	z-index: 2;
+
+	@media (max-width: ${(props) =>
+			invertBreakpoint(props.theme.breakpoints[1])}px) {
+		transform: none;
+		position: relative;
+
+		> div {
+			margin: 0;
+			width: 100%;
+			border: none;
+			box-shadow: none;
+			padding: 0;
+		}
+	}
+`;
+
+const Item = styled(Box)`
+	&:hover {
+		background: white;
+	}
+
+	> div {
+		margin: 0;
+	}
+
+	a {
+		width: 100%;
+	}
+
+	@media (max-width: ${(props) =>
+			invertBreakpoint(props.theme.breakpoints[1])}px) {
+		padding: 0;
+
+		a {
+			padding-left: 16px;
+		}
+	}
+`;
+
+interface NavDropDownProps extends LinkProps {
+	title: string;
+	showDividers?: boolean;
+}
+
+export const NavDropDown = (props: NavDropDownProps) => {
+	const [dropDownOpen, setDropDownOpen] = React.useState(false);
+
+	const showDropdown = (e: MouseEvent) => {
+		e.preventDefault();
+		setDropDownOpen(true);
+	};
+
+	const hideDropdown = (e: MouseEvent) => {
+		e.preventDefault();
+		setDropDownOpen(false);
+	};
+
+	const toggle = (e: MouseEvent) => {
+		e.preventDefault();
+		setDropDownOpen((open) => !open);
+	};
+
+	return (
+		<Box onMouseLeave={hideDropdown}>
+			<Link
+				fontSize={0}
+				href="#"
+				onClick={toggle}
+				onMouseOver={showDropdown}
+				py={2}
+				color="text.main"
+				{...props}
+			>
+				<Txt.span bold mr={3} style={{ textDecoration: 'underline' }}>
+					{props.title}
+				</Txt.span>
+				<FontAwesomeIcon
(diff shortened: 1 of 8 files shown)


thundron:
Should we maybe remove the [Navbar](https://balena-io-modules.github.io/rendition/?path=/story/core-navbar--standard) component to be replaced with this? Afaik that's broken so I don't think we're using it anywhere. cc @sradevski

src/components/Header/NavDropDown.tsx
```diff
@@ -0,0 +1,134 @@
+import React, { MouseEvent } from 'react';
```
sradevski, line 1: Let's reuse our Dropdown component instead (and we might need to improve it visually while doing that)

src/components/Header/README.md
```diff
@@ -0,0 +1,15 @@
+# Header
```
sradevski, line 1: This should replace the NavBar component instead, and make it a major change.

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+							</NavigationItems>
+							{props.actions && (
+								<Actions display={['none', 'none', 'flex', 'flex']} ml={3}>
+									{props.actions}
+								</Actions>
+							)}
+						</Flex>
+						<Button
+							display={['block', 'block', 'none', 'none']}
+							style={{ fontSize: 20 }}
```
sradevski, line 240: Would applying the fontSize on the icon itself, or on the button using `fontSize` prop work?

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+	const openMenu = () => {
+		setMobileMenuOpen(true);
+	};
+	return (
+		<Wrapper>
+			<Container px={[3, 3, 3, 0]}>
+				<Navbar>
+					<Flex justifyContent="space-between" alignItems="center" width="100%">
+						<Flex alignItems="center">
+							<Link href={props.rootRoute} style={{ lineHeight: 1 }}>
```
sradevski, line 180: As things stand now, this (and other links) will do a full page refresh/navigation, so we need to fix the Link component first to be routers-aware: https://github.com/balena-io-modules/rendition/pull/1137 (it should also support other routers for gatsby and potentially nextjs)

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+							>
+								{mobileMenuOpen && (
+									<Flex
+										display={['block', 'block', 'none', 'none']}
+										justifyContent="flex-end"
+									>
+										<Button
+											plain
+											underline={false}
+											p="5px"
```
sradevski, line 210: Let's use indices for padding and margins (here and in other places)

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+											mt="-8px"
```
sradevski, line 211: I am not a fan of using negative margins if not truly necessary, and this feels like it can be avoided (it seems it looks nice without the margins)

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+											mr="-5px"
+											style={{ opacity: 0.5 }}
```
sradevski, line 213: I don't think this is necessary, maybe we just use a quaternary button instead.

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+				? css`
+						position: fixed;
+						transform: translateX(0%);
+				  `
+				: ''}
+	}
+`;
+
+export const Navbar = styled(Flex)`
+	min-height: 80px;
```
sradevski, line 149: All of the styling for Navbar seems like it should be done using the styled-system props instead

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+			<Container px={[3, 3, 3, 0]}>
+				<Navbar>
+					<Flex justifyContent="space-between" alignItems="center" width="100%">
+						<Flex alignItems="center">
+							<Link href={props.rootRoute} style={{ lineHeight: 1 }}>
+								<Img alt={props.logoAltText} src={props.logo} height="42" />
+							</Link>
+						</Flex>
+						{mobileMenuOpen && (
+							<Fixed
```
sradevski, line 185: Can you check out the `Layer` grommet component? That's what we use for the modal, and it might be usable for the drawer as well. Speaking of which, the mobile option is a drawer, so we can define it as such (we can only have it internal for now)

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+	@media (max-width: ${(props) =>
+			invertBreakpoint(props.theme.breakpoints[1])}px) {
+		position: fixed;
+		background: white;
+		box-shadow: 0px 0px 21px 5px rgba(152, 173, 227, 0.2);
+	}
+`;
+
+const Actions = styled(Box)`
```
sradevski, line 107: Use Flex and styled system props instead

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+			<NavigationItem pl={[0, 0, 3, 3]} key={index} mx={[2, 2, 15]}>
+				<Link
+					blank={route.blank || false}
+					href={route.path}
+					style={{
+						padding: '3px 0 3px',
+						display: 'auto',
+					}}
+					color="text.main"
+					fontSize={13}
```
sradevski, line 80: Let's not override the font unnecessarily, the default should work just fine

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+				</NavigationItem>
+			);
+		}
+
+		return (
+			<NavigationItem pl={[0, 0, 3, 3]} key={index} mx={[2, 2, 15]}>
+				<Link
+					blank={route.blank || false}
+					href={route.path}
+					style={{
```
sradevski, line 75: Let's use styled system props (Link should support both of these, if it doesn't then it's a bug that needs fixing)

src/components/Header/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+	@media (max-width: ${(props) =>
+			invertBreakpoint(props.theme.breakpoints[1])}px) {
+		display: flex;
+	}
+`;
+
+const NavigationItems = styled(Flex)<{ menuOpen?: boolean }>`
+	@media (max-width: ${(props) =>
+			invertBreakpoint(props.theme.breakpoints[1])}px) {
```
sradevski, line 124: Let's try to get away without this and use styled system instead.

LucianBuzzo (approved):

JSReds:
closing for inactivity

PR #1114, opened 2020-05-26
Add support for scrollable tabs

StefKors:
Moving https://github.com/product-os/jellyfish/pull/3737 over to Rendition

Change-type: minor

Connects-to: # 
See: <url> 
Depends-on: https://github.com/balena-io-modules/rendition/pull/1120

---
##### Contributor checklist

##### Reviewer Guidelines
---


StefKors:
@grahammcculloch I've moved the CSS to the component instead of the theme file because it seems to line-up more with the other rendition components like that. I'm happy to move it back if you prefer it in the `theme` file.

I've also added support for CSS scroll snapping as a nice interaction improvement for mobile users.

grahammcculloch:
@StefKors think you need to re-generate the screenshots again for the tests to pass? Once that's done and the PR is out of draft I'll be happy to approve it!

StefKors:
Depends-on: https://github.com/balena-io-modules/rendition/pull/1120

StefKors:
@grahammcculloch could you re-review?

StefKors:
Converting to draft until I add `<` `>` arrows

StefKors:
I'm picking this PR up again. @sradevski After trying out some options to add navigation buttons to the grommet tab component, I'm running into two problems
1. I would need to resort to some easy to break CSS tricks to absolutely position the buttons and add margins + paddings in the correct places
2. To implement the scroll to tab behaviour it would require using react refs. Getting these to the correct spot will also be very messy.

**I want to propose moving forward and keeping the scroll bars for now.** If we still want to have buttons and no scrollbars, I can start a separate PR to first replace the grommetTabs component, with a DIY one or a tabs component from a different component lib that already supports buttons.

@sradevski let me know what you think

src/components/Tabs/story.js
```diff
@@ -36,3 +36,106 @@ storiesOf('Next/Tabs', module)
...
+          <Tab title='This is a very long tab name'>
+            <Txt mt={3}>Here is tab #1</Txt>
+          </Tab>
+          <Tab title='This is another very long tab name'>
+            <Txt mt={3}>Here is tab #2</Txt>
+          </Tab>
+          <Tab title='And this is a third very long tab name testing'>
+            <div>
+              <Txt mt={3}>
+                Hereeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee is tab #3
```
sradevski, line 61: Let's replace this with Lorem Ipsum as a more esthetic option :)

src/components/Tabs/index.tsx
```diff
@@ -1,10 +1,88 @@
...
+			`;
+		}
+	}}
+`;
+
+export const Tab = ({ compact, title, ...props }: InnerTabProps) => {
+	return (
+		<Flex>
+			<TabTitle compact={compact} title={title} {...props}>
+				{/* {title} */}
```
sradevski, line 65: I don't think this is intended

src/components/Tabs/index.tsx
```diff
@@ -1,10 +1,88 @@
...
+import React from 'react';
 import asRendition from '../../asRendition';
+import { Flex } from '../Flex';
 
-const TabsBase = ({ ...props }: TabsProps) => {
-	return <GrommetTabs justify="start" {...props} />;
+interface InnerTabProps extends TabProps {
+	compact?: boolean;
+	index: number;
+	length?: any;
```
sradevski, line 16:
```suggestion
	length?: number;
```

sradevski:
@StefKors the only problem I have with scroll-only option is that it is not clear when there are tabs that are hidden. We can potentially add some blur effect when there is an overflow at the end of he tab container, but I don't think that is any simpler. I'd be fine with it as it is TBH just so we can get the ball rolling, but it is not great UX.

src/components/Tabs/story.js
StefKors: agree :+1:

src/components/Tabs/index.tsx
StefKors: removed :+1:

src/components/Tabs/index.tsx
StefKors: fixed :+1:

StefKors:
@sradevski in this case the scrollbar will show that the tabs are hidden. Shall we move forward with the scroll bar for now, and discuss a better UX using either buttons, blurs or something else at a later date?

sradevski:
@StefKors yeah sure, otherwise we might never get this done :)

sradevski (approved):

sradevski:
@StefKors @grahammcculloch requested changes so he'll have to approve it

grahammcculloch (approved):
:shipit:

grahammcculloch:
@balena-ci rebase

PR #1120, opened 2020-06-01
Replace typeface Circular with Source Sans Pro

StefKors:
Connects-to: #1075 
See: <url> 
Depends-on: <url> 
Change-type: major  

---
##### Contributor checklist

- [ ] I have rebuilt the README with `npm run build:docs`
- [x] I have regenerated screenshots for any affected components with `npm run generate-screenshots`
- [x] I have regenerated jest snapshots for any affected components with `npm run jest -- -u`

##### Reviewer Guidelines
- When submitting a review, please pick:
  - '*Approve*' if this change would be acceptable in the codebase (even if there are minor or cosmetic tweaks that could be improved).
  - '*Request Changes*' if this change would not be acceptable in our codebase (e.g. bugs, changes that will make development harder in future, security/performance issues, etc).
  - '*Comment*' if you don't feel you have enough information to decide either way (e.g. if you have major questions, or you don't understand the context of the change sufficiently to fully review yourself, but want to make a comment)
---

PR #1122, opened 2020-06-02
fix: object filters

JSReds:
replace schema with flat schema to match object filters properties

Depends-on: #1120 
Change-type: patch

Description: 

On the  constructor of the component we run the `flatSchama` function on the schema prop to flat it. Then we use it to match the correct property on the `createFilter`function. At the moment all the filters object that does not have a `description: 'key'` `description: 'value'` will not be correctly represented on the `Summary` component.

Current: 
Screenshot 2020-06-02 at 18 35 02

Updated:
Screenshot 2020-06-02 at 18 34 40

---
##### Contributor checklist

##### Reviewer Guidelines
---


JSReds:
@balena-ci retest

Issue #12, opened 2017-04-20
Add Maps component


LucianBuzzo:
Add a component for displaying a google map, as well as markers on that map.

PR #1134, opened 2020-06-16
Add a Map component

sradevski:
This is a pretty basic component that needs to be extended with features. Instead of passing markers, it takes a data array of any type and a mapping object to specify the required fields (longitude, latitude).

Connects-to: #12
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Map/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+		</Box>
+	);
+};
+
+// Make name optional, and override onChange to not be of `any` type.
+export interface MapProps<T> extends DefaultProps {
+	apiKey: string;
+	data: T[];
+	dataMap: {
+		lat: Array<keyof T>;
```
sradevski, line 180: We want these to be an array of keys for somewhat selfish reasons. In balenaCloud we have two fields where we store longitude and latitude, one that is autocalculated, and the other is set by the user. I think it still is a generic solution where you can provide callbacks if a field is not defined (like for title we can have something like [`name`, `slug`].

src/components/Map/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+const Map = <T extends any>({
+	className,
+	data = [],
+	dataMap,
+	getIcon,
+	onItemClick,
+	apiKey,
+	mapClick,
+}: MapProps<T>) => {
+	const markers = React.useMemo(
```
iamsolankiamit, line 113: Why not just loop below? I don't see any of these operations as complex that they should be memoized.

sradevski: @iamsolankiamit if there are 10000 items for example, this helps with performance.

src/components/Map/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+				isNaN(marker.lat) ? 0 : marker.lat,
+				isNaN(marker.lng) ? 0 : marker.lng,
+			),
+		);
+	});
+	return bounds;
+};
+
+// Fit map to its bounds after the api is loaded
+const apiIsLoaded = (map: google.maps.Map, markers: UIMarker[]) => {
```
grahammcculloch, line 67: nit-pick: maybe rename this to `onApiLoad` to make it clear that it's a callback rather than a method that returns a boolean indicating if the API is loaded or not.

src/components/Map/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+		google.maps.event.removeListener(listener);
+	});
+};
+
+const getFieldFromMap = <T extends any, U extends any>(
+	entry: T,
+	fieldName: keyof MapProps<T>['dataMap'],
+	dataMap: MapProps<T>['dataMap'],
+) => {
+	const field = dataMap[fieldName].find((x) => !!entry[x]);
```
grahammcculloch, line 100:
```suggestion	
	const field = dataMap[fieldName]?.find((x) => !!entry[x]);
```

src/components/Map/index.tsx
grahammcculloch: Would an alternative to this array approach be to allow the `dataMap` fields to be either a `keyof T` value or a func that returns a `keyof T` value? That way you can provide a simple, static data map _or_ you can provide callbacks to determine a lat/lng/id/title value.

sradevski: @grahammcculloch yeah makes perfect sense, I was initially thinking that it would be nice if the props are serializable, but I think having it as a function is better 👍

PR #1136, opened 2020-06-18
Make loading fonts from google optional

zvin:
Changelog-entry: Make loading fonts from google optional
Change-type: patch

### DIFF
--- a/src/components/Provider/index.tsx
+++ b/src/components/Provider/index.tsx
@@ -15,16 +15,24 @@ const Base = styled(Grommet)`
 	color: ${(props) => px(props.theme.colors.text.main)};
 `;
 
-const Provider = ({ theme, ...props }: ThemedProvider) => {
+const defaultFontLink =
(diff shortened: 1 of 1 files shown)


src/components/Provider/index.tsx
```diff
@@ -15,23 +15,32 @@ const Base = styled(Grommet)`
 	color: ${(props) => px(props.theme.colors.text.main)};
 `;
 
-const Provider = ({ theme, ...props }: ThemedProvider) => {
+const defaultFontLink =
+	'https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,600;1,400&family=Ubuntu+Mono:wght@400;700&display=fallback';
+
+const Provider = ({
+	theme,
+	fontLinks = [defaultFontLink],
```
sradevski, line 23: I was thinking more about adding an `offlineFirst` property that we can use to make other decisions as well, such as not trying to load google maps, or captcha, for example. If it is offlineFirst, then we'll first try to load the locally available fonts if the user hasn't overridden the `font` theme property. If you are passing fontLinks, you might as well load them in your application.

zvin:
replaced by #1140

PR #1143, opened 2020-06-23
Generate ts declaration files for esm5 build

zvin:
Change-type: patch

### DIFF
--- a/tsconfig.esm5.json
+++ b/tsconfig.esm5.json
@@ -4,7 +4,7 @@
     "module": "es2015",
     "target": "es5",
     "outDir": "dist_esm5",
-    "declaration": false,
-    "importHelpers": true,
+    "declaration": true,
+    "importHelpers": true
   }
 }


thgreasi:
@zvin just out of curiosity, why was this needed?
I wouldn't expect this to be needed (and the UI in fact works w/o this) except if you are explicitly importing from `rendition/dist_esm5` (which you shouldn't have to since webpack will load the `module` dist as found in the package.json as soon as you configure it to do so).

zvin:
@thgreasi how doe you configure it to do so ?

zvin:
this https://webpack.js.org/configuration/resolve/#resolvemainfields ?

zvin:
@thgreasi I am importing from `rendition/dist_esm5/components/...` to reduce the generated bundle size
Example here https://github.com/balena-io-modules/balena-electron-env/blob/master/src/ui/mounts.tsx#L3-L5

thgreasi:
@zvin while using the latest webpack, you shouldn't really need to do any extra configuration.
As far as you use explicit es6 style imports (like `import { Button } from 'rendition';`), webpack should be able to tree-shake the rest out when creating a `--mode=production` build, w/o you needing to do explicit deep-imports.

Since #869 we have done all the configuration on the rendition side (marking as having no side-effects & defining a `modules` prop pointing to an es modules build in the package.json).

That's how we already do it in the UI and the site.
The respective UI PR only needed to bump rendition.
See: https://github.com/balena-io/balena-ui/pull/2854
See: https://github.com/balena-io/balena-ui/blob/master/webpack.config.js

Here are some pointers:
See: https://webpack.js.org/guides/tree-shaking/
See: https://github.com/webpack/webpack/tree/master/examples/side-effects

zvin:
@thgreasi I have tree-shaking and production mode enabled enabled and replacing 

```typescript
import { Button } from 'rendition';
```
with

```typescript
import Button from 'rendition/dist_esm5/components/Button';
```

saves around 800 kilobytes.

thgreasi:
Did some investigation on this. While tree shaking still works it definitely doesn't 100% so this LGTM for now.

PR #1154, opened 2020-07-14
Add: number match on text search

JSReds:
add the possibility to filter also fields with type number from the search bar

Change-type: patch

this is a PR that has been made to add the possibility to filter also the numeric fields from the search bar as shown in the image below:

Screenshot 2020-07-14 at 18 18 11

Screenshot 2020-08-11 at 14 47 31

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Filters/SchemaSieve.ts
+++ b/src/components/Filters/SchemaSieve.ts
@@ -6,7 +6,6 @@ import defaults from 'lodash/defaults';
 import every from 'lodash/every';
 import findIndex from 'lodash/findIndex';
 import findKey from 'lodash/findKey';
-import includes from 'lodash/includes';
 import map from 'lodash/map';
 import pickBy from 'lodash/pickBy';
 import reduce from 'lodash/reduce';
@@ -32,7 +31,6 @@ export const filter = (
 	const validators = Array.isArray(filters)
 		? filters.map((s) => ajv.compile(s))
 		: [ajv.compile(filters)];
-
 	if (Array.isArray(collection)) {
 		return collection.filter((m) => every(validators, (v) => v(m)));
 	}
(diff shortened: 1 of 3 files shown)


src/components/Filters/SchemaSieve.ts
```diff
@@ -44,21 +42,27 @@ export const createFullTextSearchFilter = (
...
-		schema.properties,
-		(carry, item: JSONSchema, key) => {
+	const items = Object.entries(schema.properties as JSONSchema).reduce(
+		(itemsAccumulator, entry) => {
+			const [key, item] = entry;
 			if (
 				item.type === 'string' ||
-				(Array.isArray(item.type) && includes(item.type, 'string'))
+				item.type === 'number' ||
+				(Array.isArray(item.type) && item.type.includes('string'))
```
sradevski, line 51: I think we generally don't use arrays for `type`, so we can drop this for now, and add it if necessary. That being said, you'd have to check both `number` and `string` for consistency.

src/components/Filters/SchemaSieve.ts
```diff
@@ -44,21 +42,27 @@ export const createFullTextSearchFilter = (
...
 			}
-
-			return carry;
+			return itemsAccumulator;
 		},
-		[] as string[],
+		[] as Array<{ key: string; type: string }>,
 	);
 
+	const filterdItems = isNaN(Number(term))
```
sradevski, line 63:
```suggestion
	const filteredItems = isNaN(Number(term))
```

JSReds: 🙏

src/components/Filters/SchemaSieve.ts
JSReds: this change was made 7 months ago by Lucian, I think it's used in jellyfish.
https://github.com/balena-io-modules/rendition/commit/ac125b02fa665429313cda8e33701b0ea06f0ac3

For now I would add the number check if you agree.

sradevski: @JSReds oh I didn't notice it existed, so yes just adding number is enough.

src/components/Filters/SchemaSieve.ts
```diff
@@ -44,21 +42,28 @@ export const createFullTextSearchFilter = (
...
-
-			return carry;
+			return itemsAccumulator;
 		},
-		[] as string[],
+		[] as Array<{ key: string; type: string }>,
 	);
 
+	const filteredItems = isNaN(Number(term))
+		? items.filter((i) => i.type !== 'number')
```
sradevski, line 65: This won't work if `type` is an array of numbers, for example, it's something to try out and write tests/story for.

JSReds: oh yeah, I've to add also the check on the `itemsAccumulator`. Btw sure I'm going to add some test cases, sorry I forgot. @sradevski

PR #1155, opened 2020-07-15
Add: container obesrver to resize the terminal

JSReds:
add observer to resize terminal on container resize event

Change-type: patch

Unfortunately with the previous PR I didn't realize that I had brought regressions to the terminal. Handling instances of the tty object does not work as expected as the new `Addons API` still has some bugs. I open this PR to add:
1) Terminal sizing based on the parent component resize event and not the window resize event.    
2) Save the `fitAddon` instance within the `ttyInstance` in order to be able to recover it when necessary.

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Terminal/index.tsx
+++ b/src/components/Terminal/index.tsx
@@ -8,6 +8,9 @@ import { Theme as ThemeType } from '../../common-types';
 import theme from '../../theme';
 import defaultXtermStyle from './XTermDefaultStyle';
 import { Flex } from '../Flex';
+// TODO: this can be replaced with types as soon as https://gist.github.com/strothj/708afcf4f01dd04de8f49c92e88093c3
+// or similar will be published as an NPM package
+import { ResizeObserver } from 'resize-observer';
 
 const TtyContainer = styled(Flex)`
 	position: relative;
@@ -38,12 +41,10 @@ const TtyInner = styled(Flex)`
 `;
 
 class Terminal extends React.Component<ThemedTerminalProps, {}> {
-	public readonly tty: Xterm;
+	public readonly tty: Xterm & { fitAddon: FitAddon };
 	// Used as the element to mount XTERM into
 	private mountElement: HTMLDivElement | null;
 	private termConfig: ITerminalOptions;
-	// Addons
-	private fitAddon: FitAddon;
 
 	constructor(props: ThemedTerminalProps) {
 		super(props);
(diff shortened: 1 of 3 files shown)


src/components/Terminal/index.tsx
```diff
@@ -85,6 +86,10 @@ class Terminal extends React.Component<ThemedTerminalProps, {}> {
 			}
 			window.addEventListener('resize', this.resize.bind(this));
```
sradevski, line 87: Why do we need both this and a resize observer?

src/components/Terminal/index.tsx
```diff
@@ -108,7 +113,15 @@ class Terminal extends React.Component<ThemedTerminalProps, {}> {
 	}
 
 	public resize() {
-		this.fitAddon.fit();
+		const mountElementHasHeight =
+			this.mountElement &&
+			parseInt(
```
sradevski, line 118: Why not use `clientHeight` or `scrollHeight` instead?

src/typings/resize-observer.d.ts
```diff
@@ -0,0 +1,242 @@
+/**
+ * The **ResizeObserver** interface reports changes to the dimensions of an
```
sradevski, line 2: Can't we install this as a dependency?

JSReds: it was to avoid one more dependency for a native method, but yes no problem :)

src/components/Terminal/index.tsx
JSReds: we don't, I was sure I removed it... thanks for noticing

src/typings/resize-observer.d.ts
sradevski: @JSReds I meant just for the typings, not for the polyfill, we should use the native method indeed. Surely the native resizeobserver will have typings?

src/components/Terminal/index.tsx
JSReds: I prefer to use the same data used by `fitAddon` to calculate the columns, since if this data is missing the `fit()` function throw an error.
here the fitAddon function that calculate the element height:
Screenshot 2020-07-16 at 12 06 31

JSReds:
@balena-ci retest

JSReds:
@balena-ci retest

PR #1157, opened 2020-07-17
Add: invalid select prop

JSReds:
add invalid select prop

Change-type: patch

This PR add the invalid prop on the select component, to use it when a form is invalid as the input component already does.

Screenshot 2020-07-17 at 20 29 57

The intent is to be able to notify the user in the event that a field is not valid when a form is validated because, for example, it is not correctly selected. In the case of the select it could be the example of a Country or Bank account that cannot have an initial default value (as shown in the image below):

Screenshot 2020-07-17 at 19 31 20

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Select/index.tsx
```diff
@@ -28,9 +40,10 @@ const StyledGrommetSelect = styled(
 `;
 
 // Set button to occupy parent's width, and a default width.
-const Container = styled.div`
+const Container = styled(withTheme(Flex))`
```
sradevski, line 43: `styled` already has access to the theme, no need for `withTheme`. Also, why do we need a `Flex` here, the div is just fine (and I see that it affected the height of the Select in the screenshot)

src/components/Select/index.tsx
```diff
@@ -28,11 +39,13 @@ const StyledGrommetSelect = styled(
 `;
 
 // Set button to occupy parent's width, and a default width.
-const Container = styled.div`
+const Container = styled.div(
+	(props: { theme: any; invalid?: boolean }) => `
```
sradevski, line 43:
```suggestion
	(props: { invalid?: boolean }) => `
```
I am not certain, but I think you don't need to specify theme as a prop (if you do, set the type as `Theme`)

src/components/Select/index.tsx
```diff
@@ -14,6 +13,18 @@ type AdjustedGrommetSelectProps = Omit<
 	'onChange' | 'children'
 >;
 
+const getBorderColor = ({
+	theme,
+	invalid,
+}: {
+	theme: any;
```
sradevski, line 20:
```suggestion
	theme: Theme;
```

sradevski: You can also do this inline (`prop.invalid ? theme.colors.danger.main : 'initial'`)

src/components/Select/index.tsx
sradevski, line 43: You don't need to do this, keep is as `styled.div''` instead. Then, it will be `border-color: ${getBorderColor}` or `border-color: ${props => props.invalid ? theme.colors.danger.main : 'initial'}`

JSReds: Yep sorry @sradevski, it makes perfect sense, it was to avoid the `Property 'invalid' does not exist on type 'IntrinsicAttributes & Pick<Pick<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>` error (related to this: https://github.com/styled-components/styled-components/issues/630), but can be solved in a better way with an interface.

src/components/Select/index.tsx
```diff
@@ -28,9 +28,11 @@ const StyledGrommetSelect = styled(
 `;
 
 // Set button to occupy parent's width, and a default width.
-const Container = styled.div`
+const Container = styled.div<ContainerProps>`
 	& > button {
 		width: 100%;
+		border-color: ${({ invalid, theme }: ContainerProps) =>
+			invalid ? theme?.colors?.danger.main : 'initial'};
```
sradevski, line 35: Ok so `initial` is not the right value here it seems, since it overrides the Select border. Instead if invalid is true, we can return `border-color: theme?.colors?.danger.main`, otherwise we can return `''`.

JSReds: 🙏 sorry again @sradevski, I hadn't tested yet. now it should be ok

PR #1158, opened 2020-07-24
feat: Add JsonSchemaRenderer component

grahammcculloch:
Used to render arbitrary JSON data using a UI schema.

Change-type: minor

---

The purpose of the `JsonSchemaRenderer` component is to provide a read-only alternative to the `Form` component for rendering JSON data based on an accompanying schema and UI schema.

This component was built based on the spec found [here](https://github.com/product-os/product-os/blob/7fe4f0cacf20a2087b6759d0bf35be730ee171be/specs/jellyfish-ui-schema.md).

In addition to the `JsonSchemaRenderer` component, I've added a `JsonSchemaRendererPlayground` component. This is a useful tool for testing out the capabilities of the `JsonSchemaRenderer`. The playground supports fairly good 'dynamic' meta schema generation and validation (see **[this video](https://recordit.co/nV4b9jieLF)** for an example of this in action).

Many of the added files are just simple 'widgets' that act as wrappers around other Rendition components.

This component and the widgets will no doubt evolve over time. But this should be a reasonable starting point to being using it in Jellyfish which will give us a much better idea of what functionality is missing.

image

##### Contributor checklist

##### Reviewer Guidelines
---


grahammcculloch:
@balena-ci retest

grahammcculloch:
@sradevski I'm getting a [Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory](https://ci.balena-dev.com/builds/277457) now on the `ResionCI/docker/build` task. Are we able to increase the memory with something like:
```
export NODE_OPTIONS="--max-old-space-size=<some-larger-value>"
```
Not sure if that is the correct fix or where it should be applied!

sradevski:
@grahammcculloch I'll have a look at it now, not sure why it would run out of memory, storybook builds shouldn't take that much

sradevski:
@grahammcculloch Can you try different source map settings (https://webpack.js.org/configuration/devtool/) and see if it succeeds? I think the culprit is the monaco editor, terser just explodes once it tries to generate source maps for it. Increasing  the max old space is just putting the problem under the rug, so i'd rather compromise on the source maps (which honestly we don't use that much in storybook)

src/extra/JsonSchemaRenderer/JsonSchemaRendererPlayground/JsonEditor.tsx
```diff
@@ -0,0 +1,143 @@
...
+	// Update the editor's value if we are passed a new value
+	React.useEffect(() => {
+		if (!isEqual(value, valueJson)) {
+			setValueStr(JSON.stringify(value, null, 2));
+		}
+	}, [value]);
+
+	// Update the editor's schemas when the schema prop changes
+	React.useEffect(() => {
+		if (schema) {
```
sradevski, line 76: Let's return early instead `if (!schema) return`.

src/extra/JsonSchemaRenderer/JsonSchemaRendererPlayground/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+import { generateUiSchemaMetaSchema } from './util';
+
+const EXTRA_FORMATS: Format[] = [
+	{
+		name: 'markdown',
+		format: '.*',
+	},
+];
+
+const SlimCard = styled(Card)`
```
sradevski, line 32: Generally speaking, overrides like this point to a problem with either the designs or the Card, and one of those should be fixed. The card does need a bit of work to match recent designs, but some are intentional, like the padding (you can use the `small` prop on the card for less padding). Other than that, the card should accept Flex props, and if it doesn't can you look into it? AFAIK the wrapper is a Flex so it should work just fine without all the overrides.

src/extra/JsonSchemaRenderer/JsonSchemaRendererPlayground/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+const EditCard = styled(SlimCard)`
+	& > div > div {
+		flex: 1;
+		display: flex;
+		flex-direction: column;
+		min-height: 0;
+	}
+`;
+
+const ExampleSelect = styled(Select)`
```
sradevski, line 58: Instead of doing this, just wrap the Select in a Box

src/extra/JsonSchemaRenderer/JsonSchemaRendererPlayground/util.ts
```diff
@@ -0,0 +1,123 @@
...
+		'ui:value': {
+			type: ['string', 'object', 'null'],
+		},
+		'ui:options': {
+			type: 'object',
+		},
+	},
+};
+
+export const generateUiSchemaMetaSchema = ({
```
sradevski, line 45: Can we break down the function a bit? It is quite nested and long, so it's difficult to follow what's going on

src/extra/JsonSchemaRenderer/types.ts
```diff
@@ -0,0 +1,36 @@
...
+export type DefinedValue =
+	| number
+	| string
+	| boolean
+	| { [key: string]: any }
+	| any[];
+
+export type Value = DefinedValue | null;
+
+export interface UiSchema {
```
sradevski, line 23: Is this the same with the Form's uiSchema, or at least a subset? If that's the case, let's use `Pick` or `Omit` instead, otherwise it's fine

src/extra/JsonSchemaRenderer/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return isArray(value) ? 'array' : typeof value;
+};
+
+export const getWidget = (value?: Value, uiSchemaWidget?: string) => {
+	const typeWidgets = get(widgets, getType(value), widgets.default);
+	return get(typeWidgets, uiSchemaWidget || 'default', widgets.default.default);
+};
+
+const processValue = ({
```
sradevski, line 54: We can just use `getValue` as that is exactly what `processValue` is :)

src/extra/JsonSchemaRenderer/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+	value,
+	uiSchema,
+}: Pick<InternalJsonSchemaRendererProps, 'value' | 'uiSchema'>) => {
+	return getValue(value, uiSchema);
+};
+
+class JsonSchemaRenderer extends React.PureComponent<
```
sradevski, line 61: Any reason not to implement this with hooks, and just wrap the export in an Error boundary?

src/extra/JsonSchemaRenderer/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+				this.validate(schemaChanged);
+			}
+		}
+	}
+
+	componentDidCatch(error: any) {
+		this.setState({ error });
+	}
+
+	validate(compile: boolean = false) {
```
sradevski, line 103: Let's turn this and the related functions into pure ones that take in all necessary parameters (schema, value) and take it out of the component. It can then just return the new state (if we still go with the class approach)

src/extra/JsonSchemaRenderer/widgets/ui-options.ts
```diff
@@ -0,0 +1,142 @@
...
+	'ml',
+	'mr',
+	'mx',
+	'my',
+].reduce((options, option) => {
+	options[option] = UiOption.space;
+	return options;
+}, SpaceUiOptions);
+
+export const FlexUiOptions: UiOptions = {
```
sradevski, line 64: This is not great, but I can't think of a better approach :)

src/extra/JsonSchemaRenderer/JsonSchemaRendererPlayground/JsonEditor.tsx
```diff
@@ -0,0 +1,143 @@
+import * as React from 'react';
+import * as monacoEditor from 'monaco-editor/esm/vs/editor/editor.api';
+import MonacoEditor, { MonacoEditorProps } from 'react-monaco-editor';
+import { isEqual, get, cloneDeep, findIndex } from 'lodash';
```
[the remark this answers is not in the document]
grahammcculloch: Ah yep - sorry, should have spotted that pattern!

src/extra/JsonSchemaRenderer/JsonSchemaRendererPlayground/index.tsx
grahammcculloch: The Card component is actually surprisingly unwieldy. 
* I've removed the padding override and settled for the 'small' padding (which is still pretty big IMHO).  
* I've moved the background-color override to a `backgroundColor` 
* I've fixed `Card` so that is now accepts the `flex` property
* The `Card` component contains various `Box` components (which are the divs I'm styling in my styled Cards). I don't want to change these to `Flex` components and specify flex properties on them as this would be a breaking change to Rendition. Ok to keep it like this for now? 

On a related note, I'm often replacing `Box` with `Flex` in my own code. I think we often assume a `Box` will suffice when actually a `Flex` copes with many more scenarios. Although obviously this is an over-simplification.

src/extra/JsonSchemaRenderer/index.tsx
grahammcculloch: Good point - I had more in this method but I cut it out and didn't spot that it was now redundant! Removed now.

src/extra/JsonSchemaRenderer/index.tsx
grahammcculloch: Converting to a functional component with hooks has simplified this a lot so I think we're good now!

src/extra/JsonSchemaRenderer/index.tsx
grahammcculloch: Ok - I've created an `ErrorBoundary` component in the `internal` directory and wrapped the new functional `JsonSchemaRenderer` component with it. Let me know if that's what you're thinking or if there's a cleaner way.

src/extra/JsonSchemaRenderer/widgets/ui-options.ts
grahammcculloch: Ha - totally agree! Got me thinking about a Typescript <--> JSONSchema converter!

LucianBuzzo:
[Add JsonSchemaRenderer component to Rendition](https://trello.com/c/IA7j3lLy/89-add-jsonschemarenderer-component-to-rendition)

src/extra/JsonSchemaRenderer/JsonSchemaRendererPlayground/JsonEditor.tsx
```diff
@@ -0,0 +1,143 @@
...
+interface JsonEditorProps {
+	title: string;
+	value: Value;
+	onChange: (value: object) => void;
+	schema?: any;
+}
+
+const JsonEditor = ({ title, value, onChange, schema }: JsonEditorProps) => {
+	const [valueJson, setValueJson] = React.useState(value);
+	const [valueStr, setValueStr] = React.useState(JSON.stringify(value));
```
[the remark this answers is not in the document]
grahammcculloch: Ok - I've refactored it. I've sacrificed some repetitive code for less sources of truth!

Note - we still need to hold separate  object and string state variables to keep things efficient in terms of parsing and updating the result once all 3 inputs are valid.

grahammcculloch:
@balena-ci retest

grahammcculloch:
@balena-ci retest

grahammcculloch:
@sradevski btw I've made various attempts to disable or reduce the source map generation but it always still gives the same error and tbh I can't even be certain my changes to .storybook/webpack.config.js were even being applied.

sradevski:
@grahammcculloch I'm off today so I won't be able to look into it much, but honestly I am also not sure what is happening and why it uses up that much memory. We can try `--max-old-space-size` and see what happens, I'd be happy if we can just fix it at this point. I leave it up to your googling skills as I'd be doing the same 😄

grahammcculloch:
Well - setting `--max-old-space-size` did the trick for sure!

PR #1160, opened 2020-07-30
Use remark+rehype for md parsing

iamsolankiamit:
This enables us to use rendition components for the converted HTML.
That is, instead of plain markup plus css, we can plug our own components.
For ex. using rendition `Link` for anchors instead of plain `<a>` tags.

Breaking changes are in styles and `sanitizerOptions` object keys.

Change-type: Major

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


README.md
```diff
@@ -1092,15 +1092,23 @@ If you need to customize the conversion of markdown to HTML you can supply the `
...
-| `sanitizerOptions` | [`sanitizeHtml.IOptions`](https://github.com/apostrophecms/sanitize-html#how-to-use) | (See `extra/utils/defaultSanitizerOptions`) | - | Specifies the options used when sanitizing the generated HTML. |
+| Name                 | Type                                                                 | Default | Required | Description                                                    |
+| -------------------- | -------------------------------------------------------------------- | ------- | -------- | -------------------------------------------------------------- |
+| `children`           | `string`                                                             | -       | ✓        | The markdown source that should be rendered                    |
+| `sanitizerOptions`   | [`Object`](https://github.com/syntax-tree/hast-util-sanitize#schema) |         | -        | Specifies the options used when sanitizing the generated HTML. |
+| `componentOverrides` | `Object`                                                             | -       | -        | Object specifying component Overrides. ex. `{ p: Txt.p }`      |
```
sradevski, line 1111: Do you need this currently? If not, I'd rather omit it for now, and we can add it later on if truly necessary.

README.md
```diff
@@ -1078,7 +1078,7 @@ tag can be used with `<Txt.p>`.
 
 A simple component for rendering [GitHub flavored markdown](https://github.github.com/gfm/). This component
 sanitizes input.
-This component is not loaded by default as it relies on a markdown parsing package 
+This component is not loaded by default as it relies on a markdown parsing package
```
sradevski, line 1081: Since we use Markdown in the Form, this is not true unfortunately. We need to rethink that and see if we want to drop the markdown support in the form description.

src/extra/Markdown/index.tsx
```diff
@@ -1,743 +1,81 @@
...
-		padding-top: 0.2em;
-		padding-bottom: 0.2em;
-		margin: 0;
+const MarkdownWrapper = styled(Txt)`
+	& code {
 		background-color: rgba(27, 31, 35, 0.05);
+		padding: 0.2em 0px;
+		margin: 0px;
 		border-radius: 3px;
+		font-family: 'Ubuntu Mono', 'Courier New', monospace;
```
sradevski, line 24: Why Ubuntu Mono? Do we need all these custom properties in the wrapper? It feels a bit random TBH

README.md
iamsolankiamit: @sradevski yes, I need it in landr.

src/extra/Markdown/index.tsx
iamsolankiamit: These styles are for inline code blocks, which were there before and we currently lack a codeblock component in rendition for styles.

src/extra/Markdown/index.tsx
```diff
@@ -1,743 +1,81 @@
 import * as React from 'react';
-import marked from 'marked';
-import highlightJS from 'highlight.js';
-import 'highlight.js/styles/default.css';
-import sanitizeHtml from 'sanitize-html';
-import styled from 'styled-components';
 import Txt, { TxtProps } from '../../components/Txt';
-import { parseMarkdown } from '../utils';
```
sradevski, line 8: We need to remove everything in utils that is not used anymore, and remove the unused packages which add up a lot to the bundle

src/extra/Markdown/index.tsx
```diff
@@ -1,743 +1,81 @@
 import * as React from 'react';
-import marked from 'marked';
-import highlightJS from 'highlight.js';
```
sradevski, line 3: Do we still get language highlighting with the new setup? I see there is `rehype-highlight` but it is significantly smaller compared to highlightJS, so I'm just wondering what we're losing and if that is important or not.

src/components/Form/spec.js
```diff
@@ -480,7 +480,7 @@ describe('Form component', () => {
       const description = component.find('.rendition-form-description').first()
 
       expect(description.html()).toMatch(
-        '<p>Lorem ipsum <em>dolor</em> sit amet</p>'
+        '<div id="root_foo__description" class="sc-fzoyAV cFKUiM sc-fzoLag jAOLUb sc-fznNTe glwMIL rendition-form-description"><div><p class="sc-fzoyAV cFKUiM sc-fzpjYC gJiVIn">Lorem ipsum <em>dolor</em> sit amet</p></div></div>'
```
grahammcculloch, line 483: I think you'll have to re-think these tests as the class names will change 'randomly'. Can this be turned into a snapshot test?

src/extra/Markdown/index.tsx
grahammcculloch: Yep - seconded that we want to make sure we don't lose highlighting support. This is very useful in Jellyfish.

(If it turns out highlightJS isn't needed here it should be removed as a dependency entirely (package.json) as this is the only place it's referenced).

iamsolankiamit: `rehype-highlight` uses [`lowlight`](https://github.com/wooorm/lowlight) which support everything supported by `highlight.js`, works only for vdom libs like react, preact, etc. so we only lose weight 🏋️‍♀️

src/extra/Markdown/index.tsx
iamsolankiamit: Sure 👍

src/extra/Markdown/index.tsx
grahammcculloch: I think you can use the exported `monospace` const from `theme.ts` here.

iamsolankiamit:
@grahammcculloch This PR shouldn't add too much weight since we are mostly replacing libs (most of them are light-weight). Also this is building fine now, so it would be helpful if you can test it on a JF branch and let me know if things workout fine.

src/components/Form/spec.js
iamsolankiamit: Good catch, I'll convert these to snapshot tests.

iamsolankiamit:
@balena-ci retest

iamsolankiamit:
@balena-ci retest

iamsolankiamit:
@balena-ci retest

grahammcculloch:
@iamsolankiamit I spent some time today testing out this new Markdown component on Jellyfish (using the beta rendition npm package from this branch).

### 👍 First the good news:
It works. By this I mean I was able to convert my old `sanitizerOptions` into the new `sanitizerOptions` and `componentOverrides` and get the same functionality from the `Markdown` component within Jellyfish as we previously had. 

I do like the ability to define component overrides with React components rather than just tag names and attributes.

### 👎 However... 
1. The generated markdown styling is _significantly_ different to what we had before (and IMHO inferior). See the comparison below:

| v16 | v17 |
|-----|------|
| <a href="https://user-images.githubusercontent.com/2925657/89630256-4c65fd80-d8c9-11ea-83a8-1476e618fa27.jpg"></a> |  <a href="https://user-images.githubusercontent.com/2925657/89630388-80d9b980-d8c9-11ea-9e7d-72afb3a9d405.jpg"></a> |

But that's not really the issue, because...

2. The performance (rendering speed) of the new Markdown component (v17) appears to be much worse than the old version (don't worry about the fact the component is `Anonymous` - it _is_ (just) the Markdown component that is visible on the timeline):

| v16 | v17 |
|-----|-----|
|  First-time render: **29.7ms**<br><br> <a href="https://user-images.githubusercontent.com/2925657/89630778-02c9e280-d8ca-11ea-9827-00ea301edd86.jpg"></a> | First-time render: **454.7ms**<br><br> <a href="https://user-images.githubusercontent.com/2925657/89631146-96031800-d8ca-11ea-9504-1de94a276ea1.jpg"></a> |
|  Re-render: **4.9ms**<br><br> <a href="https://user-images.githubusercontent.com/2925657/89631098-7c61d080-d8ca-11ea-84fa-fe001c7ddfd1.jpg"></a> | Re-render: **42.2ms**<br><br> <a href="https://user-images.githubusercontent.com/2925657/89631235-b4691380-d8ca-11ea-965d-0898ca9e091f.jpg"></a> |

### Conclusion
I think we could sort out the styling quite easily. But I can't see how we can justify a **15x speed decrease** (admittedly based on very simplistic testing). Put simply - this change looks like it would make Jellyfish extremely sluggish.

Caveats: this comparison was obviously done using code compiled and running in dev mode. Production code differences would vary - but probably not enough to make any real difference.

iamsolankiamit:
@grahammcculloch this version `17.0.0-markdown-d7852c3d32095dae991a1fdca62759e9a1026e30` should reduce the first render time to 50ms-70ms. The slowest part in the change was using highlightjs with rehype. Using prismjs fixes the issue. Let me know your thoughts.

grahammcculloch:
@iamsolankiamit sounds promising :-) I will test this out first thing tomorrow if that's alright. Don't want to hold you up - but this definitely has a big impact on Jellyfish!

grahammcculloch:
@iamsolankiamit ok I retested the latest version...

Again, I was measuring the time taken for the first render of `Markdown` component containing the sample Markdown text. I did this on a fresh Jellyfish thread after refreshing the page. I repeated the test three times and took an average.

**Before**: Avg render time: **6.7ms**
**After**: Avg render time: **20.5ms**

So your changes have probably reduced the difference in performance from 15x slower to just 3x slower - which is good. But still concerning that it's significantly slower than the current implementation.

Also, the styling still needs a lot of work to get back to something resembling/equivalent of the current implementation (see the screenshot test of the Markdown component). Any thoughts on this?

iamsolankiamit:
@grahammcculloch I tried reducing the time further down, but that would mean removing features, like `sanitize`, `raw`, etc. We can look into optimising the plugins, if you think the perceived performance is too slow. 

About the styles, the intention of the PR was to be able to use rendition component for the generated html. So the styles will resemble the theme used, work might be needed based on the usecase. I need to render whole page based on markdown, whereas in JF you need to render a message, the styles will vary. Which is the reason we have component overrides.

iamsolankiamit:
We could style some component like `blockquote`, `tables` which aren't defined in rendition at the moment.

iamsolankiamit:
@grahammcculloch Updated the styles for `kbd`, `blockquotes` and `tables`.

src/extra/Markdown/index.tsx
```diff
@@ -1,575 +1,60 @@
...
-	}
-
-	dd {
-		margin-left: 0;
-	}
-
-	code {
-		font-size: 1rem;
+	& code {
+		background-color: rgba(27, 31, 35, 0.05);
```
grahammcculloch, line 25: Any particular reason for this color? I notice from the screenshot files that it's a bit more 'yellowish' than the previous codeblock background.

src/extra/Markdown/index.tsx
```diff
@@ -1,575 +1,60 @@
...
 		padding: 3px 5px;
 		font-size: 11px;
 		line-height: 10px;
 		color: #444d56;
 		vertical-align: middle;
 		background-color: #fafbfc;
-		border: solid 1px #c6cbd1;
-		border-bottom-color: #959da5;
+		border: solid 1px #d1d5da;
+		border-bottom-color: #c6cbd1;
```
grahammcculloch, line 55: Are we able to use theme colors here?

src/extra/Markdown/index.tsx
```diff
@@ -1,575 +1,60 @@
...
-		margin-bottom: 0;
-	}
-
-	li > p {
-		margin-top: 16px;
-	}
-
-	li + li {
-		margin-top: 0.25em;
+		box-shadow: inset 0 -1px 0 #c6cbd1;
```
grahammcculloch, line 57: (Same comment about using theme colors)

grahammcculloch:
@iamsolankiamit I noticed from the screenshot files that the vertical spacing on the rendered `Form` is a bit excessive now. Might need some styling added to bring it back to a reasonable amount?

iamsolankiamit:
@grahammcculloch the vertical spacing in `form` doesn't seem to be by any change in the markdown component. It's probably through some other package which got updated.

src/extra/Markdown/index.tsx
iamsolankiamit: This is exactly the same color used before. the yellowish tint might be due to prismjs

sradevski:
@grahammcculloch do we need to match the styling to the current in JF? Maybe all that is needed is update the style in JF. Otherwise, you should only change the style through a theme, and if that doesn't work well for you we can look into how we can make it work (either update rendition components' design, or expose more theme options).

LucianBuzzo:
@sradevski The styling of JF should follow rendition. I've very concerned about the increase in render time, as this will have a very negative effect in Jellyfish, as we use the `Markdown` component *everywhere*.

iamsolankiamit:
@LucianBuzzo can we test this branch once with JF and see if there is any visible change? since the render time is below 40 ms, so there won't be any perceived speed difference.

iamsolankiamit:
latest version of this pr `v17.0.0-markdown-a39b378da5fe2f65dc59418b07effbc50ef1b580`

grahammcculloch:
@iamsolankiamit @LucianBuzzo I just chatted to @sradevski and we may have come up with a sensible way forward....

1. In Jellyfish, we sanitize the message content on the server before saving it to the database. This avoids us having to sanitize it each and every time we render it! We don't think any other uses of the `Markdown` component require sanitization - but if they do, they'll need to use the same approach.
1. We then remove the `sanitizerOptions` entirely. `componentOverrides` should now give us all the functionality we require. For example you can return `null` from a component override as the equivalent of not allowing that element.

Hopefully by not using any render-time sanitization we should increase the UI performance significantly - enough for this to still be performant on Jellyfish.

Thoughts?

iamsolankiamit:
@grahammcculloch removing sanitiser options  does reduce the re-render time to 10ms on an average. On a second thought we can make the sanitiser optional instead of removing it completely, so we can use it in place where we render user inputs.

sradevski:
@iamsolankiamit what's the user of sanitizing user input on the client side? I'd rather remove it completely to start with, and add it if necessary at some point?

LucianBuzzo:
@sradevski My background is in the Drupal community, and the general rule there is to never trust user-generated input, even if it comes directly from your own database. I think @grahammcculloch suggestions of sanitizing the HTML once in JF is reasonable, however we'd need to add some guarantees about the safety of our data first. This might require additional thought because its not just on messages that we use the markdown component, for example we also use it when rendering card fields (which also contain user input).
The current implementation of the `Markdown` component is incredibly useful, as it provides a clear guarantee about its prevention of various XSS attacks, splitting that out into some separate utility removes this guarantee and makes it harder to judge if our rendering is safe. 
These XSS concerns also exist in landr, where we could also be rendering unsafe user input

iamsolankiamit:
@sradevski @grahammcculloch as @LucianBuzzo said, we need a way to guarantee sanitized output at some places. I think we should keep sanitization enabled by default, and provide interface to disable it, so when we can guarantee that the Markdown is already sanitized we won't need double check it. 

We can extend the above thought to other processor we add to Markdown component, `raw` for raw html, `prismjs` to highlight code. By disabling these where not necessary will further improve the render time. Form `ui:help`, `ui:warning` are few places I think where they are not needed and plain Markdown will suffice the use case, I am sure we can find more such places where they are not required.

This will add an overhead to make sure we are enabling required plugins for performance gains, instead of it being handled by the lib. 

If these don't suffice I can look into the plugins and possibly combine them into a single plugin. This could improve render time, since we will not be running the processor multiple times. The con here being we need to maintain that plugin.

grahammcculloch:
@iamsolankiamit @LucianBuzzo @sradevski given the concerns expressed above, I suggest a reasonable first step is to make the sanitization optional (i.e. if you explicitly set `sanitizerOptions` to `null` you are saying 'don't sanitize this') in this PR. This should hopefully provide the performance improvement needed in Jellyfish for me to unblock this PR and the work that @iamsolankiamit has that depends on this change.

We can then come back to the issue and debate further if we can/should remove sanitization entirely.

TBH I'm less concerned with bundle size than I am with render time.

sradevski:
@grahammcculloch @LucianBuzzo @iamsolankiamit I am fine either way honestly since at the end of the day the reason for removing sanitization is for performance reasons and we can leave the discussion for when (or if) it is an issue. For what is worth, React already escapes everything that is not passed to `dangerouslySetInnerHtml`, and client-side sanitization is not what guarantees the security of a system, it is the data source that has to be clean, so it shouldn't matter that much security-wise.

PR #1161, opened 2020-07-30
Add: Avatar component

JSReds:
Add Avatar component

Change-type: minor

This PR is meant to create an `Avatar` component that will handle 3 different cases:
If `src` is passed as a prop, it will display the image.
If `firstName` and/or `lastName` are passed as props, will show the first letter of each prop.
Otherwise a normal grommet `Avatar` is returned

Screenshot 2020-07-30 at 13 19 49

---
##### Contributor checklist

##### Reviewer Guidelines
---


README.md
```diff
@@ -143,6 +144,20 @@ component.
...
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Avatar/story.js)
+
+#### Props
+
+| Name   | Type   | Default   | Required   | Description   |
+| ------ | ------ | --------- | ---------- | ------------- |
+| `email`| `string`| - | - | The email to fetch the gravatar |
+| `fullName`| `string`| - | - | Full name will be splitted and will grab only the first letters  |
+| `src`  | `string`| - | - | The path to an image that should be displayed in the background |
+| `size`  | `string` | `medium` | - | The size of the avatar (values: 'xsmall' 'small' 'medium' 'large' 'xlarge' or css size )|
```
sradevski, line 158: Normally we don't use Grommet's sizes, as it is too flexible. We probably want 2 sizes, probably normal and `emphasized` (how we do it for the spinner, input, etc.)

README.md
```diff
@@ -143,6 +144,20 @@ component.
...
 The properties of the [`Box`](#box) component are also available.
+### Avatar
+
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Avatar/story.js)
+
+#### Props
+
+| Name   | Type   | Default   | Required   | Description   |
+| ------ | ------ | --------- | ---------- | ------------- |
+| `email`| `string`| - | - | The email to fetch the gravatar |
```
sradevski, line 155: Let's just have `src` and do the transformation of email -> gravatar URL as part of the rendition client for a start

README.md
```diff
@@ -143,6 +144,20 @@ component.
...
+| `fullName`| `string`| - | - | Full name will be splitted and will grab only the first letters  |
```
sradevski, line 156: This feels a bit hacky (what if my name is `First second third fourth`)? How about having first name and last name as separate props?

package.json
```diff
@@ -128,6 +129,7 @@
     "@types/uuid": "^3.4.3",
     "ajv": "^6.7.0",
     "ajv-keywords": "^3.3.0",
+    "blueimp-md5": "^2.17.0",
```
sradevski, line 132: This won't be necessary once we drop the gravatar calculation

README.md
JSReds: IMHO it would be worth to put the `email` prop back in the future, otherwise we will end up implementing the getGravatar function in all projects

README.md
JSReds: TBH I had done exactly for that, it seemed comfortable to be able to pass a `fullName="First second third fourth"` and have as a result `FSTF`. Passing first and last name seems a little restrictive, but we can redo it in the future.

README.md
sradevski: @JSReds let's start without it, and then decide if it is something that will be used in multiple places. Maybe we'll allow uploading photos to the DB, in which case we won't use gravatar.

README.md
sradevski: @JSReds but the avatar won't fit 4 letters for example, and people normally have first name (multiple middle names) last name, so it is still valid.

README.md
```diff
@@ -143,6 +144,20 @@ component.
...
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Avatar/story.js)
+
+#### Props
+
+| Name   | Type   | Default   | Required   | Description   |
+| ------ | ------ | --------- | ---------- | ------------- |
+| `firstName`| `string`| - | - | If defined only the first letter will be taken and shown |
+| `lastName`| `string`| - | - | If defined only the first letter will be taken and shown |
+| `src`  | `string`| - | - | The path to an image that should be displayed in the background |
+| `emphasized`  | `boolean` | false | - | If true it change the Avatar size, it will make it bigger |
```
sradevski, line 158:
```suggestion
| `emphasized`  | `boolean` | false | - | If true, uses a larger size avatar |
```

README.md
```diff
@@ -143,6 +144,20 @@ component.
...
+
+[View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Avatar/story.js)
+
+#### Props
+
+| Name   | Type   | Default   | Required   | Description   |
+| ------ | ------ | --------- | ---------- | ------------- |
+| `firstName`| `string`| - | - | If defined only the first letter will be taken and shown |
+| `lastName`| `string`| - | - | If defined only the first letter will be taken and shown |
+| `src`  | `string`| - | - | The path to an image that should be displayed in the background |
```
sradevski, line 157:
```suggestion
| `src`  | `string`| - | - | URL or base64 encoded image to be displayed in the avatar |
```

src/components/Avatar/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const Avatar = ({
+	src,
+	firstName,
+	lastName,
+	emphasized,
+	...otherProps
+}: AvatarProps) => {
+	const [initials, setInitials] = React.useState<string>();
+
+	React.useEffect(() => {
```
sradevski, line 31: We don't really need an effect and state here, we can just have 
```
const initials = getInitials(firstName, lastName)
```

You can still calculate the initials, even if `src` is passed

src/components/Avatar/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { DefaultProps } from '../../common-types';
+import {
+	Avatar as GrommetAvatar,
+	AvatarProps as GrommetAvatarProps,
+} from 'grommet';
+
+export interface AvatarProps
+	extends DefaultProps,
+		Omit<GrommetAvatarProps, 'size'> {
+	email?: string;
```
sradevski, line 11: I think you forgot this one :)

src/components/Avatar/spec.js
```diff
@@ -0,0 +1,39 @@
...
+      <Provider>
+        <Avatar />
+      </Provider>
+    )
+    let tree = component.toJSON()
+    expect(tree).toMatchSnapshot()
+  })
+
+  describe('firstName and lastName properties', () => {
+    it('should render the Avatar with letters TU', () => {
```
sradevski, line 19:
```suggestion
    it('should render the Avatar with initials if src is not passed', () => {
```

We don't want to specify the assertion value as part of the test name, that's not what we are testing.

src/components/Avatar/spec.js
```diff
@@ -0,0 +1,39 @@
...
+        <Provider>
+          <Avatar firstName='test' lastName='user' />
+        </Provider>
+      )
+      expect(component.text()).toContain('TU')
+    })
+  })
+
+  describe('a null src, valid firstname and lastName properties', () => {
+    it('should render the Avatar with letters EU', () => {
```
sradevski, line 30: Same here

sradevski:
@JSReds just so I don't forget, can you ping the balenaCI or ops to see why the build wasn't required? It automerged despite the build failing.

PR #1166, opened 2020-08-13
feat: Add support for positioning notifications

grahammcculloch:
This is basically a pass-through prop to the underlying react-notifications-component.

Change-type: minor
---

Jellyfish currently has it's own Notifications implementation that uses the `Alert` component. I'd like to change to use the Rendition notifications implementation. But notifications in the top-right are not ideal as they would appear over the 'channel close' button - risking the user accidentally closing a channel if they attempt to click on a notification just after it disappears. So I've exposed the `container` prop to allow the position of the notification to be specified. Default is still `top-right`.

##### Contributor checklist

##### Reviewer Guidelines
---


grahammcculloch:
@balena-ci retest

grahammcculloch:
Ugh - I didn't spot that Rendition also doesn't currently allow the notification `type` to be specified. We need `danger`, `success` etc in Jellyfish notifications. @sradevski are you ok if I add this one other pass-through prop (in a new PR)? Otherwise Rendition notifications are a no-go on Jellyfish. (I'm actually super surprised that this wasn't a default feature. Is there another way of implementing the 'intent' (e.g. with red/green styling) of a notification?

sradevski:
@grahammcculloch We just didn't need it before, but we can definitely add alerts. As a matter of fact, I think you can just modify the DismissableContainer to be the Alert component if there is some type (danger, success), since the two are essentially the same

PR #1168, opened 2020-08-14
feat: Modify notifications API

grahammcculloch:
BREAKING CHANGE: Remove baselineHeight from notifications API

Change-type: major

---

This PR adds one prop to the notifications API and removes one prop:
* ADDED: `type` - can be used to style the notification. Value can be `'danger'`, `'warning'`, `'success'`, or `'info'`.
* REMOVED: `baselineHeight` - the Notification now uses the Rendition `Alert` which calculates and sets the baselineHeight for the `DismissableContainer` that it uses. If you were relying on this prop before, consider changing the layout of your custom notification to ensure there is a line of text at the top (see the new [Standard screenshot](https://github.com/balena-io-modules/rendition/blob/97a10d9f32d585d1ba2c1c161cf3c38670b1daf1/__screenshots__/Next/Notifications/Standard.png) for an example).

##### Contributor checklist

##### Reviewer Guidelines
---


grahammcculloch:
@sradevski confirmed with Alexandria that we do want emphasized and prefix. So I've updated the code to not disable the prefix. Ready for re-review!

PR #1169, opened 2020-08-18
Avatar: add fallback icon

JSReds:
add fallback icon when no props are passed to the avatar component

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Avatar/index.tsx
```diff
@@ -41,8 +45,18 @@ export const Avatar = ({
...
-			{!initials && !src && (
-				<GrommetAvatar size={emphasized ? 'large' : 'medium'} {...otherProps} />
+			{!initials && !src && children && (
+				<GrommetAvatar size={emphasized ? 'large' : 'medium'} {...otherProps}>
+					{children}
+				</GrommetAvatar>
+			)}
+			{!initials && !src && !children && (
+				<GrommetAvatar size={emphasized ? 'large' : 'medium'} {...otherProps}>
+					<FontAwesomeIcon
```
sradevski, line 55: No need to accept children, just default to this if there is no src and initials.

JSReds: it seems limiting IMHO not being able to have a custom element inside the avatar. Why should we avoid a custom element ?

sradevski: @JSReds why would you want a custom element? What else would you put if you neither have initials nor image? It seems completely unnecessary, as anyone would just want an avatar icon in any case

JSReds: A case that comes to mind is when you want to show another icon that is not the default one for example. My only doubt is why we want to limit grommet features (size, Avatar child ...) when we already have them. However I can remove that section without problems, it was just to understand :)

sradevski: @JSReds the reason is it is easier to add props, but it is more difficult to remove them (breaking change, meaning clients will have to make changes as well). We always try to be as constrained and opinionated as possible, and only expand the API surface if truly necessary (and in this case it isn't since we don't have a use-case that requires custom avatar children). This also ensures that we have consistent behavior over all of our projects (all avatar look the same, and are familiar to users).

src/components/Avatar/index.tsx
```diff
@@ -31,18 +33,23 @@ export const Avatar = ({
 		<>
 			{src && (
 				<GrommetAvatar
-					size={emphasized ? 'large' : 'medium'}
+					size={emphasized ? '36px' : '24px'}
```
sradevski, line 36: Hmm I'm fine with it for now, but maybe we want this to be defined using `fontSize` maybe? Let's see later on, it's fine for now.

JSReds: I thought about it but the `fontSizes` does not contain the `36px` that we would like to have when emphasized. We have 34 instead. Same for the background, we can put a default background for sure and then overwrite it with the `background` prop if necessary. I'll open a new PR, so we can discuss it there 👍

PR #1170, opened 2020-08-18
Avatar: add default background and use fontSizes

JSReds:
add default background color, add background prop to overwrite the default value and use fontSizes from theme

See: https://github.com/balena-io-modules/rendition/pull/1169#pullrequestreview-469217211
Change-type: patch

With this PR we would use the theme prop to set the fontSize and the default background color. 
but these are different from what we would like:

fontsize changes from 24px - **36px** (emphasized) to 24px - **34px** (emphasized)

the default background is setted to **#b7bed3** instead

---
##### Contributor checklist

##### Reviewer Guidelines
---


README.md
```diff
@@ -157,6 +157,7 @@ The properties of the [`Box`](#box) component are also available.
 | `lastName`| `string`| - | - | If defined only the first letter will be taken and shown |
 | `src`  | `string`| - | - |  URL or base64 encoded image to be displayed in the avatar |
 | `emphasized`  | `boolean` | false | - | If true, uses a larger size avatar |
+| `background`  | `string` | "#b7bed3" | - | If defined will overwrite the default background color |
```
sradevski, line 160: Again, I don't think we need to expose this, at least to start with. Let's keep it quartenary dark for now, and if it becomes a problem we can allow to override the background.

JSReds: Ok got it @sradevski, this was a proposal for the dashboard use case and also because the avatar is placed on different colours depends on the platform. By the way I'll remove it and in case we'll add it as soon as someone needs it

sradevski: I think the color you chose is pretty good and will apply well to all use-cases, regardless what the dashboard uses now. If necessary we can always change it :)

src/components/Avatar/index.tsx
```diff
@@ -21,31 +25,40 @@ const getInitials = (firstName?: string, lastName?: string) =>
...
+					background={theme.colors.quartenary.dark}
 					{...otherProps}
 				></GrommetAvatar>
 			)}
 			{initials && !src && (
-				<GrommetAvatar size={emphasized ? '36px' : '24px'} {...otherProps}>
-					<Txt>{initials}</Txt>
+				<GrommetAvatar
+					background={theme.colors.quartenary.dark}
+					size={emphasized ? px(theme.fontSizes[6]) : px(theme.fontSizes[5])}
```
JSReds, line 50: NOTE: theme.fontSizes[6] is 34px not 36px

sradevski: Yeah that's fine I think 👍

JSReds:
@sradevski, it's the icon color and background of the svg. We can't use the one we usually use since we can only use the `free-solid-svg-icons` in rendition. That's why I added a children before, am I missing some easy way to change the icon colors ? I know how to change the background but not the user color, the user color and the actual default Avatar background are equal.

JSReds:
delete what i said.  
Done I'll push changes soon

sradevski (approved):

PR #1180, opened 2020-08-28
JsonSchemaRenderer - new features

grahammcculloch:
Change-type: minor

---

Various improvements to the JsonSchemaRenderer component:
* New `MermaidWidget`
* Automatically pick a widget based on the schema's `format` field value
* Auto-color `Badge` and `HightlightedName` widgets based on contact if shade/colors not specified
* Use 'Start Case' for field titles
* Added an `orientation` prop to `ArrayWidget` (horizontal or vertical)
* Added support for date-time formatting in the `Txt` widget using the `ui:options.dtFormat` 
* Reset the `ErrorBoundary` if the props change.

_Note: Ideally these would be separate PRs. But we're a bit pushed to get these updates into Jellyfish._

##### Contributor checklist

##### Reviewer Guidelines
---


LucianBuzzo (approved):

grahammcculloch:
@balena-ci rebase

package.json
```diff
@@ -134,10 +134,11 @@
     "color-hash": "^1.0.3",
     "copy-to-clipboard": "^3.0.8",
     "grommet": "^2.14.0",
-    "json-e": "^4.1.0",
     "hast-util-sanitize": "^3.0.0",
+    "json-e": "^4.1.0",
     "lodash": "^4.17.11",
     "mermaid": "8.4.0",
+    "moment": "^2.27.0",
```
sradevski, line 141: Let's not use `moment` because it is pretty massive (especially if you don't remove the locales), let's use `https://date-fns.org/` instead

src/extra/JsonSchemaRenderer/widgets/widget-util.tsx
```diff
@@ -25,6 +26,22 @@ export interface Widget {
 	supportedTypes?: string[];
 }
 
+export const stringToNumber = function (text: string, max: number) {
```
sradevski, line 29: Why do we need to use bitwise operations, and not use `parseFloat` or `parseInt`?

tslint.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./node_modules/@balena/lint/config/tslint-prettier.json",
   "rules": {
-    "no-shadowed-variable": false
+    "no-shadowed-variable": false,
+    "no-bitwise": false
```
sradevski, line 5: I'd rather keep this (even if it means disabling a single line somewhere), since bitwise operations are pretty difficult to read and reason about.

src/extra/JsonSchemaRenderer/widgets/widget-util.tsx
grahammcculloch: This function doesn't _parse_ the number - it basically creates a hash of a string and then limits it to a number between 0 and `max`. So I don't think `parseFloat` etc would be any use here.

package.json
grahammcculloch: Yeah - I'd prefer `date-fns` too. I can switch to that in this PR but it will probably mean I will end up migrating all our Jellyfish repos from `moment` to `date-fns` too as it doesn't make much sense to use one library in Rendition and another in Jellyfish!

tslint.json
grahammcculloch: Yeah - not even sure why this ended up in the tslint file as I had a single line disable comment in place anyway!

grahammcculloch: ...ahhhh - it's because I was using an inline `eslint-disable-next-line` comment rather than `tslint:disable-next-line`. Fixed now.

src/extra/JsonSchemaRenderer/widgets/widget-util.tsx
sradevski: @grahammcculloch ohh I was misled by the function name and didn't really read the code thoroughly. Can we rename it to something more descriptive, `hashCode` is one naming I've seen in other languages.

src/internal/ErrorBoundary/index.tsx
```diff
@@ -31,6 +32,12 @@ export default class ErrorBoundary extends React.Component<
 		return { error };
 	}
 
+	componentDidUpdate(prevProps: any) {
+		if (!isEqual(prevProps, this.props)) {
```
sradevski, line 36: Deep comparisons are not really recommended, as they can get pretty slow, and props shouldn't be mutated anyway. If you still think you have to keep this, then check if `error` is truth first before doing the deep comparison, so we limit it to only when there is an actual error.

src/extra/JsonSchemaRenderer/widgets/widget-util.tsx
StefKors: @grahammcculloch I was going to say, we have a similar function in the `jellyfish-ui-components` but it's actually the same logic (https://github.com/product-os/jellyfish-ui-components/blob/master/lib/services/helpers.js#L429-L434). I think we should abstract this to a global location and import it here and in the jf ui components

grahammcculloch: @StefKors I'll move it to `utils/index.ts` for now. We can then always reference it in Jellyfish using `import { hashCode } from 'rendition/utils'`

src/internal/ErrorBoundary/index.tsx
grahammcculloch: Yeah fair point. I _think_ this should be ok with a shallow comparison. I'll add the check in for `this.state.error` as well as an additional efficiency.

sradevski (approved):

PR #1181, opened 2020-08-31
Markdown: Add decorator support

karaxuna:
Change-type: minor

---

Sometimes we need to decorate part of the text if it matches some condition. A good example is to display a mention if the text contains `@...`:

```typescript
// Add decorator to display `mention` component when regex matches
const decorators: Decorator[] = [{
  match: new RegExp(`((@{1,2}|#|!{1,2})[a-z\\d-_\\/]+(\\.[a-z\\d-_\\/]+)*)`, 'gmi'),
  component: 'mention'
}];

// Add new component type
const componentOverrides = {
  mention: (props: any) => {
    return <Txt.span {...props} color="red" />
  }
};
```


src/extra/Markdown/index.tsx
```diff
@@ -208,6 +208,17 @@ const decoratorPlugin = (options: DecoratorPluginOptions) => {
...
+						type: 'element',
+						tagName: 'span',
+						children: [
+							{
+								type: 'text',
+								value: text.substring(lastIndex),
+							},
+						],
+					});
+
```
thundron, line 226: what does this fix?

src/extra/Markdown/index.tsx
```diff
@@ -150,6 +150,11 @@ interface DecoratorPluginOptions {
 
 const decoratorPlugin = (options: DecoratorPluginOptions) => {
 	options = options || { decorators: [] };
+
+	if (!options.decorators || !options.decorators.length) {
+		return;
+	}
+
```
thundron, line 157: can we squash this in the above commit(s)?

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,111 @@ type components = {
...
+							children: [
+								{
+									type: 'text',
+									value: text.substring(lastIndex, result.index),
+								},
+							],
+						},
+						{
+							type: 'element',
+							tagName: 'mention',
```
iamsolankiamit, line 206: It would be better if it is a `span` with a class to style it.

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,111 @@ type components = {
...
+	const targets: Array<{ decorator: Decorator; node: any }> = [];
+
+	const visitor = (node: any) => {
+		const matchedDecorator = options.decorators.find((decorator) => {
+			return (
+				node.children &&
+				node.children.length === 1 &&
+				node.children[0].type === 'text' &&
+				decorator.match.test(node.children[0].value) &&
+				node.children[0]
```
iamsolankiamit, line 167: I suppose this is not needed

src/extra/Markdown/index.tsx
karaxuna: There can be a text left after the last match, like this:

```
"start @thundron end"
```

This code adds the `end` part.

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,111 @@ type components = {
 	[element: string]: (props: any) => JSX.Element | React.ReactElement | Element;
 };
 
+interface Decorator {
+	match: RegExp;
+	component: string;
```
iamsolankiamit, line 144: Not sure how and where this is being used

src/extra/Markdown/index.tsx
karaxuna: Thanks for reminding that, I forgot to change tagName to `target.decorator.component`. This is dynamic,  so pass whatever tag you want.

src/extra/Markdown/index.tsx
karaxuna: See the next commit.

src/extra/Markdown/index.tsx
karaxuna: That's true, thanks!

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,111 @@ type components = {
...
+							children: [
+								{
+									type: 'text',
+									value: text.substring(lastIndex, result.index),
+								},
+							],
+						},
+						{
+							type: 'element',
+							tagName: target.decorator.component,
```
iamsolankiamit, line 205: IMO it would be better to have span as default with a class, with some pre-defined styles.

src/extra/Markdown/index.tsx
JSReds:
```suggestion
				node.children &&
				node.children[0] &&
				node.children[0].type === 'text' &&
				decorator.match.test(node.children[0].value)
```

src/extra/Markdown/index.tsx
karaxuna: I don't see how having span as default component would benefit. And we should avoid using plain classes in favor of styled-components.

iamsolankiamit: Did you test this? I am not sure it would work that way

karaxuna: @iamsolankiamit yes, I've manually tested it. Can you share your concerns? You can try yourself by passing these parameters to the markdown component:

```typescript
const decorators: Decorator[] = [{
  match: new RegExp(`((@{1,2}|#|!{1,2})[a-z\\d-_\\/]+(\\.[a-z\\d-_\\/]+)*)`, 'gmi'),
  component: 'mention'
}];

const componentOverrides = {
  mention: (props: any) => {
    return <Txt.span {...props} color="red" />
  }
};
```

iamsolankiamit: Ahh didn't realise we have componentOverrides. We should document this and have a default mention component.

karaxuna: This pr is not about exactly mentions, it just unlocks mentions. Mentions could be added in separate pr. Now this is enough for me to unblock JF bug.

StefKors (approved):

karaxuna:
@balena-ci retest

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,110 @@ type components = {
 	[element: string]: (props: any) => JSX.Element | React.ReactElement | Element;
 };
 
+interface Decorator {
+	match: RegExp;
+	component: string;
+}
+
+interface DecoratorPluginOptions {
```
sradevski, line 147: Let's move everything about the "decorator" plugin in a separate file.

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,110 @@ type components = {
...
+	decorators: Decorator[];
+}
+
+const decoratorPlugin = (options: DecoratorPluginOptions) => {
+	options = options || { decorators: [] };
+
+	if (!options.decorators || !options.decorators.length) {
```
sradevski, line 154: You can eliminate the assignment of the default decorators, and turn the whole check to
```
if(!options?.decorators?.length) { return }
```

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,110 @@ type components = {
...
+		return;
+	}
+
+	const targets: Array<{ decorator: Decorator; node: any }> = [];
+
+	const visitor = (node: any) => {
+		const matchedDecorator = options.decorators.find((decorator) => {
+			return (
+				node.children &&
```
sradevski, line 163: Use optional operand, so this will be `node.children?.[0]?.type === 'text' && node.children[0].value`

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,110 @@ type components = {
...
+	const modify = () => {
+		targets.forEach((target) => {
+			const node = target.node;
+			const match = target.decorator.match;
+
+			const text = node.children[0].value;
+			const children = [];
+
+			match.lastIndex = 0;
+			do {
```
sradevski, line 187: Let's try to refactor this whole function since it's quite difficult to follow what's going on, I couldn't even understand it after reading it for 5 minutes. Let's avoid using `do while` and clean up what is pushed to the children, how we store indices, etc.

src/extra/Markdown/index.tsx
```diff
@@ -138,11 +139,110 @@ type components = {
...
+					break;
+				}
+			} while (match.global);
+
+			node.children = children;
+		});
+	};
+
+	return (tree: any) => {
+		visit(tree, null, visitor);
```
sradevski, line 235: We need to measure the performance impact of traversing the entire tree, and see how it impacts JF in particular.

sradevski (commented):
If this is blocking jellyfish you can just use the npm release for the PR, there are some comments that need to be addressed in the meantime. We will also need some tests for this, as I can imagine there can be plenty of edge cases that are currently not handled (like the if else you have)

karaxuna:
@balena-ci retest

src/extra/Markdown/plugins/decorator.ts
```diff
@@ -0,0 +1,106 @@
...
+import visit from 'unist-util-visit';
+
+const splitTextIntoNodes = (text: string, decorator: Decorator) => {
+	// Reset the regex index state
+	decorator.match.lastIndex = 0;
+
+	const nodes: any[] = [];
+	let lastIndex = 0;
+
+	while (true) {
```
sradevski, line 11: Instead of keeping indices in the decorator, how about doing:
```
const nodes = [];
let lastIndex = 0;
while ((match = re.exec(str)) != null) {
  const matchedString = match[0];
  nodes.push(
    {
      data: str.slice(lastIndex, match.index)
    },
    {
      data: matchedString,
    }
  )

  lastIndex = match.index + matchedString.length
}

if(str.length > lastIndex){
  nodes.push({
    data: str.slice(lastIndex)
  })
}

```

karaxuna: @sradevski it's not me keeping indices in the decorator. `RegExp` becomes stateful when it has `g` (global) flag. So when doing `regex.ecec`, it modifies `regex.lastIndex`. Also linter does not allow `while ((match = re.exec(str)) != null) {`.

sradevski (commented):
LGTM, just add some explanation in the README and generate the screenshot

karaxuna:
@balena-ci retest

sradevski (approved):

PR #1198, opened 2020-10-13
Add dark mode to storybook

sradevski:
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,9 +1,10 @@
 import * as React from 'react'
 import { configure, addDecorator } from '@storybook/react'
+import { useDarkMode } from 'storybook-dark-mode'
 import { withScreenshot } from 'storycap'
 import { createGlobalStyle } from 'styled-components'
 import theme from '../src/theme'
-import {Provider} from '../src/index'
+import { Provider } from '../src/index'
 
 const GlobalStyle = createGlobalStyle([], {
   '*': {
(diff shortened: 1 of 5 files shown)


.storybook/config.js
```diff
@@ -1,35 +1,55 @@
...
   return (
     <React.Fragment>
       <GlobalStyle />
-      <Provider>
-        {storyFn()}
+      <Provider
+        theme={
+          isDark
+            ? {
+                colors: {
```
sradevski, line 31: This is just temporary, we'll need to define how the theme format will look like and define both light and dark in `theme.ts`

StefKors (commented):
@sradevski I've tried running this locally and I ran into some problems:
- `scripts/tsifyCss.js` can't find a couple css files
- `.sotrybook/addons.js`fails with:
```
ERROR in ./.storybook/addons.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@babel/plugin-transform-modules-commonjs' from '/Users/stef/Code/rendition'
```

Is this my dev env or am I missing something?

sradevski:
@StefKors try removing node_modules and reinstalling it, it shouldn't fail (otherwise the build in the CI would as well)

StefKors (approved):

PR #1200, opened 2020-10-16
Add a DataGrid component

sradevski:
Let me know if you imagine a grid differently or if something is missing. I'd like to start simple and not add too many options/functionalities for a start.

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


grahammcculloch:
@sradevski a grid component would be useful. To be honest though this is very different to any other grid component implementation I've used - which are all similar to [Bootstrap's Grid System](https://getbootstrap.com/docs/4.5/layout/grid/). For example here's [Material-UI's Grid component](https://material-ui.com/components/grid/). The feature I particularly like about these is the responsive options (a grid that wasn't responsive would not really be useful in Jellyfish for example). 

Don't know how much more work it would be to implement it like this?

sradevski:
@grahammcculloch I was worried people might assume that this should be a generic grid component, but I couldn't think of a better name. I think what we need is a distinction between a **layout** and a data grid/list. This is meant to be the latter, so it can include pagination, filters, etc. similar to what the Table does. We do need a separate layouting component that can be used in a similar manner as the bootstrap and material UI grids. Does that make sense?

grahammcculloch:
@sradevski ah I see. So the (Data) Grid is a basically a wrapper around a CSS grid - is that fair? (btw `DataGrid` might be a suitable name for this component IMHO). 

I'm curious what the requirement is that prompted this that the Table component couldn't satisfy. (I'm sure there is one - or several!).

sradevski:
@grahammcculloch As I was writing the response I also realized `DataGrid` would be a suitable name, but I didn't want to prime you too much, so let's go with that :) In the dashboard we currently show a grid of cards for the applications, and we will have the same thing for hub, for example. It is a wrapper over CSS grid for sure, it just adds a bit of opinionation on top of it really, so all places where we show the data in a grid view have the same look and feel. We will probably expand it to also have pagination as part of it, but I wanted to start simple for now and add as needed.

sradevski:
@grahammcculloch can I get a review please?

StefKors (commented):
@sradevski I'm don't really understand what the advantage is for including this component in Rendition. Are there features intended to be added to this component that's more than a CSS grid shorthand?

If this is just a constrained shorthand for CSS grid there must be a pre-existing grid component we could use instead of implementing it ourselves.

sradevski:
@StefKors there are features to be added, yes, but even as a shorthand it is better than rewriting the same code over and over again. The advantage of having it in rendition is consistency, similar to other components. Do you have an existing component in mind? I mean even as it is it's not a lot of code, so I'm not sure if an existing component will be better than just creating this wrapper over CSS's grid.

StefKors:
The [Bootstrap's Grid System](https://getbootstrap.com/docs/4.5/layout/grid/) or the [Material-UI's Grid component](https://material-ui.com/components/grid/) could be useful. Knowing a bit more about which features are planned or it's context might be helpful to find a well suited component. @sradevski

sradevski:
@StefKors they have different purposes, the ones you linked are generic Grid systems, mostly from the pre-css grid era. This has a single purpose, to display a list of data in a grid/card layout

The features I'm thinking of are pagination/infinite scroll, filtering, sorting, etc. A lot of things you'd do on a table. See the application list in the dashboard as an example of how this component will be used, it's basically a grid lens.

grahammcculloch (approved):
Let's do this! It's not like it's a complex component that will be difficult to maintain.

PR #1201, opened 2020-10-16
Button: Add confirmation property

karaxuna:
Change-type: minor

---

Usage:

```jsx
<Button confirmation="Are you sure?" onClick={handleClick}>Do smth</Button>
```

Screen Shot 2020-10-22 at 10 37 19 AM
Screen Shot 2020-10-22 at 10 36 56 AM


src/components/Button/index.tsx
```diff
@@ -204,7 +204,7 @@ const getStyledButton = (
 	return ColouredButton;
 };
 
-const Base = (props: ThemedButtonProps) => {
+const Base = React.forwardRef((props: ThemedButtonProps, ref: any) => {
```
sradevski, line 207: Let's type the `ref`

src/asRendition.tsx
```diff
@@ -100,6 +101,25 @@ export const withTooltip = (Base: React.ComponentType) => {
 	});
 };
 
+export const withConfirmation = (Base: React.ComponentType) => {
```
sradevski, line 104: I don't think we need this, since we only want the confirmation to be available to the `button` component for a start. We can always add it in the future if necessary, but let's start with a more limited usage

src/components/Confirm/README.md
```diff
@@ -0,0 +1,3 @@
+# Confirm
```
sradevski, line 1: We can remove this as we don't need this as a component

src/components/Confirm/index.tsx
```diff
@@ -0,0 +1,97 @@
+import * as React from 'react';
```
sradevski, line 1: Let's move this component to `internal`

src/components/Confirm/story.js
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react'
+import { storiesOf } from '@storybook/react'
+import withReadme from 'storybook-readme/with-readme'
+import { Flex, Button } from '../../'
+import ConfirmReadme from './README.md'
+
+storiesOf('Next/Confirm', module)
```
sradevski, line 7: Let's just add this story as part of the Button stories.

src/components/Confirm/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+		}
+	});
+
+	return innerRef;
+};
+
+export const withConfirm = (Base: React.ComponentType) => {
+	return React.forwardRef(
+		({ onClick, confirmProps, ...rest }: any, ref: any) => {
+			const targetRef = useForwardedRef(ref);
```
sradevski, line 63: Why do we need `useForwardedRef` here, why can't we just use `ref`?  I guess because it can be undefined? Maybe `React.useRef(ref)` would work?

sradevski (commented):
There are some things to potentially clean up, but it looks pretty good and simple enough. I'd probably improve the design a bit (add a small triangle-like arrow towards the button and put some margin on the Drop). We can make it much more powerful (like autoselecting the position based on available space) but we don't have to complicate it for now

src/components/Confirm/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { Drop } from 'grommet';
+import { Button, Box } from '../../';
+
+const defaultProps = {
+	align: {
+		bottom: 'top',
+	},
+};
+
+const Confirm = (props: any) => {
```
grahammcculloch, line 11: Can you add a type for the props (rather than `any`)?

grahammcculloch (changes requested):
Could you add a story where the confirmation drop is aligned underneath the button (for example)?

src/components/Confirm/index.tsx
karaxuna: > Why do we need `useForwardedRef` here, why can't we just use `ref`?  I guess because it can be undefined?

Yes.

> Maybe React.useRef(ref) would work?

I don't think just `useRef` will work. Ref is a regular object `{ current: ... }`, so `React.useRef({ current: ... })` will just result in `ref.current.current`.

src/components/Button/index.tsx
grahammcculloch: @karaxuna any reason why you marked this as resolved? I still see the `ref` typed as `any`.

src/internal/Confirm.tsx
```diff
@@ -0,0 +1,165 @@
...
+const StyledDrop = styled(Drop)`
+	border: 0;
+`;
+
+const StyledDropInner = styled.div`
+	margin: 5px;
+	background: white;
+	box-shadow: ${(props) => '1px 1px 5px' + props.theme.colors.gray.light};
+	border-radius: ${(props) => px(props.theme.radius)};
+	border: ${(props) => '1px solid ' + props.theme.colors.gray.main};
```
grahammcculloch, line 17: Can we make the triangle 'background' to be the same color as the background of the `Drop` (with no line between them)

(In due course we'll have to replace 'white' with a theme color as well but I appreciate there isn't a suitable defined color at the moment! We're working on this!)

src/internal/Confirm.tsx
```diff
@@ -0,0 +1,165 @@
...
+	placement?: TooltipArrowProps['placement'];
+}
+
+interface ConfirmProps extends DropProps, ConfirmOptions {
+	targetRef: React.RefObject<HTMLElement>;
+	onOK: () => any;
+	onCancel: () => any;
+}
+
+const Confirm = ({
```
grahammcculloch, line 54: This might be annoying to hear at this stage but this feels like it should be two components:
1. A `Popover` component that just handles the placement and styling of the absolutely positioned element (including triangle) and doesn't care what is inside it.
2. A `Confirm` component  that uses a `Popover` internally and handles the text and cancel & confirm buttons.

There are definitely other uses for the `Popover` so I would promote that to be an exported component. (I'd use it in Jellyfish for sure!)

src/internal/Confirm.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+		if (typeof ref === 'function') {
+			ref(innerRef.current);
+		} else {
+			ref.current = innerRef.current;
+		}
+	});
+
+	return innerRef;
+};
```
grahammcculloch, line 118: Could we add a comment here to explain what's going on here in this component? Or why it's needed?

src/internal/Confirm.tsx
```diff
@@ -0,0 +1,165 @@
...
+export function withConfirm<T extends HTMLElement>(
+	Base: React.ForwardRefExoticComponent<any>,
+) {
+	return React.forwardRef<
+		T,
+		{
+			onClick: (e: React.SyntheticEvent<T>) => any;
+			confirmOptions: ConfirmOptions;
+		}
+	>(({ onClick, confirmOptions, ...rest }, ref) => {
```
grahammcculloch, line 129: As a general comment, It can make it more readable if you define the types separately so that not so much of the inline code is taken up with complex type definitions. 

e.g. 
```typescript
type Confirm<T> = {
			onClick: (e: React.SyntheticEvent<T>) => any;
			confirmOptions: ConfirmOptions;
}
...
React.forwardRef<T, Confirm<T>>
```

grahammcculloch (changes requested):
Main request is regarding splitting up the Confirm component into separate components.

src/components/Popover/README.md
```diff
@@ -0,0 +1 @@
+# Popover
```
grahammcculloch, line 1: Think you need to add details of props in here?

src/index.ts
```diff
@@ -83,6 +83,7 @@ export { default as Map, MapProps } from './components/Map';
 export { default as Navbar, NavbarProps } from './components/Navbar';
 export { default as Link, LinkProps } from './components/Link';
 export { default as List, ListProps } from './components/List';
+export * from './components/Popover';
```
grahammcculloch, line 86: nit-pick: can we keep the exports explicit?
```suggestion
export { default as Popover, PopoverProps, PopoverOptions } from './components/Popover';
```

src/components/Popover/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Drop, DropProps } from 'grommet';
+import styled from 'styled-components';
+import { px } from '../../utils';
+import TooltipArrow, { TooltipArrowProps } from '../../internal/TooltipArrow';
+
+const StyledDrop = styled(Drop)`
+	border: 0;
+`;
+
+const StyledDropInner = styled.div`
```
grahammcculloch, line 11: Do we need to set a `max-width` here? What happens if the content is huge? Maybe `overflow` as well?

grahammcculloch (changes requested):
Looking much cleaner now with the separation of Popover and Confirm!

src/components/Popover/README.md
karaxuna: Sure, I left this for the end when everything else looks fine.

grahammcculloch (approved):
:shipit: great work @karaxuna - that was not an easy one.

Thoughts on the future of these components...

Most UI component libraries I've seen use a pattern where specifying a 'target' is an option. By default you put the target component inside the `Popover` or `Confirm` component as a child component and `Popover` automatically targets it's (first) child component if `target` is not specified.

Additionally, `Popover` components can generally be _controlled_ - e.g. using an `isOpen` prop. This would probably make it easier to work with in other use cases in the future.

cc @sradevski

PR #1204, opened 2020-10-20
Tabs: changed interfaces and props

JSReds:
remove unused props and export extended interfaces

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Tabs/index.tsx
+++ b/src/components/Tabs/index.tsx
@@ -1,22 +1,18 @@
 import styled from 'styled-components';
 import {
-	TabProps,
+	TabProps as GrommetTabProps,
 	Tab as GrommetTab,
 	Tabs as GrommetTabs,
-	TabsProps,
+	TabsProps as GrommetTabsProps,
 } from 'grommet';
-export { TabProps, TabsProps } from 'grommet';
 import React from 'react';
 import asRendition from '../../asRendition';
 
-interface InnerTabProps extends TabProps {
-	compact?: boolean;
-	index?: number;
-	length?: number;
+export interface TabProps extends GrommetTabProps {
 	children?: React.ReactNode;
 }
 
-interface InnerTabsProps extends TabsProps {
+export interface TabsProps extends GrommetTabsProps {
 	compact?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


src/components/Tabs/index.tsx
```diff
@@ -5,14 +5,11 @@ import {
 	Tabs as GrommetTabs,
 	TabsProps,
 } from 'grommet';
-export { TabProps, TabsProps } from 'grommet';
 import React from 'react';
 import asRendition from '../../asRendition';
 
 interface InnerTabProps extends TabProps {
 	compact?: boolean;
```
sradevski, line 12: I guess this prop is internal since it doesn't do anything if you set it per tab

src/components/Tabs/index.tsx
```diff
@@ -70,6 +67,8 @@ const TabTitle = styled(GrommetTab)<{
 	}}
 `;
 
+export { InnerTabProps as TabProps, InnerTabsProps as TabsProps };
```
sradevski, line 70: Let's create `TabProps` that extends `GrommetTabProps` and adds children, and then the `InternalTabProps` extends `TabProps` and it adds `compact`. For the `TabsProps` we don't need separate internal typings cause there is nothing to add there. I know, it's confusing :)

src/components/Tabs/index.tsx
```diff
@@ -5,14 +5,10 @@ import {
 	Tabs as GrommetTabs,
 	TabsProps,
 } from 'grommet';
-export { TabProps, TabsProps } from 'grommet';
 import React from 'react';
 import asRendition from '../../asRendition';
 
 interface InnerTabProps extends TabProps {
-	compact?: boolean;
```
StefKors, line 13: Wait, why is this being removed?

sradevski (changes requested):
I don't think you understood me, I didn't suggest removing the `compact` prop, that is used, I just suggested not including it as part of the exported prop typings.

src/components/Tabs/index.tsx
sradevski: @StefKors after chatting with Andrea it made sense, he added the compact CSS as part of the `Tabs` component, so we can stop doing `children.map`.

sradevski (approved):

src/components/Tabs/index.tsx
JSReds: The `compact` property is usable only on the tabs component and that property is used to make some graphic changes. By adding the Tab style as a nested Tabs style gives the same result and allow to use the property only in one component.

StefKors (approved):
Nice cleanup 👍

Issue #997, opened 2019-10-10
Revisit our usage of storybook


sradevski:
I think we don't utilize the true power of storybook, and visually it is also not very pleasing.

The first step would be to upgrade to the latest version of Storybook. For example, 5.2 has an auto generation feature for documentation. The UI should also be updated to the default theme, as it looks much better than what we have now.

[Here](https://medium.com/storybookjs/storybook-5-2-794958b9b111?utm_source=Chromatic+Early+Access&utm_campaign=c4ceec77e9-EMAIL_CAMPAIGN_2019_06_12_06_26_COPY_01&utm_medium=email&utm_term=0_57e599f726-c4ceec77e9-90831089) is an article about the documentation feature.

We should also investigate what other addons can help us streamline the usage of Rendition.

Issue #811, opened 2019-04-24
If screenshot generation fails due to a faulty story, its very hard to debug


LucianBuzzo:
It would be good to surface page errors from the screenshot generation to help debug any problems.
Adding an option to `storybook-chrome-screenshot` that prints page errors via https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#event-pageerror would be great.
The option could be added here https://github.com/tsuyoshiwada/storybook-chrome-screenshot/blob/master/src/core/app/Browser.ts#L42

PR #1208, opened 2020-11-06
Improve the entire development setup

LucianBuzzo:
Fixes: https://github.com/balena-io-modules/rendition/issues/811
Fixes: https://github.com/balena-io-modules/rendition/issues/997

Notable changes:
- Remove screenshot testing in favor of snapshot testing using storybook stories. We had a lot of issues with screenshot tests, some of which were flakiness, slowing down development and contributions. Although there is a benefit of having screenshot tests, we decided to optimize for development speed instead, since visual changes are not as common recently. We might introduce them again in the future only if they are stable and fast. Generating storybook preview deployments is a better alternative IMO.

- Upgrade to storybook 6. This was a big one, that simplified our tooling (no more custom webpack setups, etc.), but it required a lot of rewrite on how we write stories. One of the main benefits is the autogenerated Docs page for each component where you can interact with the component and change props on the fly. This also means we removed the `README.md` files in favor of inline comments inside the source code that gets picked up by SB 6.

- Moved everything to Typescript. Although the components code was in TS, the tests and stories weren't which meant we had 2 different development tools (eslint + balena-lint), and this just converges to a single tooling stack.

- Removed a lot of snapshot tests (and `spec.ts` files) since the storyshot testing was superior. This means we only need to write unit tests for more complex components, and the storyshot tests should cover the rest.

- Switch to named exports for all components (outside of Extra). This shouldn't be a breaking change unless you imported the components by deep imports (which you shouldn't do)

- Removed babel as it is no longer needed (we use ts-jest for the tests now)

I won't ask you to review the code, but I would ask you to just clone the branch and run it locally, and share any opinions or issues you might notice. There are some rough edges that need to be polished, and I'd prefer to tackle the non-critical ones separately (by putting them here: https://github.com/balena-io-modules/rendition/issues/1215). Also, please try to install the branch version on your project and see if there are any issues (there shouldn't be any, but as a sanity check)


.storybook/main.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  stories: ['../src/**/story.js'],
+  addons: [
+    'storybook-readme',
+    'storybook-dark-mode',
+    '@storybook/addon-actions',
+    'storycap'
```
sradevski, line 7: We don't need this anymore :)

sradevski (commented):
Code-wise LGTM, but I assume it's not working yet since there are no new snapshots :)

src/components/Accordion/index.tsx
```diff
@@ -60,6 +60,8 @@ interface InternalAccordionProps {
 }
 
 export type AccordionProps = InternalAccordionProps & RenditionSystemProps;
+
+/** [View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Accordion/Accordion.stories.tsx) */
```
sradevski, line 64: You can actually see the code in the storybook docs page, so this is unnecessary.

sradevski:
@balena-ci retest

sradevski (approved):

sradevski:
@balena-ci retest

kaisoz:
@balena-ci retest

PR #1212, opened 2020-11-16
Fix accidental complete lodash import

thgreasi:
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---


sradevski (approved):

src/components/Provider/index.tsx
```diff
@@ -1,5 +1,5 @@
 import { Grommet } from 'grommet';
-import { PartialObject, cloneDeep } from 'lodash/index';
```
thgreasi, line 2: I guess this was b/c of an IDE auto-import.
Also dropped `PartialObject` in favor of the TS 2.1 `Partial`, which we already use in the `Map` component.

PR #1216, opened 2020-11-26
Add star rating component

karaxuna:
Change-type: minor

---

Screen Shot 2020-12-01 at 1 18 23 PM


sradevski:
@karaxuna can you please check with Alexandra on the designs, and link them here? I don't want to add components that don't follow our design guide.

apexider:
@karaxuna looking good! I'd just like to suggest different color codes:
* filled star: FCA321
* outlined star: DDE1F0

thank you!

src/components/Accordion/__snapshots__/Accordion.stories.storyshot
```diff
@@ -59,13 +59,7 @@ exports[`Storyshots Core/Accordion Default 1`] = `
           <div
             aria-hidden="true"
             class="StyledBox-sc-13pk1d4-0 dYebPD Collapsible__AnimatedBox-sc-15kniua-0 euNsBL"
-          >
```
sradevski, line 62: This doesn't look right, can you check if something broke? It might have to do with the `setupTests` changes.

scripts/setupTests.js
```diff
@@ -8,9 +8,10 @@ const jsdom = new JSDOM('<!doctype html><html><body></body></html>');
 const { window } = jsdom;
 
 function copyProps(src, target) {
-  Object.defineProperties(target, {
-    ...Object.getOwnPropertyDescriptors(src),
-    ...Object.getOwnPropertyDescriptors(target),
+  Object.entries(Object.getOwnPropertyDescriptors(src)).forEach(([ key, descriptor ]) => {
+    if (!target.hasOwnProperty(key)) {
+      Object.defineProperty(target, key, descriptor);
+    }
```
karaxuna, line 14: Tests fail without this change:

```
should test useBreakpoint hook in multiple resolutions › useBreakpoint should return name: 992 value with 992 resolution

    TypeError: Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'.
```

src/components/Rating/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+interface InternalRatingProps {
+	children?: React.ComponentType;
+	value: number;
+	onChange: (newValue: number) => any;
+}
+
+const BaseRating = React.memo(
+	({ value, onChange, ...rest }: InternalRatingProps) => {
+		const [hoverValue, setHoverValue] = React.useState(0);
```
sradevski, line 51: This looks pretty hacky, why not just depend on CSS to apply hover effects? What I suggest is have an icon SVG and just apply fill on hover: https://css-tricks.com/the-many-ways-to-change-an-svg-fill-on-hover-and-when-to-use-them

karaxuna: If you hover on 4th star, all four stars should be highlighted. How to do that with css?

src/components/Accordion/__snapshots__/Accordion.stories.storyshot
karaxuna: I reverted setupTests change and tried again, snapshots are updated anyway. I suspect that this is due to change in grommet's `GrommetAccordionPanel` which does not hide panel any more, instead it does not render html at all: https://github.com/grommet/grommet/pull/4701/files

sradevski: @karaxuna makes sense, thanks!

src/components/Rating/index.tsx
sradevski: I forgot that all previous stars should be highlighted. You can do something like:
```
<div class="container">
  <a href="#" class="hover">H</a>
  <a href="#" class="hover">H</a>
  <a href="#" class="hover">H</a>
</div>
```

and
```
.container:not(:focus-within):not(:hover) {
  .hover { 
    background: #blue;
  }
}

.hover {
  background: #fad29c;
  &:focus ~ &,
  &:hover ~ & {
    background: #6ef1cd;
  }
}
```
with CSS,but it might become more complicated than a JS solution. I'm fine with it then 👍

sradevski (approved):

karaxuna:
@balena-ci retest

PR #1225, opened 2020-12-11
Add data-* field to form component

vipulgupta2048:
Add data-field attribute to basicinput component. 

Change-type: patch
Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>
See: https://www.flowdock.com/app/rulemotion/p-testbot/threads/XP0RXXksenBf-A3HOGb_erEorWS

### DIFF
--- a/src/components/Form/widgets/BaseInput.tsx
+++ b/src/components/Form/widgets/BaseInput.tsx
@@ -46,6 +46,7 @@ const BaseInput = (props: FormWidgetProps) => {
 				// Form suggestions will clash with browser suggestions if autocomplete is on
 				autoComplete={hasExamples ? 'off' : (options.autoComplete as string)}
 				emphasized={options.emphasized as boolean}
+				data-field={`${schema.type} + ${widgets} + ${placeholder}`}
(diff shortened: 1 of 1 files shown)


src/components/Form/widgets/BaseInput.tsx
```diff
@@ -46,6 +46,7 @@ const BaseInput = (props: FormWidgetProps) => {
 				// Form suggestions will clash with browser suggestions if autocomplete is on
 				autoComplete={hasExamples ? 'off' : (options.autoComplete as string)}
 				emphasized={options.emphasized as boolean}
+				data-field={`${schema.type} + ${widgets} + ${placeholder}`}
```
sradevski, line 49: Why do you need the placeholder? The value should be something like `fieldPath_fieldName`, where fieldPath is going from root to the field (in case of nested objects). Also, we'll need to add it to the other widgets as well, but for simplicity maybe we added to the FieldTemplate until we have custom widgets for everything (sorry about misleading you to go down this path)

LucianBuzzo (approved):

PR #1226, opened 2020-12-14
feat: Add new StatsBar component

grahammcculloch:
This will be used in both Jellyfish and the Balena Dashboard.

Change-type: minor

---
image

image

This PR copies the `StatsBar` component, along with dependencies, from the [DeviceMetrics.tsx file in the balena-ui repo](https://github.com/balena-io/balena-ui/blob/master/src/pages/dashboard/device/device-summary/DeviceMetrics.tsx). 

Once this PR is deployed we can then remove this code from `balena-ui` and reference the Rendition `StarsBar` component instead. We can also make use of it in Jellyfish where there is [currently a copy of this component](https://github.com/product-os/jellyfish/pull/5321).

##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/StatsBar/index.tsx
+++ b/src/components/StatsBar/index.tsx
@@ -1,0 +1,126 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { IconProp } from '@fortawesome/fontawesome-svg-core';
+import range from 'lodash/range';
+import React from 'react';
+import { Box } from '../Box';
+import { Flex, FlexProps } from '../Flex';
+import { ProgressBar, ProgressBarProps } from '../ProgressBar';
+import { Txt } from '../Txt';
+
+export type LabelFormatterProp = (props: {
+	value: number;
+	max: number | null;
+}) => React.ReactElement | string;
+
+export interface StatsBarProps extends Omit<FlexProps, 'title'> {
+	title: React.ReactElement;
+	labelFormatter?: LabelFormatterProp;
+	value: number;
+	max: number | null;
+	numSlices?: number;
+}
+
+export interface StatsTitleProps {
+	icon: IconProp;
+	title: string;
+	description?: string;
+}
+
+interface DiscreteBarProps extends ProgressBarProps {
+	numSlices: number;
+}
+
+export interface ValueWithMaxTitleProps {
+	value: string;
+	max: string;
+}
+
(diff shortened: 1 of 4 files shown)


grahammcculloch:
Note: failing snapshots have been updated in [this PR](https://github.com/balena-io-modules/rendition/pull/1227/commits/10a00f17eab6e9530ab3317d89fdc3ce9db81704) which hopefully can be merged first and we can just rebase this branch.

src/components/StatsBar/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+import range from 'lodash/range';
+import React from 'react';
+import { Box } from '../Box';
+import { Flex, FlexProps } from '../Flex';
+import { ProgressBar, ProgressBarProps } from '../ProgressBar';
+import { Txt } from '../Txt';
+
+export interface StatsBarProps extends Omit<FlexProps, 'title'> {
+	title: React.ReactElement;
+	statsLabel: React.ReactElement;
```
sradevski, line 12: How about having a labelFormatter prop instead, to which we pass `value` and `max`? Like that the caller doesn't have to keep track of those two in two separate places. Since this is a read-only component it doesn't matter as much, but IMO it is a bit cleaner

sradevski (commented):
LGTM, just that one comment

src/components/StatsBar/index.tsx
grahammcculloch: @sradevski ok - I've made that change. Let me know if you think it's suitable or if I've missed something. (Apologies for the force-push).

src/components/StatsBar/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+	return (
+		<Flex flex={1} minWidth={152} flexDirection="column" {...props}>
+			<Flex
+				mb={2}
+				flexDirection="row"
+				alignItems="flex-start"
+				justifyContent="space-between"
+			>
+				<Txt.span>{title}</Txt.span>
+				<Txt.span>{labelFormatter({ value, max })}</Txt.span>
```
sradevski, line 119: NIT: Make `labelFormatter` optional and just render the value as-is

sradevski (commented):
LGTM, feel free to self-certify once you make the labelFormatter optional

grahammcculloch:
@balena-ci I self certify!

PR #1228, opened 2020-12-14
Fix disabled button tooltip

JSReds:
fix tooltip on button disabled

Change-type: patch

this change allow to see the tooltips even on the disabled buttons. (based on react changes on disabled elements https://reactjs.org/blog/2020/02/26/react-v16.13.0.html)

Screenshot 2020-12-14 at 12 06 59

test width 25% || minWidth 25%
Screenshot 2020-12-15 at 12 26 12

test width 50% || minWidth 50%
Screenshot 2020-12-15 at 12 25 58

test width 70% || minWidth 70%
Screenshot 2020-12-15 at 12 25 46

ISSUES:

https://github.com/facebook/react/issues/19419
https://github.com/facebook/react/issues/10109

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/asRendition.tsx
+++ b/src/asRendition.tsx
@@ -92,11 +92,22 @@ export const withStyledSystem = (child: React.ComponentType) => {
 
 export const withTooltip = (Base: React.ComponentType) => {
 	return React.forwardRef(({ ...props }: any, ref: any) => {
-		if (props.tooltip) {
-			props = tooltip.bindProps(props);
+		if (!props.tooltip) {
+			return <Base {...props} ref={ref} />;
 		}
+		const tooltipProps = tooltip.bindProps(props);
 		delete props.tooltip;
-		return <Base {...props} ref={ref} />;
+		return props.disabled ? (
+			<span
+				onClick={tooltipProps.onClick}
+				onMouseEnter={tooltipProps.onMouseEnter}
+				onMouseLeave={tooltipProps.onMouseLeave}
+			>
+				<Base {...props} ref={ref} />
+			</span>
+		) : (
+			<Base {...tooltipProps} ref={ref} />
+		);
 	});
 };
 


src/asRendition.tsx
```diff
@@ -92,11 +92,16 @@ export const withStyledSystem = (child: React.ComponentType) => {
...
-		if (props.tooltip) {
-			props = tooltip.bindProps(props);
+		if (!props.tooltip) {
+			return <Base {...props} ref={ref} />;
 		}
+		const tooltipProps = tooltip.bindProps(props);
 		delete props.tooltip;
-		return <Base {...props} ref={ref} />;
+		return (
+			<div {...tooltipProps}>
```
sradevski, line 101: should we maybe use `span` instead? This might cause some layouting issues if the base was originally inline

thgreasi: That could still have a bad impact if we are using flexbox on the parent :thinking:

JSReds: Nice catch @sradevski, I'm thinking that span can be also a problem for components like `Flex` or `Box` since this would break "children's logic". I therefore also added a last check on the disabled prop.

sradevski:
@JSReds can you try rebasing and installing the prerelease in the dashboard and see if anything breaks?

sradevski (commented):
Feel free to self-certify once you've tested it, LGTM otherwise

JSReds:
@balena-ci I self-certify!

src/asRendition.tsx
```diff
@@ -92,11 +92,22 @@ export const withStyledSystem = (child: React.ComponentType) => {
...
-			props = tooltip.bindProps(props);
+		if (!props.tooltip) {
+			return <Base {...props} ref={ref} />;
 		}
+		const tooltipProps = tooltip.bindProps(props);
 		delete props.tooltip;
-		return <Base {...props} ref={ref} />;
+		return props.disabled ? (
+			<span
+				onClick={tooltipProps.onClick}
```
thgreasi, line 102: @JSReds will this allow users to trigger the `onClick` even on disabled buttons?

sradevski: As I understood it will show the tooltip if you click on it (for mobile devices, for example), but I might have misunderstood

JSReds: @thgreasi @sradevski I confirm that this is for mobile devices since there is no mouseEnter event

PR #1234, opened 2020-12-28
Add truncate property to Txt component

JSReds:
Add truncate property to Txt component to truncate the text when bigger
than the container

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Txt/index.tsx
+++ b/src/components/Txt/index.tsx
@@ -34,6 +34,14 @@ export const bold = (props: ThemedTxtProps) =>
 		  `
 		: null;
 
+export const truncate = (props: ThemedTxtProps) =>
+	props.truncate &&
+	css`
+		overflow: hidden;
+		text-overflow: ellipsis;
+		white-space: nowrap;
+	`;
+
 export const align = style({
 	key: 'text-align',
 	prop: 'align',
@@ -48,6 +56,8 @@ const BaseTxt = styled.div<TxtProps>`
 	${caps}
 	${bold}
 	${italic}
+
+	${truncate}
 `;
 
 const Factory = (tag?: string) => {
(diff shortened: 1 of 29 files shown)


src/components/Txt/index.tsx
```diff
@@ -34,6 +34,15 @@ export const bold = (props: ThemedTxtProps) =>
...
 		: null;
 
+export const truncate = (props: ThemedTxtProps) =>
+	props.truncate
+		? css`
+				overflow: hidden;
+				text-overflow: ellipsis;
+				white-space: nowrap;
+		  `
+		: null;
```
sradevski, line 44: I think it's better to return `''` here

src/components/Txt/index.tsx
```diff
@@ -34,6 +34,14 @@ export const bold = (props: ThemedTxtProps) =>
 		  `
 		: null;
 
+export const truncate = (props: ThemedTxtProps) =>
+	props.truncate &&
```
sradevski, line 38: This will return `false` if truncate is false, and that will be added to the style as a string. We explicitly want to return an empty string here

JSReds: I was using `null` to keep the same pattern with those stated above.  This will not be added as false. 
ref:
```
Speaking of which, during flattening, styled-components ignores interpolations that evaluate to undefined, null, false, or an empty string (""), which means you're free to use short-circuit evaluation to conditionally add CSS rules.
```

https://styled-components.com/docs/advanced

luckily, the above rule applies, otherwise it would have been a problem even putting an empty string :)

sradevski: Empty string should have been fine, but I see SC handles it well :)

sradevski (approved):

JSReds:
@balena-ci retest

PR #1241, opened 2021-01-05
Move JsonSchemaRenderer to core components and remove default widgets…

sradevski:
…/formats

### JsonSchemaRenderer:
This is a breaking change that will require an action. Firstly, if you were
using JsonSchemaRenderer from `extra`, you now need to rename JsonSchemaRenderer
to Renderer, and import it from 'rendition'. The next thing to do is pass
all extra formats that you want the renderer to support that are not default
(namely Markdown and Mermaid).

### Form:
This is a breaking change and requires some action if you
are using extra widgets. You need to remove all `registerWidget`
calls and instead pass the extra widgets you want to use to either
the Form as `extraFormats` or set them on the Provider's `widgets` prop

Moreover, if you used a reference to call `change` and `submit`
you will only be able to call submit now by passing the ref directly to the form
and change was completely removed.

Change-type: major

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Renderer/widgets/ArrayWidget.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
 import map from 'lodash/map';
 import get from 'lodash/get';
-import { Flex } from '../../../components/Flex';
+import { Flex } from '../../Flex';
 import { Widget, WidgetProps, getArrayItems } from './widget-util';
-import { RenditionJsonSchemaRenderer } from '../index';
+import { Renderer } from '../index';
```
grahammcculloch, line 6: This change actually has the effect of adding an `ErrorBoundary` at every level of the Renderer (as `Renderer` includes an `ErrorBoundary` but `RenditionJsonSchemaRenderer` did not). Whilst it might be considered useful in some cases to catch errors at any level down while rendering a nested object, it's probably overkill and inefficient for normal use.

grahammcculloch (commented):
Tested these changes against Jellyfish and everything seems to still work (I verified the Rating, and MarkdownEditor widgets with the Form and the, MarkdownWidget and MermaidWidget widgets with the Renderer).

One slight improvement suggestion as this is a major bump - I noticed that we use a default export from `extra/Renderer/MarkdownWidget` and `extra/Renderer/MermaidWidget`. Perhaps we should change these to named exports in line with other exports? (Probably my fault originally as I tended to use default exports back then!)

sradevski:
@balena-ci I self-certify!

Issue #452, opened 2018-06-15
Deprecate `title` and `titleDetails` on the Modal component


LucianBuzzo:
The `title` and `titleDetails` properties on the Modal component can be removed as the `titleElement` can entirely replace both these properties.

PR #1242, opened 2021-01-08
Actions refactor

JSReds:
Add ActionButtonGroup internal component to allow actions props on modal
and card components

Connects-to: #452
Change-type: major

This PR: 
1) Add actions to Card component
2) Change those of the Modal component to always follow the same model
3) Small restyle Card component
4) Remove properties that are no longer used both on the Card and on the Modal Components.

Screenshot 2021-01-08 at 18 30 14
Screenshot 2021-01-08 at 18 30 04

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/common-types.ts
+++ b/src/common-types.ts
@@ -133,3 +133,11 @@ export interface TooltipProps {
 export interface Tooltip {
 	tooltip?: string | TooltipProps;
 }
+
+export const orderedActionTypes: string[] = ['none', 'primary', 'danger'];
+export interface ActionButtonDefinition {
+	title: string;
+	onTriggerAction: (...args: any) => void;
+	type?: typeof orderedActionTypes[number];
+	disabled?: string | boolean;
+}
--- a/src/components/Card/index.tsx
+++ b/src/components/Card/index.tsx
@@ -1,82 +1,84 @@
-import map from 'lodash/map';
 import * as React from 'react';
 import styled from 'styled-components';
-import { RenditionSystemProps } from '../../common-types';
-
+import {
+	ActionButtonDefinition,
+	RenditionSystemProps,
+} from '../../common-types';
 import asRendition from '../../asRendition';
 import { DismissableContainer } from '../../internal/DismissableContainer';
-import { Box } from '../Box';
-import { Divider } from '../Divider';
 import { Flex, FlexProps } from '../Flex';
 import { Heading } from '../Heading';
 import { useBreakpoint } from '../../hooks/useBreakpoint';
+import { Divider } from '../Divider';
+import { ActionButtonGroup } from '../../internal/ActionButtonGroup';
 
 const Wrapper = styled(DismissableContainer)<WrapperProps>`
 	& {
-		padding: ${(props) => (props.small ? '24px' : '32px 44px')};
+		padding: ${(props) => (props.emphasized ? '32px 44px' : '24px')};
 	}
 `;
 
 const BaseCard = ({
-	title,
-	cta,
-	rows,
+	header,
 	children,
-	small,
+	emphasized,
+	actions,
 	...props
 }: InternalCardProps) => {
-	const hasHeader = title || cta;
-	const isSmall = useBreakpoint(small);
+	const isEmphasized = useBreakpoint(emphasized);
 
-	const Header = hasHeader && (
-		<React.Fragment>
-			<Flex justifyContent="space-between" alignItems="center">
-				<Heading.h5 fontSize={4}>{title}</Heading.h5>
-				{cta}
+	const Header =
+		header &&
+		(typeof header === 'string' ? (
+			<Flex
+				mb={2}
+				flex="0 0 auto"
+				justifyContent="space-between"
+				alignItems="center"
+			>
+				<Heading.h3>{header}</Heading.h3>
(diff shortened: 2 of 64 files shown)


src/common-types.ts
```diff
@@ -133,3 +133,17 @@ export interface TooltipProps {
...
+	'primary',
+	'warning',
+	'danger',
+];
+export interface ActionButtonDefinition {
+	title: string;
+	onTriggerAction: (...args: any) => void;
+	type?: typeof orderedActionTypes[number];
+	disabled?: string | boolean;
+	hidden?: boolean;
```
sradevski, line 148: Hmm I think we can start with not having this, disabled should be enough

src/components/Card/index.tsx
```diff
@@ -1,82 +1,85 @@
...
-			<Flex justifyContent="space-between" alignItems="center">
-				<Heading.h5 fontSize={4}>{title}</Heading.h5>
-				{cta}
-			</Flex>
-			<Divider my={2} />
-		</React.Fragment>
+	const Header = header && (
+		<Flex mb={2}>
+			{typeof header === 'string' ? (
+				<Flex
```
sradevski, line 33: I don't think we need Flex here

src/components/Card/index.tsx
```diff
@@ -1,82 +1,85 @@
...
-		map(rows, (row, index: number) => (
-			<Box key={index} fontSize={2}>
-				{index > 0 && <Divider />}
-				{row}
-			</Box>
-		));
-
 	const Body = children && (
-		<Box fontSize={2} height="100%">
+		<Flex flex="1" flexDirection="column" fontSize={2}>
```
sradevski, line 47: Let's remvoe the fontSize prop

src/components/Modal/index.tsx
```diff
@@ -61,8 +51,6 @@ class BaseModal extends React.Component<ThemedModalProps, any> {
 		if (!BaseModal.mountedCount) {
 			document.body.classList.remove(bodyNoOverflowClass);
 		}
-
-		window.removeEventListener('keydown', this.handleKeyDown);
```
sradevski, line 65: Any reason we don't need this anymore? I think this might have been done for nested modals, not sure..

src/common-types.ts
```diff
@@ -133,3 +133,17 @@ export interface TooltipProps {
 export interface Tooltip {
 	tooltip?: string | TooltipProps;
 }
+
+export const orderedActionTypes: string[] = [
+	'none',
+	'primary',
+	'warning',
```
sradevski, line 140: I really don't think we need a warning action, it's either dangerous or it isn't

src/components/Modal/index.tsx
```diff
@@ -74,14 +62,15 @@ class BaseModal extends React.Component<ThemedModalProps, any> {
 		if (!e.defaultPrevented && e.which === 13) {
 			e.preventDefault();
 			e.stopPropagation();
-
-			if (this.props.primaryButtonProps?.disabled) {
+			const firstPrimaryAction = this.props.actions?.find(
```
sradevski, line 65: Make sure this works when the primary action is dangerous and is to the left ( so I guess the primary will be to cancel)

src/internal/ActionButtonGroup.tsx
```diff
@@ -0,0 +1,47 @@
...
+					),
+			),
+		])
+		.filter((action) => !!action);
+
+	if (!actionEntries.length) {
+		return null;
+	}
+
+	return <>{actionEntries}</>;
```
sradevski, line 46: I think you can just do `return actionEntries`

src/components/Modal/index.tsx
JSReds: this was a mistake, I was trying to move the entire `keydown` event on `ActionsButtons`, but then I realized that we don't want it, and I forgot to add the `removeEventListener` back. :pray:

src/common-types.ts
JSReds: I was thinking that maybe at some point we might need something like: "this action requires a restart" which is not dangerous but warns the user that there will be a consequence. For now, however, I take it off, then we can add it back in case.

src/components/Modal/index.tsx
JSReds: I did it like this, since I would expect that a button without type or type `none`can just be a `cancel`action which is neither primary nor dangerous. But you clearly made me think that the name is wrong :).

JSReds: I did it like this since I was expecting to not see a button with both `primary` and `dangerous` types. Is this thought wrong ?

src/common-types.ts
sradevski: yeah let's do that. It just seems weird figuring out what s dangerous, normal, or a warning, and it's simpler if we just have dangerous/not dangerous (as almost any other app really)

src/components/Modal/index.tsx
sradevski: No you are right, as long as it works fine I'm ok, I just wanted to double-check :)

LucianBuzzo (approved):

JSReds:
closing for inactivity

PR #1249, opened 2021-01-22
Renderer: add support for `ui:explicit`

grahammcculloch:
Change-type: minor

---

_From the [README](https://github.com/balena-io-modules/rendition/blob/0148bbd735cd5ad8cd6a8445775d229b5803e000/src/components/Renderer/index.tsx#L394):_

The default behaviour of the `Renderer` component is to render all fields in the `value` object. If you would like
to restrict the fields that are rendered to the ones you explicitly reference in the UI schema, just set the `ui:explicit`
property in the UI schema. For example, given the following value and UI schema, the `Renderer` will display `name` and `data.email` but _not_ `data.name`:

```javascript
const value = {
  name: 'Joe Bloggs',
  data: {
    name: {
      first: 'Joe',
      last: 'Bloggs'
    },
    email: 'joe@bloggs.com'
  }
}
const uiSchema = {
  data: {
    'ui:explicit': true,
    email: {
      'ui:widget': 'Link'
    }
  }
}
```

This change was discussed in Jellyfish: https://jel.ly.fish/67534398-9397-49f6-b8fb-13394cc7f8e1

##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Form/__snapshots__/Form.stories.storyshot
+++ b/src/components/Form/__snapshots__/Form.stories.storyshot
@@ -48,7 +48,7 @@ exports[`Storyshots Core/Form Custom Submit Text 1`] = `
                         
                         <input
                           autocomplete="off"
-                          class="StyledTextInput-sc-1x30a0s-0 cxIKgN sc-jgHCyG cnloRI sc-gTgzIj eNaHvD"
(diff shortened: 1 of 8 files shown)


grahammcculloch:
@balena-ci I self certify!

PR #1255, opened 2021-02-05
Fix SchemaSieve.flatten for nested required fields

grahammcculloch:
SchemaSieve now handles nested required fields correctly when flattening and unflattening.

##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Filters/SchemaSieve.ts
+++ b/src/components/Filters/SchemaSieve.ts
@@ -6,11 +6,13 @@ import defaults from 'lodash/defaults';
 import every from 'lodash/every';
 import findIndex from 'lodash/findIndex';
 import findKey from 'lodash/findKey';
+import includes from 'lodash/includes';
 import map from 'lodash/map';
 import pickBy from 'lodash/pickBy';
 import reduce from 'lodash/reduce';
 import startsWith from 'lodash/startsWith';
 import trimStart from 'lodash/trimStart';
+import without from 'lodash/without';
 import { FilterSignature } from '.';
 import { randomString, regexEscape } from '../../utils';
 import { getDataModel } from '../DataTypes';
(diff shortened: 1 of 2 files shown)


src/components/Filters/SchemaSieve.ts
```diff
@@ -220,6 +223,22 @@ const flattenAccumulator = (
 			}
 
 			const newKey = parentKey + delimiter + key;
+
+			// Move any required fields into flattened items in the top-level required array
+			forEach(value.required, (requiredKey) => {
```
sradevski, line 228: NIT: Use the native `forEach` instead

src/components/Filters/SchemaSieve.ts
```diff
@@ -220,6 +223,22 @@ const flattenAccumulator = (
...
+				if (!result.required) {
+					result.required = [];
+				} else {
+					// Remove any unnecessary required items pointing at parent fields
+					result.required = without(
```
sradevski, line 233: You could also use `result.required.filter` here instead of lodash

sradevski (commented):
LGTM, just I would prefer if we used the native methods instead of lodash, feel free to self certify once you're happy with the resulting code

src/components/Filters/SchemaSieve.ts
LucianBuzzo: @sradevski Lodash `forEach` has the nice quality of safely handling falsey values, which means we don't have to check for the presence of the `required` field before iterating.

src/components/Filters/SchemaSieve.ts
LucianBuzzo: @sradevski the `without` method is providing utility here as it allows you to shorthand the filtering function

LucianBuzzo:
@sradevski I disagree that we should be using native methods instead of lodash here, as the lodash methods are providing clear utility and cannot be replaced 1:1 with native methods.
In my opinion, we should be preferring native methods of lodash if the lodash method is providing zero additional functionality.
For example, in the following code, lodash is providing zero additional functionality and should be changed to a native method:
```
const arr = [1,2,3]
const sqr = _.map(arr, (x) => x * x)
```
However, something like the following would be providing utility that is not available in a native method and should not be changed:
```
const collection = [
  {
    foo: 1,
    bar: 2
  },
  {
    foo: 3,
    bar: 4
  }
]
const fooArr = _.map(collection, 'foo')
```

Lets me know your thoughts

LucianBuzzo:
TIL we can use the `?` operator in Typescript, so I'll update the foreach to use that, and leave the `without` as is.

grahammcculloch:
@balena-ci rebase

src/components/Filters/SchemaSieve.ts
grahammcculloch: Using `value.required?.forEach` now

grahammcculloch:
@balena-ci I self certify!

PR #1260, opened 2021-02-12
Tag: add support for limiting max items to display

grahammcculloch:
If there are too many items (more than 3), only the
first item is displayed, suffixed with `... and # more`. All
items are then displayed in a tooltip for convenience.

Change-type: minor

---

Overflowing Tags

##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Tag/index.tsx
+++ b/src/components/Tag/index.tsx
@@ -8,6 +8,9 @@ import { Button } from '../Button';
 import { Flex } from '../Flex';
 import { Txt } from '../Txt';
 
+// Prevent the filters taking up too much horizontal space
+const MAX_ITEMS_TO_DISPLAY = 3;
+
 const Container = styled(Flex)`
 	background-color: ${(props) => props.theme.colors.info.light};
 	border: 1px solid ${(props) => props.theme.colors.info.main};
@@ -19,6 +22,19 @@ const DeleteButton = styled(Button)`
 	color: ${(props) => props.theme.colors.tertiary.semilight};
 `;
 
+const tagItemsToString = (items: TagItem[]) => {
+	return items
+		.map((item, index) => {
+			const prefix = index > 0 ? `${item.prefix || ','} ` : '';
+			const separator = item.operator ? ` ${item.operator} ` : ': ';
+			return (
+				prefix +
+				(item.value ? `${item.name}${separator}${item.value}` : item.name)
+			);
+		})
+		.join('\n');
+};
+
 const BaseTag = ({
 	name,
 	operator,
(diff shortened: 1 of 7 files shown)


src/components/Tag/index.tsx
```diff
@@ -90,6 +117,13 @@ const BaseTag = ({
 	);
 };
 
+interface TagItem {
```
sradevski, line 120: NIT: Let's export this as it might be useful outside of the component

src/components/Tag/index.tsx
```diff
@@ -98,17 +132,17 @@ export interface InternalTagProps {
...
-		operator?: string;
-		prefix?: string;
-	}>;
+	multiple?: TagItem[];
 	/** Callback method, that if passed, a "close" button will be added to the right-hand side of the tag */
 	onClose?: (event: React.MouseEvent<HTMLElement, MouseEvent>) => void;
 	/** Callback method, that if passed, the tag will become clickable */
 	onClick?: (event: React.MouseEvent<HTMLElement, MouseEvent>) => void;
 	className?: string;
+	/** If set, limits the max number of items to display if multiple name-value pairs are specified.
```
sradevski, line 141: What I suggest is removing this for now, and just defaulting to 3 in all cases. If we really feel like we want it to be customizable, we can add it in the future, but I think it makes sense as a general rule to limit it to the same value, eg. 3

src/components/Tag/index.tsx
```diff
@@ -19,6 +19,19 @@ const DeleteButton = styled(Button)`
 	color: ${(props) => props.theme.colors.tertiary.semilight};
 `;
 
+const tagItemsToString = (items: TagItem[]) => {
+	return items.reduce((itemsString, item, index) => {
```
sradevski, line 23:
```suggestion
	return items.map((item, index) => {
	       const prefix = index > 0 ? `${item.prefix || ','} ` : '';
         return prefix + item.value ? `${item.name} ${item.operator ?? ':'} ${item.value}` : item.name)
     }
	 .join('\n')
```

sradevski: I think something like this would work and is slightly more concise (and cleaner than using a reducer)

src/components/Tag/index.tsx
LucianBuzzo: This sounds good to me

src/components/Tag/index.tsx
grahammcculloch: Ah - yep, I like that much better!

sradevski (approved):
Looks great 👍

PR #1261, opened 2021-02-16
Add elapsed time widget and filter rule

JSReds:
add elapsed time widget and filter rule

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/DataTypes/index.ts
+++ b/src/components/DataTypes/index.ts
@@ -8,6 +8,9 @@ import * as objectType from './object';
 import * as oneOfType from './oneOf';
 import * as stringType from './string';
 
+const isDateTimeFormat = (format: string | undefined) =>
+	format?.endsWith('date-time');
+
 export const getDataModel = (schema?: JSONSchema) => {
 	if (!schema) {
 		return null;
@@ -25,7 +28,7 @@ export const getDataModel = (schema?: JSONSchema) => {
 		return arrayType;
 	}
 	if (type === 'string') {
-		if (format === 'date-time') {
+		if (isDateTimeFormat(format)) {
 			return dateTimeType;
 		}
 		return stringType;
--- a/src/components/Renderer/widgets/widget-util.tsx
+++ b/src/components/Renderer/widgets/widget-util.tsx
@@ -12,6 +12,7 @@ import get from 'lodash/get';
 import pick from 'lodash/pick';
 import { DefinedValue, JSONSchema, UiSchema, Format, Value } from '../types';
 import { UiOptions } from './ui-options';
+import { formatDistance } from 'date-fns';
 
 export interface WidgetProps {
 	value: DefinedValue;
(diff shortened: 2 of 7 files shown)


src/components/Renderer/widgets/widget-util.tsx
```diff
@@ -28,11 +29,19 @@ export interface Widget {
...
+const DATE_FORMAT = 'MMM do yyyy';
+const TIME_FORMAT = 'h:mm a';
+
 export function formatTimestamp(timestamp: string, uiSchema: UiSchema = {}) {
-	const uiFormat = get(uiSchema, ['ui:options', 'dtFormat']);
-	return uiFormat ? format(new Date(timestamp), uiFormat) : timestamp;
+	const uiFormat =
+		get(uiSchema, ['ui:options', 'dtFormat']) ||
+		`${DATE_FORMAT}, ${TIME_FORMAT}`;
+	return format(new Date(timestamp), uiFormat);
```
JSReds, line 39: Hey @grahammcculloch, would that be still ok for you if I add a default format ?

src/components/DataTypes/index.ts
```diff
@@ -25,7 +25,7 @@ export const getDataModel = (schema?: JSONSchema) => {
 		return arrayType;
 	}
 	if (type === 'string') {
-		if (format === 'date-time') {
+		if (format === 'date-time' || format === 'elapsed-time') {
```
JSReds, line 28: Tempted to also add an `includes('date-time')` to also allow custom extraFormats

src/components/DataTypes/index.ts
```diff
@@ -25,7 +25,7 @@ export const getDataModel = (schema?: JSONSchema) => {
 		return arrayType;
 	}
 	if (type === 'string') {
-		if (format === 'date-time') {
+		if (format?.endsWith('-time')) {
```
JSReds, line 28: Changed check to be able to also support custom date-time format (e.g. custom extraFormat)

sradevski: @JSReds this is too hacky for my taste, let's just create an array of supported widgets for now, it is a bit cleaner in the short-term

grahammcculloch: Can format sometimes just be `'time'`?

src/components/Renderer/widgets/widget-util.tsx
grahammcculloch: @JSReds yes - I think this is a good change and won't cause any issues with existing JF code.

grahammcculloch (commented):
LGTM. I'll wait until Stevche's comment has been addressed before approving though.

src/components/DataTypes/index.ts
JSReds: @sradevski we cannot know the supported widgets since we can add custom widgets in autoUI context. (e.g. in the device list `last_connectivity_event` needs a custom `device-elapsed-time` widget, and it also needs to display the calendar component)

@grahammcculloch sure it makes sense, I was skipping it because I thought we could implement a different filter for time. If it's just time, I don't expect a calendar in the filters but maybe just the time selection. Does this make sense ?

grahammcculloch: @JSReds yep - that makes sense.

sradevski: @JSReds I think there is a limited number of ways how we can represent time (and it should be limited). I don't think we need to have device elapsed time, what does that even mean? Is there some different type of time I don't know about? 😄

JSReds: @sradevski maybe in the future we will have a new type of time 😄 , but it's not what I was considering in this context. There are no differences in time but in presentation. In a normal `elapsed-time` representation you would see something like `14 days ago` in a device elapsed time custom widget you would see `Online (for 4 months) `

JSReds: The solution we came up with by talking about it during a call is to rename the widget `elapsed-date-time` so we can check if the format `endsWith('date-time')` which seems more appropriate.

JSReds:
@balena-ci I self-certify!

PR #1271, opened 2021-03-16
Optimize renderer base

Page-:
Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Renderer/index.tsx
+++ b/src/components/Renderer/index.tsx
@@ -3,7 +3,6 @@ import forEach from 'lodash/forEach';
 import uniqBy from 'lodash/uniqBy';
 import keys from 'lodash/keys';
 import pick from 'lodash/pick';
-import get from 'lodash/get';
 import isArray from 'lodash/isArray';
 import ajv from 'ajv';
 import asRendition from '../../asRendition';
@@ -25,11 +24,13 @@ export const getValue = (
 	schema?: JSONSchema,
 	uiSchema?: UiSchema,
 ) => {
-	const calculatedValue = get(uiSchema, 'ui:value', value);
+	const calculatedValue = uiSchema?.['ui:value'];
 	// Fall back to schema's default value if value is undefined
 	return calculatedValue !== undefined
 		? calculatedValue
-		: get(schema, 'default');
+		: value !== undefined
+		? value
+		: schema?.default;
 };
 
 const widgetWrapperUiOptionKeys = keys(WidgetWrapperUiOptions);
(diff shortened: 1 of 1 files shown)


src/components/Renderer/index.tsx
```diff
@@ -25,14 +25,13 @@ export const getValue = (
...
 	// Fall back to schema's default value if value is undefined
-	return calculatedValue !== undefined
-		? calculatedValue
-		: get(schema, 'default');
+	return calculatedValue ?? schema?.default;
 };
 
-const widgetWrapperUiOptionKeys = keys(WidgetWrapperUiOptions);
+// Convert these to lodash paths once here to avoid having to do it for every single render
+const widgetWrapperUiOptionKeyPaths = keys(WidgetWrapperUiOptions).map(toPath);
```
Page-, line 34: This preparsing is the main win, the path conversion in general is pretty quick but when rendering a large list of items it gets called enough times to be noticeable, and also allocates arrays each time which will contribute to time spent in the gc

src/components/Renderer/index.tsx
```diff
@@ -25,14 +25,13 @@ export const getValue = (
 	schema?: JSONSchema,
 	uiSchema?: UiSchema,
 ) => {
-	const calculatedValue = get(uiSchema, 'ui:value', value);
+	const calculatedValue = uiSchema?.['ui:value'] ?? value;
```
Page-, line 28: These `get` changes were primarily to clean up/simplify the code so it was easier to see what was going on once it has all been transpiled/etc and it does also improve the ability for typescript to apply type-checking

src/components/Renderer/index.tsx
```diff
@@ -25,14 +25,13 @@ export const getValue = (
...
 	// Fall back to schema's default value if value is undefined
-	return calculatedValue !== undefined
-		? calculatedValue
-		: get(schema, 'default');
+	return calculatedValue ?? schema?.default;
```
thgreasi, line 30: From the comment it looks like the `undefined` only check might have been intentional :thinking: 
@grahammcculloch ^ ?

thgreasi (commented):
I'm not confident about the `undefined ? :` => `??` change, but other than that LGTM :+1:

src/components/Renderer/index.tsx
grahammcculloch: @thgreasi yes exactly. `null` is a valid value in this case so we want to explicitly test for `undefined`.

grahammcculloch (changes requested):
Mostly looks good - just one issue where we still need to explicitly check for `undefined`.

Thanks for these optimizations @Page- !

src/components/Renderer/index.tsx
Page-: Urg, you're right. I saw so many cases around the codebase where it's pre `??` that I assumed this was one as well. I've fixed it so explicit `null` vs `undefined` in `ui:value` will also be respected

grahammcculloch (approved):
👍

PR #1272, opened 2021-03-16
Convert usage of lodash/get to native versions

Page-:
This allows for better performance and type-checking

Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


package.json
```diff
@@ -83,7 +83,7 @@
     "strip-css-comments": "^3.0.0",
     "ts-jest": "^26.4.4",
     "ts-node": "^8.0.2",
-    "typescript": "^3.9.6",
+    "typescript": "^4.2.3",
```
Page-, line 86: This adds support for the `h${size}` to be expanded into `'h1' | 'h2' | ...`

src/components/Renderer/RendererPlayground/util.ts
```diff
@@ -53,8 +52,8 @@ const generateArrayMetaSchema = (
 	{ value, schema, uiSchema, extraContext, extraFormats }: MetaSchemaArgs,
 ) => {
 	const arrayValue = value as Value[];
-	const itemsUiSchema = get(uiSchema, 'items', {});
-	const itemsSchema = get(schema, 'items', {});
+	const itemsUiSchema = uiSchema?.items ?? {};
+	const itemsSchema = (schema.items as JSONSchema) ?? {};
```
Page-, line 56: There are casts like this all over the place which were previously hidden behind the `get`, since `get` ends up casting to `any` in most places. These explicit casts may look (and are imo) ugly/bad but they're just exposing issues that already exist in the code so we can no longer ignore them as easily

src/components/Renderer/types.ts
```diff
@@ -28,6 +28,8 @@ export interface UiSchema
 	'ui:title'?: string;
 	'ui:description'?: string;
 	'ui:value'?: Value;
+	'ui:explicit'?: boolean;
+	items?: UiSchema;
```
Page-, line 32: These are used and expected to be typed like this at runtime so I added them here to better represent the true type of UiSchema

src/components/Renderer/widgets/WidgetMeta.tsx
```diff
@@ -17,8 +16,8 @@ export default function WidgetMeta({
 	schema = {},
 	uiSchema = {},
 }: WidgetMetaProps) {
-	const title = get(uiSchema, 'ui:title', schema.title || valueKey);
```
thgreasi, line 20: I hope we were not exploiting empty string titles anywhere...

src/utils/colorUtils.ts
```diff
@@ -83,7 +83,7 @@ export const getColor = (
 
 	const type = getColoringType(props);
 	if (type) {
-		const color = get(props.theme, `colors.${type}`) as Theme['colors']['key'];
+		const color = get(props.theme.colors, type) as Theme['colors']['key'];
```
thgreasi, line 86: You missed this intentionally?
```suggestion
		const color = props.theme.colors?.[type];
```

thgreasi (changes requested):

src/utils/colorUtils.ts
Page-: That's only valid if `type` is a simple string @thgreasi, if it includes `.` or `[` it could do magic, eg `a.b.c` or `a[0].b`. I've been trying to work through these and simplify all the ones where it looks unintentional to support that magic but I haven't convinced myself for all of them yet

thgreasi: Indeed you are right :+1:

src/components/Renderer/widgets/WidgetMeta.tsx
grahammcculloch: Er... I believe we do. We explicitly set `ui:title` to null to prevent it displaying. This is important.

grahammcculloch (changes requested):
I think we need to respect `'ui:title': null` as a way to specify that the title should not be displayed. See inline comment.

src/components/Renderer/widgets/WidgetMeta.tsx
Page-: Fwiw you could set it to `''` and it'd still work the same, but I'll keep the same behavior

src/components/Table/TableRow.tsx
```diff
@@ -26,7 +25,10 @@ export type TableSortFunction<T> = (a: T, b: T) => number;
 export interface TableColumn<T> {
 	cellAttributes?:
 		| React.AnchorHTMLAttributes<HTMLAnchorElement>
-		| ((value: any, row: T) => React.AnchorHTMLAttributes<HTMLAnchorElement>);
+		| ((
+				row: T,
+				value: T[keyof T],
```
Page-, line 30: With the switch away from `get` typescript was able to point out that these typings were wrong, so I've correctly typed them and switched the names around to represent what they actually are. So this doesn't change the existing runtime behavior and instead just fixes the typings

src/components/Table/TableRow.tsx
```diff
@@ -9,7 +8,7 @@ import { Checkbox, CheckboxProps } from '../Checkbox';
  * If a `render` function is available, use it to get the display value.
  */
 const renderField = <T extends {}>(row: T, column: TableColumn<T>): any => {
-	const value = get(row, column.field);
+	const value = row[column.field];
```
Page-, line 11: I looked in the ui and it seems like the `column.field` is always a straight property name - if we do want to support eg `a.b.c` I think it would be better to do it explicitly rather than accidentally

src/components/Renderer/widgets/widget-util.tsx
```diff
@@ -154,7 +154,7 @@ export function getObjectPropertyNames({
 	}
 	const uiSchemaPropertyNames = getMaterializedPropertyNames(uiSchema);
 	const schemaPropertyNames =
-		get(uiSchema, ['ui:order'], keys(get(schema, 'properties'))) || [];
+		(uiSchema?.['ui:order'] ?? keys(schema.properties)) || [];
```
Page-, line 157: This does mean that passing `null` now defaults to `keys(schema.properties)` instead of `[]` - I think that's correct but if not then let me know

src/components/Renderer/widgets/DropDownButtonWidget.tsx
```diff
@@ -23,7 +23,8 @@ const DropDownButtonWidget: Widget = ({
 			`DropDownButtonWidget cannot be used to render a value of type '${typeof value}'`,
 		);
 	}
-	const itemType = get(schema, ['items', 'type'], 'undefined');
+	const itemType =
+		((schema.items as JSONSchema7)?.type as JSONSchema7TypeName) ?? 'undefined';
 	if (!validItemTypes.includes(itemType)) {
```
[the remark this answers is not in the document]
Page-: I'd prefer to leave the PR as just doing the equivalent of what `.get` was doing (outside of the magic that makes things slow, and potentially surprising) - I think there's a whole load of casts that would be worth sorting as part of a different pass/PR in order to improve the ability of typescript to verify the rendition code

src/components/Renderer/RendererPlayground/util.ts
```diff
@@ -156,7 +152,7 @@ export const generateUiSchemaMetaSchema = ({
 	});
 	const input = {
 		// The value may be overridden in the UI Schema
-		value: get(processedUiSchema, 'ui:value', unprocessedValue),
+		value: processedUiSchema['ui:value'] ?? unprocessedValue,
```
Page-, line 155: Again I'm assuming we don't want to handle explicit null here, but let me know

src/components/Renderer/RendererPlayground/index.tsx
```diff
@@ -129,7 +128,7 @@ const RendererPlaygroundBase = ({
 	}, [uiSchemaStr]);
 
 	React.useEffect(() => {
-		const example = get(examples, selectedExample, defaultJsonProps);
+		const example = examples?.[selectedExample] ?? defaultJsonProps;
```
Page-, line 131: If this was an explicit null before then the `example.value` below would have thrown a TypeError so I'm pretty confident the `??` is safe :P

thgreasi (approved):
Awesome :tada:

LucianBuzzo (approved):

PR #1274, opened 2021-03-19
Export multiple components from Renderer component

JSReds:
Export multiple components from Renderer component

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Renderer/index.tsx
+++ b/src/components/Renderer/index.tsx
@@ -4,7 +4,6 @@ import uniqBy from 'lodash/uniqBy';
 import keys from 'lodash/keys';
 import pick from 'lodash/pick';
 import toPath from 'lodash/toPath';
-import isArray from 'lodash/isArray';
 import ajv from 'ajv';
 import asRendition from '../../asRendition';
 import { RenditionSystemProps, Theme } from '../../common-types';
@@ -34,9 +33,6 @@ export const getValue = (
 		: schema?.default;
 };
 
-// Convert these to lodash paths once here to avoid having to do it for every single render
-const widgetWrapperUiOptionKeyPaths = keys(WidgetWrapperUiOptions).map(toPath);
-
 export const getType = (value?: Value) => {
 	if (value === undefined) {
 		return 'default';
@@ -44,7 +40,7 @@ export const getType = (value?: Value) => {
 	if (value === null) {
 		return 'null';
 	}
-	return isArray(value) ? 'array' : typeof value;
+	return Array.isArray(value) ? 'array' : typeof value;
 };
 
 export const getWidget = (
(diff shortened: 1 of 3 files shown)


src/components/Renderer/index.tsx
```diff
@@ -155,21 +185,11 @@ const RendererBase = ({
 	});
 	const processedValue = getValue(value, schema, processedUiSchema);
 
-	if (processedValue === undefined || processedValue === null) {
-		return null;
-	}
```
thgreasi, line 160: This seems to change the behavior.
Was the current behavior a bug?

src/components/Renderer/index.tsx
```diff
@@ -70,6 +66,40 @@ export const getWidget = (
...
+
+export const Widget = ({
+	value,
+	schema,
+	extraFormats,
+	...otherProps
+}: WidgetProps) => {
+	const processedValue = getValue(value, schema);
+
+	if (processedValue === undefined || processedValue === null || !schema) {
```
thgreasi, line 86: Let's simplify it a bit:
```suggestion
	if (processedValue == null || !schema) {
```

src/components/Renderer/index.tsx
```diff
@@ -155,21 +185,11 @@ const RendererBase = ({
 	});
 	const processedValue = getValue(value, schema, processedUiSchema);
```
thgreasi, line 186: I guess we now should be passing the plain `value` to the `Widget`, since the `Widget` is also doing the processing now, right?

src/components/Renderer/index.tsx
```diff
@@ -70,6 +66,40 @@ export const getWidget = (
...
+export interface WidgetProps {
+	value?: Value;
+	schema?: JSONSchema;
+	extraContext?: object | undefined;
+	uiSchema?: UiSchema;
+	uiSchemaWidget?: UiSchema['ui:widget'];
+	extraFormats?: Format[];
+}
+
+export const Widget = ({
```
thgreasi, line 78: Not a strong opinion, but now that we exporting it, how making this a bit more specific?
How about `ValueWidget` or `RendererWidget`?

src/extra/MarkdownEditor/defaultStyle.ts
```diff
@@ -19,7 +19,7 @@ export default css`
 	}
 	.CodeMirror-gutter-filler,
 	.CodeMirror-scrollbar-filler {
-		background-color: transparent;
+		background-color: #fff;
```
thgreasi, line 22: Can we get some context on why these changes are needed?
Could they be in a separate PR?

src/index.ts
```diff
@@ -27,7 +27,13 @@ export {
 	RenditionUiSchema,
 } from './components/Form';
 
-export { Renderer, RendererProps } from './components/Renderer';
+export {
+	Renderer,
+	RendererBase,
```
thgreasi, line 32: Do we also need the RendererBase?
I mean if we only need the Widget for autoUI, then I would avoid also exporting any *Base components.

src/extra/MarkdownEditor/defaultStyle.ts
JSReds: This is auto generated when running npm run generate-snapshots

src/index.ts
JSReds: Sorry this remained from tests. Removing it

src/components/Renderer/index.tsx
JSReds: no you right, changed to value == null

src/extra/MarkdownEditor/defaultStyle.ts
thgreasi: @JSReds let's then have it as the first commit on its own just to be clear.

LucianBuzzo:
@JSReds Why is this change needed? We're breaking an abstraction layer by exposing the internals of this component, making it harder to change in the future (we can't modify the implementation without it being a breaking change).

src/components/Renderer/index.tsx
```diff
@@ -153,7 +184,8 @@ const RendererBase = ({
 		uiSchema,
 		extraContext,
 	});
-	const processedValue = getValue(value, schema, processedUiSchema);
+
+	const processedValue = getValue(value, schema, uiSchema);
```
thgreasi, line 188: Is it equivalent that we are now passing `uiSchema` instead of `processedUiSchema`?

thgreasi (commented):
Other than that LGTM :+1: 
Feel free to self-certify

JSReds:
@LucianBuzzo, The Renderer component is very heavy as an export, as it carries asRenditon with it and has validation checks ( that in multiple cases are completely skipped). By dividing the component in this way, we would be able in an autoUI context to bring with us only the Widget component which is what we need in a cell context (cell of a table). I was splitting the component in order to avoid any duplication in the code. But if this is inconvenient for others, we can safely create another component. What do you think ?

LucianBuzzo:
@JSReds If the intention is to re-use the individual widget fields in multiple places then the widgets should probably be a separate component, though this can happen as a follow-up PR if its deemed necessary.

JSReds:
Closing in favour of a PR to come as AutoUI will be inside Rendition and will have the chance to use the component internally

PR #1275, opened 2021-03-22
Fix tests reverting typescript version to 3.9.6

JSReds:
Revert typescript version to 3.9.6 to fix several tests errors

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "strip-css-comments": "^3.0.0",
     "ts-jest": "^26.5.4",
     "ts-node": "^9.1.1",
-    "typescript": "~4.1.5",
+    "typescript": "^3.9.6",
     "zxcvbn": "^4.4.2"
   },
   "dependencies": {
--- a/src/components/Renderer/index.tsx
+++ b/src/components/Renderer/index.tsx
@@ -155,7 +155,7 @@ const RendererBase = ({
 	});
 	const processedValue = getValue(value, schema, processedUiSchema);
 
-	if (processedValue === undefined || processedValue === null) {
+	if (processedValue == null) {
 		return null;
 	}
 
@@ -166,7 +166,7 @@ const RendererBase = ({
 	);
 	const Widget = getWidget(
 		processedValue,
-		schema.format,
+		schema?.format,
 		processedUiSchema['ui:widget'],
 		formats,
 	);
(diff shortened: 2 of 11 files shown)


src/components/Renderer/index.tsx
```diff
@@ -25,7 +25,7 @@ export const getValue = (
 	schema?: JSONSchema,
 	uiSchema?: UiSchema,
 ) => {
-	const calculatedValue = uiSchema?.['ui:value'];
+	const calculatedValue = uiSchema?.['ui:value'] ?? value;
```
thgreasi, line 28: The previous refactor on this didn't actually change the behavior at all.
In my attempts, the tests passed even w/o this.

Page-: Yeah this is wrong as it treats null and undefined the same @JSReds, you can see below we specifically handle null vs undefined separately and do the defaulting to value as need be

src/components/Renderer/index.tsx
```diff
@@ -166,7 +166,7 @@ const RendererBase = ({
 	);
 	const Widget = getWidget(
 		processedValue,
-		schema.format,
+		schema?.format,
```
Page-, line 169: Can you update `ThemedRendererProps` to denote that `schema` is actually optional and can be undefined please

package.json
```diff
@@ -83,7 +83,7 @@
     "strip-css-comments": "^3.0.0",
     "ts-jest": "^26.5.4",
     "ts-node": "^9.1.1",
-    "typescript": "~4.1.5",
+    "typescript": "^3.9.6",
```
Page-, line 86: I'd prefer to make things work with the newer typescript rather than pin to 3.x if possible

Page- (changes requested):
Request changes for the null vs undefined issue

src/components/Renderer/index.tsx
JSReds: Just realized that this is not making much sense, I'm going to check the examples and fix them

thgreasi: This seems to only happen for the examples about materialized and invalid fields.
See: https://github.com/balena-io-modules/rendition/blob/2becdac6dfc642f6e02ec7c5fa5b5c90343c3823/src/components/Renderer/examples.ts#L116-L126
See: https://github.com/balena-io-modules/rendition/blob/2becdac6dfc642f6e02ec7c5fa5b5c90343c3823/src/components/Renderer/examples.ts#L341-L366

Imo the typings are correct, it's just that atm we try to use the Renderer recursively before checking whether the resulting schema is valid. What'st the correct thing to do for the materialized field though is harder to answer.

package.json
thgreasi: I agree on principle @Page- , but the TS bump has caused a weird `window.matchMedia is undefined` error in storybook, so it's not a TS compilation issue thing.
I do see some profit on getting master to have its tests passing asap and fix the storybook issue in a follow-up PR.

thgreasi (approved):

Page- (approved):

PR #1281, opened 2021-04-06
Table with custom columns

JSReds:
Add Table column selector component

Change-type: minor

This PR is adding new functionalities on the Table component. Until now these are custom features that we have in external projects, but since these features can be useful, we decided to bring them directly to rendition.

- Custom columns: columns that can be showed and hidden 
- Save Preferences: save showed and hidden columns and sorting preferences.
- Checkbox column scroll while scrolling the table

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/components/Table/TableBase.tsx
```diff
@@ -0,0 +1,678 @@
+import { faSort } from '@fortawesome/free-solid-svg-icons/faSort';
```
thgreasi, line 1: Could we have a single commit w/o change type that only includes a `git mv` from the old table to the TableBase.tsx plz?
It will definitely help with history as well as making it easier to review this PS.

JSReds:
closing in favour of #1281

PR #1301, opened 2021-06-11
Add array oneOf filter case

JSReds:
Add array oneOf filter case

Change-type: minor

This PR will add a new array filter case, where we want to be able to make a many-to-many relation. This will allow to filter an array type with a Select component.

Screenshot 2021-06-11 at 11 35 03
Screenshot 2021-06-11 at 11 35 32

Example of use case:

A table column display an array of results and we want to allow the user to filter this array with an array of possible items that are contained on the column array:

(column Works with)
Screenshot 2021-06-11 at 11 50 20
Screenshot 2021-06-11 at 11 52 21

---
##### Contributor checklist

##### Reviewer Guidelines
---


JSReds:
@balena-ci I self certify!

PR #1310, opened 2021-06-22
Add analytics client

JSReds:
Add analytics client

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/extra/AutoUi/Collection/Filters.tsx
+++ b/src/extra/AutoUi/Collection/Filters.tsx
@@ -7,6 +7,8 @@ import {
 	Filters as RenditionFilters,
 } from '../../../components/Filters';
 import { useHistory } from '../../../hooks/useHistory';
+import { useAnalytics } from '../../../hooks/useAnalytics';
+import { FULL_TEXT_SLUG } from '../../../components/Filters/SchemaSieve';
 
 interface FiltersProps<T> {
 	schema: JSONSchema;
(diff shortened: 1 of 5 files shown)


src/extra/AutoUi/Collection/Filters.tsx
```diff
@@ -24,6 +26,47 @@ export const Filters = <T extends AutoUIBaseResource<T>>({
 	renderMode,
 }: FiltersProps<T>) => {
 	const history = useHistory();
+	const { analyticsClient } = useAnalytics();
+
+	React.useEffect(() => {
+		if (!analyticsClient || !filters?.length) {
```
gelbal, line 32: Probably it'd be useful to add a similar heuristic like @laziob added on the number of characters in the search box. So we don't send an event for each single character typed in the search bar.

src/extra/AutoUi/Collection/Filters.tsx
```diff
@@ -24,6 +26,47 @@ export const Filters = <T extends AutoUIBaseResource<T>>({
...
+					if (filterObj) {
+						analyticsObj.field.push(filterObj.name);
+						analyticsObj.operator.push(filterObj.operator);
+						analyticsObj.value.push(filterObj.value);
+					}
+				});
+			}
+		});
+
+		analyticsClient.track('Filters Action', analyticsObj);
```
gelbal, line 67: Ah I realized that only the web tracker has this `prefix` defined in the initialization. See [the module's code](https://github.com/balena-io-modules/analytics-client/blob/23064da75027d9013cc34fc8364f2577cfe564f5/src/web.ts#L12) and how [Hub does it now](https://github.com/balena-io/balenahub/blob/cad538072b330492e949e6c287dc30460ea6e577/src/api/analytics.ts#L29). When we use the regular method `client.track` (not web tracker one), we prepend the prefix ourselves. So this becomes:
```js
analyticsClient.track('[Hub] Filters Action', analyticsObj);
```

If we'd like to keep prepending the component name, we need to make it a configuration option. So the regular client also prepends such prefix to each event.

gelbal: I like the prefix method and I'd like to keep it. What do you think @laziob?

gelbal (commented):
Thanks again for the session yesterday Andrea. I left couple of comments elaborating what I said at the end of our session.

src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,28 @@
...
+			!analytics?.componentName ||
+			!analytics?.endpoint ||
+			!analytics?.projectName
+		) {
+			return {};
+		}
+		const urlParamsHandler = new AnalyticsUrlParams();
+		const newQuery =
+			urlParamsHandler.consumeUrlParameters(window.location.search) ?? null;
+		const passedDeviceId = urlParamsHandler.getPassedDeviceId();
```
thgreasi, line 18: Maybe, instead of having the deviceId param defined here, the analytics context could have an extra optional callback prop for retrieving the extra props to pass to the `createClient` call.
Eg:
```ts
export interface Analytics {
	endpoint: string;
	projectName: string;
	componentName: string;
	componentVersion?: string;
	initializationExtra?: (urlParamsHandler: AnalyticsUrlParams) => AnyObject
}
```
where we would then provide something like:
```
initializationParameterTransformer: (urlParamsHandler) => ({
    passedDeviceId: urlParamsHandler.getPassedDeviceId(),
})
```

JSReds:
closing this for inactivity

PR #1330, opened 2021-07-12
Form: Support using null values in the widgets when it's valid type

thgreasi:
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Renderer/__snapshots__/spec.tsx.snap
+++ b/src/components/Renderer/__snapshots__/spec.tsx.snap
@@ -1160,6 +1160,43 @@ exports[`Renderer component A default value 1`] = `
   -webkit-font-smoothing: antialiased;
 }
 
+.c2 {
+  box-sizing: border-box;
+}
+
+.c3 {
+  margin-bottom: 4px;
+  min-width: 0;
+  min-height: 0;
+}
+
+.c5 {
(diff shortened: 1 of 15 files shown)


grahammcculloch:
@thgreasi once it's building, I'd like to test the beta release of this in Jellyfish. IIRC we rely on `ui:value: null` to explicitly mean 'do not render this field at all'. We might need to rethink this but in any case I'm worried this change might result in us showing a whole load more (empty) fields in Jellyfish.

thgreasi:
@grahammcculloch that sounds a bit inconsistent with what rjsf is doing for null values, isn't it?
See: https://rjsf-team.github.io/react-jsonschema-form/

The point of this PR was to support implementing a widget which is able to have its own handling for `null` values (eg show a special "unset" label with special formatting or something (something that default/emptyValue doesn't offer).
For null-able properties, that's just another valid value which we do need to render.

grahammcculloch:
@thgreasi please ignore all my comments yesterday! I was remembering completely wrongly how we do things in Jellyfish. So I'm happy with these changes!

grahammcculloch (approved):
LGTM - we have an easy way in Jellyfish to deal with extra null value fields showing up, using the ui-schema.

PR #1341, opened 2021-07-16
Replace custom filter edit

JSReds:
Replace custom filter Edit with Form

Change-type: minor

work in progress to accept array types RJSForm: https://github.com/rjsf-team/react-jsonschema-form/issues/282 (partially solved)

Screenshot 2022-02-15 at 13 56 03
Screenshot 2022-02-15 at 13 56 16
Screenshot 2022-02-15 at 13 56 36
Screenshot 2022-02-15 at 13 56 43

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Filters/FilterModal.tsx
+++ b/src/components/Filters/FilterModal.tsx
@@ -7,13 +7,13 @@ import React, { useState } from 'react';
 import styled from 'styled-components';
 import { Button } from '../Button';
 import { Box } from '../Box';
-import { getDataModel } from '../DataTypes';
 import { Flex } from '../Flex';
 import { Modal } from '../Modal';
 import { Select } from '../Select';
 import { Txt } from '../Txt';
 import * as SchemaSieve from './SchemaSieve';
-
+import { Form } from '../Form';
+import { getDataModel } from '../DataTypes';
 export interface FilterFieldOption {
 	field: string;
 	title: string;
@@ -31,14 +31,12 @@ export interface FilterModalProps {
 	edit: EditModel[];
 	fieldCompareFn?: FilterFieldCompareFn;
 }
-
 export interface FilterInputProps {
 	schema: JSONSchema;
 	value: any;
 	operator: string;
 	onUpdate: (value: any) => void;
 }
-
 export interface EditModel {
 	field: string;
 	operator: string;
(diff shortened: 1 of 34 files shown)


LucianBuzzo:
This is a great change! It looks like there are a few issues where the `Form` component doesn't work with values that the previous edit types could handle. I think `type: ['string', null]` is an example of this. We can use this PR as a forcing function to make the `Form` component more capable.

src/components/Filters/FilterModal.tsx
```diff
@@ -50,22 +48,27 @@ const defaultFilterCompareFn: FilterFieldCompareFn = (a, b) => {
 };
 
 const FilterInput = (props: FilterInputProps) => {
+	const { schema, value, onUpdate } = props;
 	const model = getDataModel(props.schema);
-
-	if (!model) {
-		return null;
```
thgreasi, line 56: Maybe `if (!model) return null;` is fine to keep?
My concern is whether the filters will be able to work if we find a type for which we don't have a DataModel for, even though the Form will be able to render it.

src/components/Filters/FilterModal.tsx
```diff
@@ -50,22 +48,27 @@ const defaultFilterCompareFn: FilterFieldCompareFn = (a, b) => {
 };
 
 const FilterInput = (props: FilterInputProps) => {
+	const { schema, value, onUpdate } = props;
```
thgreasi, line 51: I guess we can inline the destructuring and stop using `props.X` completely :)
```suggestion
const FilterInput = ({ schema, value, onUpdate }: FilterInputProps) => {
```

thgreasi (commented):
This looks like magic :tada:

src/components/Filters/FilterModal.tsx
JSReds: if model is not defined we should try to render the form in any case no ? The model is not needed to render the Form component, right ?

src/components/Filters/FilterModal.tsx
```diff
@@ -197,14 +200,14 @@ export const FilterModal = ({
 									/>
 								</Box>
 							)}
-							<Box flex={1}>
+							<Flex flex={1} className="stocazzo" alignItems="center">
```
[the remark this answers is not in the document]
JSReds: 🤦

thgreasi (approved):
That's just magic :sparkles:  :tada: :+1:

PR #1346, opened 2021-07-21
Add Focus Searching to AutoUICollection and add Subselection Checkbox

myarmolinsky:
Add Focus Searching to AutoUICollection and add Subselection Checkbox

Change-type: major

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Filters/SchemaSieve.ts
+++ b/src/components/Filters/SchemaSieve.ts
@@ -4,7 +4,6 @@ import { JSONSchema7 as JSONSchema } from 'json-schema';
 import cloneDeep from 'lodash/cloneDeep';
 import defaults from 'lodash/defaults';
 import every from 'lodash/every';
-import findIndex from 'lodash/findIndex';
 import findKey from 'lodash/findKey';
 import includes from 'lodash/includes';
 import map from 'lodash/map';
@@ -112,21 +111,14 @@ export const createFullTextSearchFilter = (
 	return filter as JSONSchema;
 };
 
-// Update or insert a full text search filter into an array of filters
-export const upsertFullTextSearch = (
+// Insert a full text search filter into an array of filters
+export const insertFullTextSearch = (
 	schema: JSONSchema,
 	filters: JSONSchema[] = [],
 	term: string,
 ) => {
 	const searchFilter = createFullTextSearchFilter(schema, term);
-	const existingSearchIndex = findIndex(filters, { title: FULL_TEXT_SLUG });
-
-	if (existingSearchIndex > -1) {
-		return filters.map((filter, i) =>
-			existingSearchIndex === i ? searchFilter : filter,
-		);
-	}
-	return [...filters, searchFilter as JSONSchema];
+	return filters.concat(searchFilter);
 };
 
 // Removes a full text search filter from an array of filters
--- a/src/components/Table/TableBase.tsx
+++ b/src/components/Table/TableBase.tsx
@@ -11,7 +11,6 @@ import some from 'lodash/some';
 import sortBy from 'lodash/sortBy';
 import * as React from 'react';
 import styled from 'styled-components';
-
 import { Button } from '../Button';
 
 // TODO: Remove explicit import and depend on provider instead.
@@ -21,7 +20,6 @@ import { CheckboxProps, Checkbox } from '../Checkbox';
 import { Pager } from '../Pager';
 import { CheckboxWrapper, TableBaseColumn, TableRow } from './TableRow';
 import filter from 'lodash/filter';
-import every from 'lodash/every';
 
 const highlightStyle = `
 	background-color: ${theme.colors.info.light};
@@ -180,25 +178,22 @@ export class TableBase<T> extends React.Component<
 	}
 
 	public componentDidUpdate(prevProps: TableBaseProps<T>) {
-		if (this.props.sort && !isEqual(prevProps.sort, this.props.sort)) {
+		const { sort, checkedItems, data, itemsPerPage } = this.props;
+		if (sort && !isEqual(prevProps.sort, sort)) {
 			this.setState({
-				sort: this.props.sort,
+				sort,
 			});
 		}
 
-		if (
-			(this.props.checkedItems &&
-				prevProps.checkedItems !== this.props.checkedItems) ||
-			(this.props.data && prevProps.data !== this.props.data)
-		) {
-			if (this.props.checkedItems) {
-				this.setRowSelection(this.props.checkedItems);
-			}
+		if (checkedItems && prevProps.checkedItems !== checkedItems) {
+			this.setRowSelection(checkedItems);
 		}
 
-		const totalItems = this.props.data?.length ?? 0;
-		const itemsPerPage = this.props.itemsPerPage ?? 50;
-		if (this.state.page !== 0 && totalItems <= this.state.page * itemsPerPage) {
+		const totalItems = data?.length ?? 0;
+		if (
+			this.state.page !== 0 &&
+			totalItems <= this.state.page * (itemsPerPage ?? 50)
+		) {
 			this.resetPager();
 		}
 	}
@@ -244,19 +239,15 @@ export class TableBase<T> extends React.Component<
 		return includes(this.props.disabledRows, identifier);
 	}
 
-	public isEachRowChecked(checkedItems: T[]): CheckedTypes {
+	public howManyRowsChecked(checkedItems: T[]): CheckedTypes {
 		const { rowKey, data } = this.props;
 		if (!rowKey || !data) {
 			return 'none';
 		}
 
-		const allDataCount: number = this.props.allDataCount || data.length;
-		const selectedKeys = map(checkedItems, rowKey);
-
-		return every(data, (x) => includes(selectedKeys, x[rowKey])) &&
-			allDataCount === checkedItems.length
+		return data.length === checkedItems.length
 			? 'all'
-			: some(data, (x) => includes(selectedKeys, x[rowKey]))
+			: checkedItems.length > 0
 			? 'some'
 			: 'none';
 	}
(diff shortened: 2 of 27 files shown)


src/extra/AutoUi/Collection/index.tsx
```diff
@@ -231,7 +233,13 @@ export const AutoUICollection = <T extends AutoUIBaseResource<T>>({
...
 
 	React.useEffect(() => {
-		setSelected((oldSelected) => getSelectedItems(oldSelected, filtered));
+		setFilteredSelected(getSelectedItems(selected, filtered));
+	}, [selected, filtered]);
+
+	React.useEffect(() => {
+		setFilteredSelected((oldFilteredSelected) =>
+			getSelectedItems(oldFilteredSelected, filtered),
+		);
```
JSReds, line 242: Since we don't want to change the selection while filtering, the entire useEffect can be  removed.

src/extra/AutoUi/Collection/index.tsx
```diff
@@ -266,20 +274,23 @@ export const AutoUICollection = <T extends AutoUIBaseResource<T>>({
...
 
-	const onActionDone = React.useCallback((isSuccessful: boolean) => {
-		if (isSuccessful) {
-			setSelected([]);
-		}
+	const onActionDone = React.useCallback(
+		(isSuccessful: boolean) => {
+			if (isSuccessful) {
+				setSelected(difference(selected, filteredSelected));
+			}
```
JSReds, line 288: If we approve to keep the selection after the action, we can just remove this piece of code.

JSReds (commented):
I would recommend doing this change adding one line of code and removing the commented code. This would avoid adding useState, useEffect and another lodash dependency:

line 374: `affectedEntries: !!actionData.affectedEntries?.length ? filter(filters, actionData.affectedEntries) as T[] : []`

src/utils/index.ts
```diff
@@ -51,7 +51,7 @@ export const withConditional = <TProps extends {}>(
 	};
 };
 
-export const stopEvent = (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+export const stopEvent = (event: any) => {
```
thgreasi, line 54: Can't we do anything better?

myarmolinsky: Don't even remember why I made this change, it's not necessary. Reverting

src/components/Table/TableBase.tsx
```diff
@@ -310,18 +334,20 @@ export class TableBase<T> extends React.Component<
...
-		const checkedItems = allChecked
-			? (data || []).slice().filter((r) => !this.isDisabled(r))
-			: [];
+		const { data, onCheck } = this.props;
 
-		if (this.props.onCheck) {
-			this.props.onCheck(checkedItems);
+		const checkedItems =
+			this.state.checkedItems.length === 0
+				? (data || []).slice().filter((r) => !this.isDisabled(r))
```
JSReds, line 341: filter already create a new array, why do you need slice here ?

src/extra/AutoUi/Collection/List.tsx
```diff
@@ -189,6 +189,7 @@ export const List = <T extends AutoUIBaseResource<T>>({
 				<Table<T>
 					rowKey={'id'}
 					data={filtered}
+					allDataCount={data?.length || 0}
```
JSReds, line 192:
```suggestion
					allDataCount={data?.length ?? 0}
```

src/extra/AutoUi/Collection/index.tsx
```diff
@@ -68,17 +76,30 @@ const HeaderGrid = styled(Flex)`
...
-		return selectedItems;
+const Focus = styled(Box)`
+	flex-basis: 500px;
+	background-color: white;
+	border: solid 1px ${(props) => props.theme.colors.quartenary.dark};
+	max-height: 200px;
+	position: absolute;
+	width: 100%;
+	z-index: 1;
+	border-radius: 0 0 5px 5px;
```
JSReds, line 87: 0 0 props.theme.global.drop.border.radius props.theme.global.drop.border.radius long but at least we will use the same border radius for every drop

src/extra/AutoUi/Collection/index.tsx
```diff
@@ -320,6 +337,111 @@ export const AutoUICollection = <T extends AutoUIBaseResource<T>>({
 									filters={filters}
 									autouiContext={autouiContext}
 									changeFilters={setFilters}
+									onSearch={(searchTerm) => {
```
JSReds, line 340: Let's make another component that accept searchTerm as prop. Just to make it more readable

src/components/Filters/SchemaSieve.ts
```diff
@@ -112,21 +111,15 @@ export const createFullTextSearchFilter = (
...
 	term: string,
 ) => {
 	const searchFilter = createFullTextSearchFilter(schema, term);
-	const existingSearchIndex = findIndex(filters, { title: FULL_TEXT_SLUG });
 
-	if (existingSearchIndex > -1) {
-		return filters.map((filter, i) =>
-			existingSearchIndex === i ? searchFilter : filter,
-		);
-	}
```
JSReds, line 128: By removing this piece, we will no longer be able to override a search made with the search bar as they will be added in &. This removes the override behaviour of the previous filter, is this something we really want?

src/components/Table/TableBase.tsx
```diff
@@ -184,10 +186,13 @@ export class TableBase<T> extends React.Component<
...
-			this.props.checkedItems &&
-			prevProps.checkedItems !== this.props.checkedItems
+			(this.props.checkedItems &&
+				prevProps.checkedItems !== this.props.checkedItems) ||
+			(this.props.data && prevProps.data !== this.props.data)
 		) {
-			this.setRowSelection(this.props.checkedItems);
+			if (this.props.checkedItems) {
+				this.setRowSelection(this.props.checkedItems);
+			}
```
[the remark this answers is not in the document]
myarmolinsky:
```
if (page !== undefined && prevProps.page !== page) {
			this.setState({ page });
		}
```
There's no `page` prop, not sure if you noted that from an older change I made 😅 

WRT the data, it changes when the filters change (for example adding a filter might remove some entities). With that said, I do not believe this is necessary anymore because now if the data changes (or rather the filters update, adding or removing data), we remove all selections, just like the behavior of filter application in gmail.

src/components/Table/TableBase.tsx
```diff
@@ -238,15 +243,21 @@ export class TableBase<T> extends React.Component<
...
+		if (!rowKey || !data) {
+			return 'none';
 		}
 
+		const allDataCount: number = this.props.allDataCount || data.length;
 		const selectedKeys = map(checkedItems, rowKey);
 
-		return every(this.props.data, (x) => includes(selectedKeys, x[rowKey]));
+		return every(data, (x) => includes(selectedKeys, x[rowKey])) &&
+			allDataCount === checkedItems.length
```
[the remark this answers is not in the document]
myarmolinsky: I think you're right that we don't need that `every` check. This also brought to my attention that we no longer need to compare against `allDataCount` in general given that we reset selections when filtering, so selections are only ever based on filtered data, meaning we can show an `all` checkbox if all filtered items are selected

src/components/Table/TableBase.tsx
```diff
@@ -286,20 +297,33 @@ export class TableBase<T> extends React.Component<
...
+		let checkedItems: T[] = [];
+		let allChecked: CheckedTypes = 'none';
+
+		if (!data) {
+			return { checkedItems, allChecked };
+		}
+
+		const allDataCount: number = this.props.allDataCount
+			? this.props.allDataCount
+			: data.length;
```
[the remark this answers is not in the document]
myarmolinsky: We actually no longer need this prop so I've removed it

src/components/Table/TableBase.tsx
```diff
@@ -286,20 +297,33 @@ export class TableBase<T> extends React.Component<
...
 		if (data) {
-			checkedItems = filter(this.props.data, (x) =>
-				includes(selectedRowsIds, x[rowKey]),
-			);
-			allChecked = data.length > 0 && checkedItems.length === data.length;
+			checkedItems = filter(data, (x) => {
+				return includes(selectedRowsIds, x[rowKey]);
+			});
+			allChecked =
+				allDataCount > 0 && checkedItems.length === allDataCount
```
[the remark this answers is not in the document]
myarmolinsky: Removed `allDataCount`, using data.length instead and thus it is not necessary here

src/components/Table/TableBase.tsx
```diff
@@ -286,20 +297,33 @@ export class TableBase<T> extends React.Component<
...
+					? 'all'
+					: checkedItems.length < allDataCount
```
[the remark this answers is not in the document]
myarmolinsky: Agreed, using that logic instead

src/extra/AutoUi/Collection/List.tsx
myarmolinsky: Removed that property

src/extra/AutoUi/Collection/index.tsx
myarmolinsky: Didn't know we had this, thanks!!

src/components/Filters/SchemaSieve.ts
myarmolinsky: Well the overriding behavior of the previous filter is one of the things that created the illusion that the search bar is a *focus search*. With these changes, we will no longer replace the old filter but rather add a new one when the user hits enter (classic search behavior). If the user does not hit enter, they are searching through their filtered list and getting a preview of their search results in the focus search dropdown.

If what you are saying is that hitting enter should replace the old entry you made by hitting enter, I believe this was a point we considered in FD. See the following thread: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/A2X-xEh5uPoGVTR-h7a8p3pb71X
Me: "Should pressing enter to add the term to the filter replace the current fulltext filter? Or should it add another filter? (for example I press enter on 'rf'. If I press enter on 'l', in its current state the 'l' filter replaces the 'rf' filter, but I'm thinking it should now have both filters instead. Not 100% sure which is why I'm checking)"
Chris: "I think it should add a second filter; each press of enter should keep adding filters"

JSReds: I'm lost when you say `classic search behavior`, I can't think of any site that currently does this type of search with a search bar (adding N filters in AND). Could you give me an example ?

Screenshot 2021-11-15 at 16 33 45

IMHO something like this can be really confusing

myarmolinsky: I hear your argument. I could go either way I suppose. I saw it as adding a fulltext filter rather than submitting a *new* search. Perhaps that's the wrong way to look at it. What do you think @chrisys ?

myarmolinsky: @JSReds If you want I can revert to upsert for this PR and then we can switch to insert in a separate followup PR if we decide that's really the functionality we want

myarmolinsky: Sounds good to me 👍

myarmolinsky: Readding insert as per our meeting discussion. Will determine whether to switch back to upsert based on user feedback

myarmolinsky:
@balena-ci I self certify!

PR #1351, opened 2021-07-26
Fix DownloadImageModal alignment when there are not multiple compatible device types

myarmolinsky:
Fix DownloadImageModal alignment when there are not multiple compatible device types

See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/6s0dQE2ma902xeyIg3Fbg1dJxei
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/unstable-temp/DownloadImageModal/OsConfiguration.tsx
+++ b/src/unstable-temp/DownloadImageModal/OsConfiguration.tsx
@@ -240,9 +240,9 @@ export const OsConfiguration = ({
 
 	return (
 		<>
-			<Flex mb={3}>
+			<Flex mb={3} mx={-2}>
 				{compatibleDeviceTypes.length > 1 && (
-					<Box flex={3} mr={2}>
+					<Box flex={3} mx={2}>
 						<DownloadImageLabel>
 							{t('placeholders.select_device_type')}
 							<QuestionMark
@@ -259,7 +259,7 @@ export const OsConfiguration = ({
 					</Box>
 				)}
 				{(!isInitialDefault || !selectedOsType) && hasEsrVersions && (
-					<Box flex={2} ml={2}>
+					<Box flex={2} mx={2}>
 						<DownloadImageLabel>
 							{t('placeholders.select_os_type_status')}{' '}
 							<DocsLink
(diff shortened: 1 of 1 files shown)


src/unstable-temp/DownloadImageModal/OsConfiguration.tsx
```diff
@@ -259,7 +261,7 @@ export const OsConfiguration = ({
 					</Box>
 				)}
 				{(!isInitialDefault || !selectedOsType) && hasEsrVersions && (
-					<Box flex={2} ml={2}>
+					<Box flex={2} ml={hasMultipleCompatibleDeviceTypes ? 2 : undefined}>
```
thgreasi, line 264: I would really prefer to reduce the conditionals. An extra pointer of how badly this scales is that when the OsTypeSelector is hidden, we end up with an orphan `mr={2}` in DeviceTypeSelector.
On top of that, this component will be converted to a purely data driven form, in which case we will not even be able to do this kind of conditionals for formatting.
Going with negative margins or even better using css grid sounds what we will need to do later anyway.

PS:We already have a DataGrid, which might fit here as well, but it needs to be able accept the margin as a parameter (and preserve the current value as a default).

myarmolinsky: Ah I see, I wasn't aware that the conditionals had this much of an effect. Thank you for explaining, I'll reduce the conditionals 👍

src/unstable-temp/DownloadImageModal/OsConfiguration.tsx
```diff
@@ -240,9 +240,9 @@ export const OsConfiguration = ({
 
 	return (
 		<>
-			<Flex mb={3}>
+			<Flex mb={3} ml={-2} mr={-2}>
 				{compatibleDeviceTypes.length > 1 && (
-					<Box flex={3} mr={2}>
+					<Box flex={3} ml={2} mr={2}>
```
thgreasi, line 245: You can combine all `ml={2} mr={2}` pairs into v :)
```suggestion
					<Box flex={3} mx={2}>
```

thgreasi (approved):

src/unstable-temp/DownloadImageModal/OsConfiguration.tsx
myarmolinsky: True 👍

myarmolinsky:
@balena-ci rebase

PR #1358, opened 2021-07-29
Add Renovate config

joshbwlng:
Change-type: patch
Signed-off-by: Josh Bowling <josh@balena.io>

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,16 @@
+{
+  "extends": [
+    "config:base"
+  ],
+  "enabledManagers": [
+    "npm"
+  ],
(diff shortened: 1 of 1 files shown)


joshbwlng:
@JSReds Updated the default Renovate config to something a bit closer to what we probably want. Take a look and let me know what you think. The other big thing to consider is whether we want to group multiple bumps into a single PR or not. Doing so reduces the workload passed on to our CI workers, but can make it difficult to debug which package is causing test failures when more than one is a part of the PR. Have a look at our Renovate config to see what we're doing on the productOS side of things: https://github.com/product-os/jellyfish-config/blob/master/config/renovate.json#L27

Issue #1329, opened 2021-07-11
[table] Unknown event handler property `onSort`. It will be ignored.


atb00ker:
**Describe the bug**

While creating a new table, I am seeing `Unknown event handler property onSort. It will be ignored.` in console, even when I am not using `onSort`

**To Reproduce**

1. Create a new create component and the following:
```
<Box my={5} mx={['auto', 15]}>
  <Table
    columns={[
      { field: 'id', label: "ID" }
    ]}
    data={<_YOUR_DATA_>}
  />
</Box>
```
2. Open the console on the browser and load the component.
3. Check the console.logs, it should contain:
```
Warning: Unknown event handler property `onSort`. It will be ignored.
    in div (created by styled.div)
    in styled.div (created by TableBase)
    in div (created by styled.div)
    in styled.div (created by TableBase)
    in TableBase (created by Styled(TableBase))
    in Styled(TableBase) (created by Table)
    in div (created by styled.div)
    in styled.div (created by ForwardRef)
    in ForwardRef (created by Component)
    in Component (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in ForwardRef (created by Box)
    in Box (created by Flex)
    in Flex (created by Table)
    in div (created by styled.div)
    in styled.div (created by ForwardRef)
    in ForwardRef (created by Component)
    in Component (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in ForwardRef (created by Box)
    in Box (created by Styled(Flex))
    in Styled(Flex) (created by Table)
    in Table (created by Home)
    in div (created by styled.div)
    in styled.div (created by ForwardRef)
    in ForwardRef (created by Component)
    in Component (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in ForwardRef (created by Box)
    in Box (created by Home)
    in ErrorBoundary (created by Home)
    in Home (created by Context.Consumer)
    in Route (created by App)
    in Switch (created by App)
    in Router (created by HashRouter)
    in HashRouter (created by App)
    in div (created by StyledGrommet)
    in StyledGrommet (created by Grommet)
    in Grommet (created by Styled(Grommet))
    in Styled(Grommet) (created by BaseProvider)
    in BreakpointProvider (created by BaseProvider)
    in BaseProvider (created by App)
    in App
```

**Expected behavior**

No error should be visible.

**Specifications (please complete the following information):**
  - Browser: Mozilla Firefox 78.11.0esr
[...]

PR #1363, opened 2021-08-13
Fix unknown event handler onSort in TableBase

myarmolinsky:
Fix unknown event handler onSort in TableBase

Resolves: #1329 
Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Table/TableBase.tsx
+++ b/src/components/Table/TableBase.tsx
@@ -450,7 +450,7 @@ export class TableBase<T> extends React.Component<
 			onRowClick,
 			getRowHref,
 			getRowClass,
-			...props
+			className,
 		} = this.props;
 
 		const { page, sort } = this.state;
@@ -485,7 +485,7 @@ export class TableBase<T> extends React.Component<
 
 				<BaseTableWrapper>
 					<Base
-						{...props}
+						className={className}
 						hasRowClick={!!onRowClick}
 						hasGetRowRef={!!getRowHref}
 						hasCheckbox={!!onCheck}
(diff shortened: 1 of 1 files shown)


src/components/Table/TableBase.tsx
```diff
@@ -450,6 +450,7 @@ export class TableBase<T> extends React.Component<
 			onRowClick,
 			getRowHref,
 			getRowClass,
+			onSort,
```
JSReds, line 453: Let's not destructure an object property that we don't use. To solve this problem, I would just remove the `{...props}` from the Base component, since it only accept the `InternalTableBaseProps` a these are already passed to the component. This will probably also allow you to also remove the `...props`  right below 

```
interface InternalTableBaseProps {
	hasCheckbox: boolean;
	hasRowClick: boolean;
	hasGetRowRef: boolean;
}
```

myarmolinsky:
@balena-ci retest

thgreasi (approved):
Assuming that we are not assigning any crazy property that's outside what's in the typings, that looks correct.
Let's just double check in the dashboard that everything looks the same when compared with prod. I hope we do not rely that object spread to passthrough any props like `style` & `className` (which are not in the typings).

You probably need to do a fresh `npm i` and update the snapshots, which cause the tests to fail.

src/components/Table/TableBase.tsx
```diff
@@ -485,7 +484,6 @@ export class TableBase<T> extends React.Component<
 
 				<BaseTableWrapper>
 					<Base
-						{...props}
```
thgreasi, line 488: Actually, I'm pretty sure that we need to wire in the className property, since we are using `styled(TableBase)` in `~src/components/Table/index.tsx` :thinking:

myarmolinsky: Ah you're right, I'm seeing differences in styling between prod and local

thgreasi (commented):
LGTM

myarmolinsky:
@balena-ci I self certify!

PR #1393, opened 2021-09-07
Link headings in markdown

iamsolankiamit:
Change-type: minor


LucianBuzzo:
@iamsolankiamit Does this plugin create anchor links for headings? Is there an instance where we would not want this behaviour? It seems really useful.

iamsolankiamit:
@LucianBuzzo made it optional, since i don't know how and where this might affect in jellyfish. If you think it won't affect (like adding anchor tags and ids in chat maybe?) any of those projects, then i'll flip the condition to optionally disable adding links and ids to headings.

LucianBuzzo:
@iamsolankiamit I think we should make it the standard, and remove the option. If we have a really strong reason we can always make it optional later. We should push back quite hard on adding new options, as it leads to a combinatorial explosion of potential outputs, and also potential errors.

iamsolankiamit:
@balena-ci I self-certify!

PR #1400, opened 2021-09-14
Add actionFn property to AutoUICollection

myarmolinsky:
Add actionFn property to AutoUICollection

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/extra/AutoUi/schemaOps.ts
+++ b/src/extra/AutoUi/schemaOps.ts
@@ -38,11 +38,12 @@ export interface AutoUIModel<T> {
 export interface AutoUIAction<T> {
 	title: string;
 	type: 'create' | 'update' | 'delete';
-	renderer: (props: {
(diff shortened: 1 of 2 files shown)


src/extra/AutoUi/schemaOps.ts
```diff
@@ -43,6 +43,7 @@ export interface AutoUIAction<T> {
 		affectedEntries?: T[];
 		onDone: (isSuccessful: boolean) => void;
 	}) => React.ReactNode;
+	actionFn?: (props: { affectedEntries?: T[] }) => void;
```
JSReds, line 46: I like it! I would just also make the `renderer` function to not be required, so we can avoid having renderer: null, actionFn: () => ...

JSReds (approved):
LGTM

PR #1404, opened 2021-09-23
Remove deprecated recompose package

LucianBuzzo:
Recompose is no longer being actively maintained and since we are using
a very small subset of its API its a simple thing to replace.
The `getDisplayName()` method is trivial and only used in a single
place.
The `compose()` method is functionally the same as `lodash.flowRight()`
so can be swapped out.
This leaves `withProps()` which again is relatively straightforward.
I've added a new utils function that handles the props wrapping this
method previously provided.

Connects-to: #1361
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


LucianBuzzo:
@balena-ci retest

LucianBuzzo:
@balena-ci retest

src/asRendition.tsx
```diff
@@ -118,7 +131,7 @@ export default function asRendition<T>(
 	> = [],
 	passthroughProps: string[] = [],
 ) {
-	return compose(
+	return flowRight(
```
thgreasi, line 134: I think recompose was doing it in order, in which case we should go with `flow` right? (fwiw)
See: https://github.com/acdlite/recompose/blob/3db12ce7121a050b533476958ff3d66ded1c4bb8/src/packages/recompose/compose.js#L2

Maybe we could just copy that ^ in out utils as is, so that is has a better name & we don't add lodash imports since it definitely has way more code under the hood...
See: https://github.com/lodash/lodash/blob/67389a8c78975d97505fa15aa79bec6397749807/lodash.js#L5091

src/components/Navbar/index.tsx
```diff
@@ -125,7 +126,7 @@ export interface NavbarProps extends React.HTMLAttributes<HTMLElement> {
  *
  * [View story source](https://github.com/balena-io-modules/rendition/blob/master/src/components/Navbar/Navbar.stories.tsx)
  */
-export const Navbar = compose(setDefaultProps)(
+export const Navbar = flowRight(setDefaultProps)(
```
thgreasi, line 129: As above I think this should be
```suggestion
export const Navbar = flow(setDefaultProps)(
```

src/utils/reactUtils.tsx
```diff
@@ -0,0 +1,37 @@
...
+			return <BaseComponent {...combinedProps} />;
+		};
+
+		return hoc;
+	};
+
+	return wrapper as any;
+}
+
+export default withProps;
```
thgreasi, line 37: I think that's not used anywhere right?

src/utils/reactUtils.tsx
```diff
@@ -0,0 +1,37 @@
...
+	fn: TInner | mapper<TOutter, TInner>,
+): InferableComponentEnhancerWithProps<TInner & TOutter, TOutter> {
+	const wrapper = (BaseComponent: any) => {
+		const hoc = (props = {}) => {
+			const combinedProps = {
+				...props,
+				...(typeof fn === 'function' ? (fn as any)(props) : fn),
+			};
+
+			return <BaseComponent {...combinedProps} />;
```
thgreasi, line 28: That removed an unnecessary destructuring I think :thinking: 
```suggestion
			return <BaseComponent
				{...props}
				{...(typeof fn === 'function' ? (fn as any)(props) : fn)}
			/>;
```

src/asRendition.tsx
```diff
@@ -18,6 +18,19 @@ import {
 import { StyledSystemProps } from './common-types';
 import { Tooltips } from './tooltips';
 
+// Returns the display name of a React component. Falls back to 'Component'.
+export const getDisplayName = (component: React.ComponentType): string => {
```
thgreasi, line 22: I guess that could also be in `reactUtils` right?
I mean, asRendition is already one our most complex files...

thgreasi (commented):
It feel like magic how easy it was to drop that dependency... :clap:

src/asRendition.tsx
LucianBuzzo: Confusing, because their docs say it's the same as lodash's `flowRight` 😂  https://github.com/acdlite/recompose/blob/3db12ce7121a050b533476958ff3d66ded1c4bb8/docs/API.md#compose

src/asRendition.tsx
LucianBuzzo: @thgreasi It's not used anywhere else, so is more likely to rot if it gets moved into a seperate file

src/utils/reactUtils.tsx
LucianBuzzo: Good call, I'll change this

src/asRendition.tsx
thgreasi: Oh my... Another clue that ppl get confused by `.reduce`... Even them got confused by it :disappointed:

LucianBuzzo: @thgreasi I'll just copy the recompose function

src/utils/reactUtils.tsx
LucianBuzzo: Yep, this can get removed.

LucianBuzzo:
@thgreasi I resolved your comments, can you re-review please?

thgreasi (approved):

PR #1408, opened 2021-09-24
Remove new browser tab from download image process

JSReds:
Remove new browser tab from download image process

Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/6pSNP96tIs2hwDa5s_0CvBI0nQm

Screenshot 2021-09-24 at 16 18 18

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/unstable-temp/DownloadImageModal/ImageForm.tsx
+++ b/src/unstable-temp/DownloadImageModal/ImageForm.tsx
@@ -14,6 +14,11 @@ import { DownloadFormModel, FormModel } from './FormModel';
 import { DeviceType } from './models';
 import { DownloadOptions } from './DownloadImageModal';
 
+export enum DownloadTypeEnum {
+	config,
+	image,
+}
+
 const debounceDownloadSize = debounce(
 	(getDownloadSize, deviceType, rawVersion, setDownloadSize) =>
 		getDownloadSize(deviceType.slug, rawVersion)
@@ -72,7 +77,7 @@ interface ImageFormProps {
 		downloadConfigOnly: boolean,
 		downloadOptions: DownloadOptions,
 	) => void;
-	setIsDownloadingConfig: (isDownloading: boolean) => void;
+	setIsDownloadingType: (isDownloading: DownloadTypeEnum | null) => void;
 	downloadConfig?: (model: FormModel) => Promise<void> | undefined;
 	getDownloadSize?: () => Promise<string> | undefined;
 	modalActions?: Array<
@@ -91,7 +96,7 @@ export const ImageForm = ({
 	deviceType,
 	authToken,
 	onDownloadStart,
-	setIsDownloadingConfig,
+	setIsDownloadingType,
 	downloadConfig,
 	getDownloadSize,
 	modalActions,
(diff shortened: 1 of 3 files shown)


src/unstable-temp/DownloadImageModal/ImageForm.tsx
```diff
@@ -237,11 +261,19 @@ export const ImageForm = ({
...
+						onClick={async (event) => {
+							event.preventDefault();
+							event.stopPropagation();
 							if (model.downloadConfigOnly && downloadConfig) {
-								setIsDownloadingConfig(true);
+								setIsDownloadingType(DownloadTypeEnum.config);
 								await downloadConfig(model);
-								setIsDownloadingConfig(false);
+								setIsDownloadingType(null);
+								setDownloadConfigOnly(false);
```
myarmolinsky, line 271: It feels redundant to have both setIsDownloadingType and setDownloadConfigOnly. The latter already tells you which type we're downloading, so it sounds like the former should just be a simple boolean which gets set to true while downloading and to false when done

src/unstable-temp/DownloadImageModal/ImageForm.tsx
```diff
@@ -152,23 +157,42 @@ export const ImageForm = ({
...
+			const newBlob = new Blob([blob]);
+
+			const blobUrl = window.URL.createObjectURL(newBlob);
+
+			const link = document.createElement('a');
+			link.href = blobUrl;
+			link.setAttribute('download', `${appId}.zip`);
+			document.body.appendChild(link);
+			link.click();
+			link.parentNode?.removeChild(link);
```
myarmolinsky, line 179: Can't we say `document.body.removeChild(link)` instead of this since that's what we appended it to?

src/unstable-temp/DownloadImageModal/ImageForm.tsx
myarmolinsky: I'm sure this part is obvious but if you want to use the enum and the boolean typing of `setDownloadConfigOnly` isn't working for you, we can make this state of the type of the enum's options and set it according where necessary. Something like this:
```
const [downloadType, setDownloadType] = React.useState<DownloadEnum as const>("image");

...

setDownloadType("image");

...

setDownloadType("config");
```

myarmolinsky: I'll also note that the current method causes the button label to reset from "Download configuration file" back to "Download OS" after clicking to download the config file

src/unstable-temp/DownloadImageModal/ImageForm.tsx
```diff
@@ -262,7 +294,6 @@ export const ImageForm = ({
 									? t('warnings.image_deployed_to_docker')
 									: ''
 							}
-							onClick={() => setDownloadConfigOnly(false)}
```
myarmolinsky, line 265: Removing this onClick means clicking this option now does nothing. So if I click the config option and the button switches to config, there is no (reasonable) way for me to switch back to OS

myarmolinsky (changes requested):
Looking good, just a noted some things I think can improve and a bug caused by removing the onClick from one of the buttons

I'm also iffy on the loading effect. I get why we blur only the left half of the modal, so that we don't blur out the instructions, but it feels a bit weird to me. Maybe it's just me though, maybe we can check with product team about the best way to communicate the loading?

src/unstable-temp/DownloadImageModal/ImageForm.tsx
JSReds: This would remove the first 'a' element from the body and that's not what we want

src/unstable-temp/DownloadImageModal/ImageForm.tsx
JSReds: thank yu for noticing, this is a last minute change that does not have any sense

src/unstable-temp/DownloadImageModal/ImageForm.tsx
JSReds: thank you for noticing, this is a last minute change that does not have any sense

src/unstable-temp/DownloadImageModal/ImageForm.tsx
myarmolinsky: Ah okay

src/unstable-temp/DownloadImageModal/ImageForm.tsx
```diff
@@ -152,23 +157,42 @@ export const ImageForm = ({
...
+			const res = await fetch(url, {
+				method: 'GET',
+				headers: new Headers({
+					Authorization: 'Bearer ' + authToken,
+				}),
+			});
+			const blob = await res.blob();
+			const newBlob = new Blob([blob]);
+
+			const blobUrl = window.URL.createObjectURL(newBlob);
```
thgreasi, line 171: Gave a try on the dashboard and
```
<a href="https://api.balena-cloud.com/download?deviceType=raspberrypi3&version=2.83.21+rev1.prod" download="asdf.img">asdf</a>
```
(note the missing `target="_blank"`) opens the download prompt right away and defaults to the name that the API returns (`balena-cloud-raspberrypi3-2.83.21+rev1-v12.10.3.img`).
Peek 2021-09-28 14-05

JSReds:
Closing since we don't like the refresh page behviour

Issue #1322, opened 2021-07-06
Custom columns dropdown menu closes if the click is just below the word


JSReds:
Clicking just outside of one of the menu items the custom columns dropdown menu closes automatically. It should only close by clicking outside the dropdown.

Issue #1442, opened 2021-10-26
Add hover effect to Plain buttons in dropdown menu


jmakivic:
We should add a hover effect to plain buttons in the drop down menu in order to provide immediate feedback to users.

The need for a hover effect was first brought up here: https://github.com/balena-io/balena-ui/issues/3183

Issue #1323, opened 2021-07-06
Dropdown menu should have a scroll bar if longer then the page and not make the page scroll


JSReds:
dropdown menu should be scrollable if it contains too many things and not scroll the page.

<img width="1324" alt="Screenshot 2021-07-06 at 18 09 25" src="https://user-images.githubusercontent.com/7238159/124633172-4f74bf80-de85-11eb-97f1-911839c40b28.png">

PR #1458, opened 2021-11-10
Improve DropDownButton component

myarmolinsky:
Improve DropDownButton component

Resolves: #1322, #1323, #1442
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/DropDownButton/index.tsx
+++ b/src/components/DropDownButton/index.tsx
@@ -1,6 +1,7 @@
 import { faChevronDown } from '@fortawesome/free-solid-svg-icons/faChevronDown';
 import { faChevronUp } from '@fortawesome/free-solid-svg-icons/faChevronUp';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import flatten from 'lodash/flatten';
 import * as React from 'react';
 import styled from 'styled-components';
 import asRendition from '../../asRendition';
@@ -42,7 +43,6 @@ const MenuBase = styled(Box)<{
 }>`
 	background: white;
 	position: absolute;
-	box-shadow: ${(props) => '1px 1px 5px' + props.theme.colors.gray.light};
 	border-radius: ${(props) => px(props.theme.radius)};
 	border: ${(props) => '1px solid ' + props.theme.colors.gray.main};
 	z-index: 1;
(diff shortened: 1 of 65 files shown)


src/components/DropDownButton/index.tsx
```diff
@@ -230,25 +265,34 @@ class BaseDropDownButton extends React.Component<
...
-							}
-							if ((child as any).type === Divider) {
-								return child;
-							}
-							return (
-								<Item border={Boolean(border && i)} key={i}>
-									{child}
-								</Item>
-							);
-						})}
```
JSReds, line 251: I'm not too sure about removing the Child concept... I would keep both, as the Select component does.

myarmolinsky: The child concept is not entirely removed, it has just been moved to `item.content` @JSReds

src/components/DropDownButton/index.tsx
```diff
@@ -230,25 +265,57 @@ class BaseDropDownButton extends React.Component<
...
+											(item, j) =>
+												item && (
+													<Item
+														px={3}
+														py={1}
+														onClick={
+															!item.disabled && item.onClick
+																? item.onClick
+																: stopEvent
+														}
```
JSReds, line 283: What if we just do something like :

```suggestion
{...(!item.disabled && item.onClick && { onClick: item.onClick })}
```

so we can remove the stopEvent function, which could lead to strange behaviours

src/components/Table/TableColumnSelector.tsx
```diff
@@ -153,60 +140,60 @@ export const TableColumnSelector = function <T extends TableColumnState>(
...
-			primary
-			plain
-			label={<SmallFontAwesomeIcon icon={faCog} />}
-		>
-			{filter(props.columns, (col) => !col.locked)
-				.map((col) => (
-					<Item
+	const memoizedItems = React.useMemo(
+		(): DropdownOption[][] => [
+			filter(columns, (col) => !col.locked).map((col) => ({
```
JSReds, line 145: I know it was already like this but, let's use the normal filtrer here, no need to use lodash

src/components/DropDownButton/index.tsx
myarmolinsky: Removing the stopEvent makes it so that the dropdown closes when you click a disabled item. We want it to stay open when a disabled item is click. One thing this does is help differentiate between when clicking an action has actually done something or not.

myarmolinsky:
@balena-ci I self certify!

PR #1473, opened 2021-11-30
Add autoui app

JSReds:
Add autoUIApp to generate an app from a openApi json 

Description:

this is a starting point to generate an application from a openAPI json file

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/extra/AutoUI/utils.ts
```diff
@@ -1,5 +1,34 @@
...
+				// in case it is an object
+				result = findInObject(obj[property], key);
+
+				if (typeof result !== 'undefined') {
+					return result;
+				}
+			}
+		}
+	}
+};
```
myarmolinsky, line 23: I think we may be taking extra steps here by checking if the object has the property before checking if the property matches the key. If the property matches the key then we can just return the value, right? And *if* it does not match the key we can check if `obj[property]` is of type object (if the property does not exist on the object this check will not return true so we do not need the `hasOwnProperty` check).

```suggestion
export const findInObject = (obj: any, key: string): any => {
	for (const property in obj) {
	  if (property === key) {
	    return obj[key]; // same as obj[property]
	  } else if (typeof obj[property] === 'object') {
	    const result = findInObject(obj[property], key);
	    
	    if (typeof result !== 'undefined') {
			  return result;
		  }
	  }
	}
};
```

src/extra/AutoUIApp/Content.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+	if (!model) {
+		return <>Could not generate a model</>;
+	}
+
+	if (!(data as any[])?.length) {
+		return <>No data</>;
+	}
+
+	// TODO: since we don't have a way to render an object without a format yet, we will filter it.
```
myarmolinsky, line 83: Can you clarify what you mean by "without a format" please? Doesn't the default entity lens for AutoUI successfully render an object?

src/extra/AutoUIApp/Sidebar.tsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+import styled from 'styled-components';
+import { useLocation } from 'react-router-dom';
+import { OpenApiJson } from './openApiJson';
+import { Flex, FlexProps } from '../../components/Flex';
+import { Txt } from '../../components/Txt';
+
+const SidebarWrapper = styled(Flex)`
+	height: calc(100vh - 60px);
```
myarmolinsky, line 10: Can you please add a comment here noting that the 60px is the header height? So that it doesn't feel like we're just randomly subtracting 60px

src/extra/AutoUIApp/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { Provider } from '../..';
+import { history } from '../AutoUI/utils';
+import { Flex } from '../../components/Flex';
+import { Box } from '../../components/Box';
+import { OpenApiJson } from './openApiJson';
+
+const GlobalStyle = createGlobalStyle`
+	html,
+	body {
+		height: 100%;
```
myarmolinsky, line 16: The indents seem bigger in this file than in the others

src/extra/AutoUIApp/odata.ts
```diff
@@ -0,0 +1,39 @@
+import { Methods, OpenApiJson } from './openApiJson';
+import PineClientFetch from 'pinejs-client-fetch';
+import { JSONSchema } from '~/components/Renderer/types';
+
+export const pine = new PineClientFetch(
+	{ apiPrefix: process.env.REACT_APP_API_HOST ?? window.REACT_APP_API_HOST },
+	{ fetch },
```
myarmolinsky, line 7: Indents seem bigger in this file too

src/extra/AutoUIApp/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+				<Navbar title={title} logo={logo} />
+				<Flex>
+					<Sidebar
+						width="166px"
+						openApiJson={openApiJson}
+						isCollapsed={false}
+					/>
+					<Box
+						width="calc(100% - 166px)"
+						height="calc(100vh - 60px)"
```
myarmolinsky, line 50: A comment here about the 60px being subtracting being the header height might be nice :)

src/extra/AutoUIApp/index.tsx
myarmolinsky: Or maybe the indents of the first few files are just smaller? 🤔 I see this file (and subsequent files) are indenting with tabs, while earlier ones indent with spaces ...

src/extra/AutoUI/utils.ts
JSReds: @myarmolinsky the check is done since this is how to iterate over the enumerable properties of an object without executing on inherited properties. Also declaring the `result` inside the `for`  would redeclare the same variable N times and this is something you don't want for reasons of space and performance.

src/extra/AutoUIApp/Content.tsx
JSReds: As you can see from the line below, I'm filtering an array of data, which is represented as a Collection. What I mean with my comment is that we still cannot render an object inside a cell of a collection

src/extra/AutoUIApp/index.tsx
myarmolinsky: Seems like it may have been my github view

JSReds:
@balena-ci I self-certify!

PR #1475, opened 2021-12-02
fix broken badges

Hades32:
badge fury seems to be overwhelmed by our many versions (times out) and "david" has been replaced by "librariesio"


Hades32:
@balena-ci I self-certify!

PR #1481, opened 2021-12-07
Remove AutoUICollection and AutoUIEntity in favor of AutoUI

myarmolinsky:
Remove AutoUICollection and AutoUIEntity in favor of AutoUI

Change-type: major

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/extra/AutoUI/utils.ts
```diff
@@ -71,3 +77,41 @@ export const autoUIGetDisabledReason = <T extends AutoUIBaseResource<T>>(
...
+
+export const formatSorters: Dictionary<TableSortFunction<any>> = {};
+
+export const getSortingFunction = <T extends any>(
+	schemaKey: keyof T,
+	schemaValue: JSONSchema,
+): TableSortFunction<T> => {
+	if (formatSorters[schemaValue.format ?? '']) {
+		return formatSorters[schemaValue.format ?? ''];
+	}
```
JSReds, line 89: Like this is probably more readable 

```suggestion
	if (schemaValue.format && formatSorters[schemaValue.format]) {
		return formatSorters[schemaValue.format];
	}
```

src/extra/AutoUI/utils.ts
```diff
@@ -71,3 +77,41 @@ export const autoUIGetDisabledReason = <T extends AutoUIBaseResource<T>>(
 		});
 	}
 };
+
+export const formatSorters: Dictionary<TableSortFunction<any>> = {};
```
JSReds, line 81: This sounds wrong. That was an angular 1 service stile. What is the use case ?

src/extra/AutoUI/utils.ts
```diff
@@ -71,3 +77,41 @@ export const autoUIGetDisabledReason = <T extends AutoUIBaseResource<T>>(
...
+
+	const types = castArray(schemaValue.type);
+	if (types.includes(JsonTypes.string)) {
+		return (a: T, b: T) => {
+			const aa = (a[schemaKey] ?? '') as string;
+			const bb = (b[schemaKey] ?? '') as string;
+			if (typeof aa === 'string' && typeof bb === 'string') {
+				return aa.toLowerCase().localeCompare(bb.toLowerCase());
+			}
+			return diff(aa, bb);
```
JSReds, line 99: we cannot cast aa and bb as string and check if they are of type string right after. This sounds wrong as well. Add a check of number and boolean on the return diff otherwise return 0. Or we can add object handler in diff function.

src/extra/AutoUI/utils.ts
```diff
@@ -71,3 +77,41 @@ export const autoUIGetDisabledReason = <T extends AutoUIBaseResource<T>>(
...
+		};
+	}
+	return (a: T, b: T) =>
+		// @ts-expect-error
+		diff(a[schemaKey], b[schemaKey]);
```
JSReds, line 104: same as above

src/extra/AutoUI/Filters/FocusSearch.tsx
```diff
@@ -0,0 +1,146 @@
...
+const FocusContent = styled(Box)`
+	max-height: 180px;
+	overflow-y: auto;
+	overflow-x: auto;
+`;
+
+const FocusItem = styled(Box)<{ hasGetBaseUrl: boolean }>`
+	cursor: ${(props) => (props.hasGetBaseUrl ? "pointer" : "default")};
+	&:hover {
+		background: #dde1f0; // This is the background color Select uses for entities on hover. We do not have it in our theme
```
JSReds, line 36: I'm ok if we do it in another PR but let's add it in the theme or we will soon probably make a mistake

src/extra/AutoUI/Filters/FocusSearch.tsx
```diff
@@ -0,0 +1,146 @@
...
+	}
+`;
+
+interface FocusSearchProps<T extends { id: number; [key: string]: any }> {
```
JSReds, line 40: I think that this can also be done like:

```suggestion
interface FocusSearchProps<T extends { id: number; searchTerms: string[] }> {
```

src/extra/AutoUI/Filters/FocusSearch.tsx
```diff
@@ -0,0 +1,146 @@
...
+	hasUpdateActions,
+}: FocusSearchProps<T>) => {
+	const history = useHistory();
+
+	const queryTerms = searchTerm.split(" ").map((x) => x.toLowerCase());
+
+	const filteredById = groupBy(
+		filtered.map((entity) => ({
+			id: entity.id,
+			searchTerms: Object.values(entity).filter((val) => val?.length > 0),
```
JSReds, line 66: by doing this you are keeping all values that are strings or arrays, skipping numbers. 
Probably better to keep only strings and numbers or only strings.

In case you want to also keep numbers then searchTerms: (string | number)[]

JSReds (commented):
Some strange things in this PR, let's try to improve them

src/extra/AutoUI/CustomWidget.tsx
```diff
@@ -0,0 +1,125 @@
...
+		<Widget
+			extraContext={extraContext}
+			extraFormats={extraFormats}
+			value={processedValue}
+			schema={schema}
+		/>
+	);
+};
+
+export const getColumnsFromSchema = <T extends AutoUIBaseResource<T>>({
```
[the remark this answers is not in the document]
myarmolinsky: Understandable, I put it here because it was rendering CustomWidgets. By name it sounds like it fits better inside schemaOps but it does not work in there (not exactly sure why). I've put it inside AutoUI's index for now, if you kbow why it may not be working inside schemaOps or if you can think of a better place to put it, please let me know

src/extra/AutoUI/utils.ts
myarmolinsky: @JSReds I do not know, this was pre-existing code. I have just moved some things around, which includes bringing this to this file

src/extra/AutoUI/utils.ts
myarmolinsky: Another instance of pre-existing code @JSReds. I've changed the code as per your suggestion but I'm not 100% sure I did what you had in mind, could you please check it out and confirm?

src/extra/AutoUI/utils.ts
myarmolinsky: > Another instance of pre-existing code @JSReds. I've changed the code as per your suggestion but I'm not 100% sure I did what you had in mind, could you please check it out and confirm?

Same here, could you please confirm?

src/extra/AutoUI/Filters/FocusSearch.tsx
myarmolinsky: This is from the Focus Search PR, not a change from this PR. Agreed though, we should do so in a separate PR sooner than later. Only thing is I don't know if we want to keep using grommet's Select's hover effect color rather than using a different hover effect color of our choice

src/extra/AutoUI/Filters/FocusSearch.tsx
myarmolinsky: I don't think so, I don't recall exactly but judging by what happens when I accept your suggestion, I think `[key: string]` is meant to be any key in the data object and `any` is to allow the value of the key to be any type (string, boolean, array)

src/extra/AutoUI/Filters/FocusSearch.tsx
myarmolinsky: I just add `typeof val === 'number' ||` before the `val?.length > 0`. Refer to the above comment for why I think `searchTerms: (string | number)[]` won't work

src/extra/AutoUI/utils.ts
JSReds: I was referring to the @ts-expect-error, this is clearly something forced here.  let's leave it like this I'll refactor this part as soon as I can

src/extra/AutoUI/Filters/FocusSearch.tsx
JSReds: 🤔 ok let's leave it like this and I'll fix the typing as soon as I can

myarmolinsky:
@balena-ci I self certify!

PR #1484, opened 2021-12-09
Make getBaseUrl accept full URLs as strings

myarmolinsky:
Make getBaseUrl accept full URLs as strings

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Table/TableRow.tsx
+++ b/src/components/Table/TableRow.tsx
@@ -91,13 +91,22 @@ export class TableRow<T> extends React.PureComponent<TableRowProps<T>, {}> {
 						typeof column.cellAttributes === 'function'
 							? column.cellAttributes(data, get(data, column.field))
 							: column.cellAttributes || {};
+					let url: URL | null;
+					try {
+						url = new URL(href ?? '');
+					} catch (err) {
+						url = null;
+					}
 					return (
 						<a
 							{...attributes}
-							href={href}
 							data-display="table-cell"
+							href={href}
 							data-key={keyAttribute}
-							onClick={this.props.onRowClick}
+							onClick={(props) => {
+								this.props.onRowClick(props);
+							}}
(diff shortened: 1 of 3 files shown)


src/components/Table/TableRow.tsx
```diff
@@ -91,19 +91,38 @@ export class TableRow<T> extends React.PureComponent<TableRowProps<T>, {}> {
...
+								data-key={keyAttribute}
+								onClick={(props) => {
+									this.props.onRowClick(props);
+									window.open(url.toString(), '_blank');
+								}}
+								{...cellAttributes}
+								key={column.key || (column.field as string)}
+							>
+								{renderField(data, column)}
+							</a>
```
thgreasi, line 109: How about parsing the url in its own try-catch (check if there is one first), and then have a single render which sets the correct properties (rather than repeating it).

myarmolinsky: I happened to do that right after creating the PR, thought I pushed the code but I guess not, my bad 😓

src/components/Table/TableRow.tsx
```diff
@@ -91,13 +91,22 @@ export class TableRow<T> extends React.PureComponent<TableRowProps<T>, {}> {
 						typeof column.cellAttributes === 'function'
 							? column.cellAttributes(data, get(data, column.field))
 							: column.cellAttributes || {};
+					let url: URL | null;
+					try {
+						url = new URL(href ?? '');
+					} catch (err) {
+						url = null;
+					}
```
JSReds, line 99: I'm not too sure, but I would probably use something like:
```
const isInternalUrl = !href.startsWith('http')
```

so we can avoid the try catch and new URL function

myarmolinsky:
@balena-ci I self certify!

PR #1487, opened 2021-12-13
Filters: Stop saving filters to local storage

myarmolinsky:
Filters: Stop saving filters to local storage

Connects-to: https://github.com/balena-io/balena-ui/issues/5244
Change-type: major

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/extra/AutoUI/Filters/PersistentFilters.tsx
+++ b/src/extra/AutoUI/Filters/PersistentFilters.tsx
@@ -102,7 +102,6 @@ const loadRulesFromUrl = (
 
 interface PersistentFiltersProps extends FiltersProps {
 	viewsRestorationKey: string;
-	filtersRestorationKey: string;
 	history: History;
 }
 
@@ -113,7 +112,6 @@ export const PersistentFilters = ({
 	onViewsUpdate,
 	onFiltersUpdate,
 	viewsRestorationKey,
-	filtersRestorationKey,
 	history,
 	onSearch,
 	...otherProps
@@ -124,11 +122,8 @@ export const PersistentFilters = ({
 	);
 	const locationSearch = history?.location?.search ?? '';
 	const storedFilters = React.useMemo(() => {
-		const urlRules = loadRulesFromUrl(locationSearch, schema);
-		return !!urlRules?.length
-			? urlRules
-			: getFromLocalStorage<JSONSchema[]>(filtersRestorationKey) ?? [];
-	}, [locationSearch, schema, filtersRestorationKey]);
+		return loadRulesFromUrl(locationSearch, schema);
+	}, [locationSearch, schema]);
 
 	React.useEffect(() => {
 		if (!views?.length && storedViews.length && onViewsUpdate) {
@@ -144,15 +139,11 @@ export const PersistentFilters = ({
 		});
 	};
 
-	// When the component mounts, if the parent component hasn't provided any filters
-	// and we were able to load some from the local storage, then communicate them
-	// back to the parent component.
+	// When the component mounts, filters from the page URL,
+	// then communicate them back to the parent component.
 	React.useEffect(() => {
-		updateUrl(storedFilters);
-		if (!filters?.length && storedFilters.length && onFiltersUpdate) {
-			onFiltersUpdate(storedFilters);
-		}
-	}, []);
+		onFiltersUpdate?.(storedFilters);
+	}, [storedFilters]);
 
 	const viewsUpdate = (views: FiltersView[]) => {
 		setToLocalStorage(viewsRestorationKey, views);
@@ -163,7 +154,6 @@ export const PersistentFilters = ({
 	};
 
 	const filtersUpdate = (filters: JSONSchema[]) => {
-		setToLocalStorage(filtersRestorationKey, filters);
 		updateUrl(filters);
 
 		if (onFiltersUpdate) {
--- a/src/extra/AutoUI/index.tsx
+++ b/src/extra/AutoUI/index.tsx
@@ -59,7 +59,7 @@ const HeaderGrid = styled(Flex)`
 	}
 `;
 
-export interface AutoUIProps<T> extends BoxProps {
+export interface AutoUIProps<T> extends Omit<BoxProps, 'onChange'> {
 	/** Model is the property that describe the data to display with a JSON structure */
 	model: AutoUIModel<T>;
 	/** Array of data or data entity to display */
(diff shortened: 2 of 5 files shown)


thgreasi:
tbh this sounds like a something you can call a major ;)

src/extra/AutoUI/index.tsx
```diff
@@ -88,6 +89,9 @@ export interface AutoUIProps<T> extends BoxProps {
 		entry: T,
 		event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
 	) => void;
+	// TODO: onChange should also be called when data in the table is sorted and when columns change
+	/** Function that gets called when filters change */
+	onChange?: (changes: { resource: string; filters?: JSONSchema[] }) => void;
```
thgreasi, line 94: I would be fine even if the `resource` wasn't returned, given that the consumer is providing it anyway as part of the `model`.

src/extra/AutoUI/index.tsx
```diff
@@ -142,6 +148,16 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
 
 	React.useEffect(() => {
 		setSelected([]);
+		if (filters.length) {
+			const { pathname } = window.location;
+			history?.replace?.({
```
thgreasi, line 153: Do we really need this (if so let's leave a comment on why)?
Isn't the `history.replace` that the PersistentFilters do enough?

src/extra/AutoUI/index.tsx
```diff
@@ -142,6 +148,16 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
...
+				pathname,
+				search: listFilterQuery(model.schema, filters),
+			});
+		}
+		if (onChange) {
+			onChange({ resource: model.resource, filters });
```
thgreasi, line 159: Aren't we already doing this in the `Filter`'s `changeFilters` callback (potentially resulting double calls)?
Would defining a single `onChangeFilters = useCallback(...)` function and using it in both `Filter` components allow us to drop this?

src/extra/AutoUI/index.tsx
myarmolinsky: I needed it at some point while testing my changes, but I guess it's no longer necessary. Removed it

src/extra/AutoUI/index.tsx
myarmolinsky: Seems doing `changeFilters` here is unnecessary so I've removed it

myarmolinsky:
@balena-ci retest

src/extra/AutoUI/Filters/PersistentFilters.tsx
```diff
@@ -125,10 +123,8 @@ export const PersistentFilters = ({
 	const locationSearch = history?.location?.search ?? '';
 	const storedFilters = React.useMemo(() => {
 		const urlRules = loadRulesFromUrl(locationSearch, schema);
-		return !!urlRules?.length
-			? urlRules
-			: getFromLocalStorage<JSONSchema[]>(filtersRestorationKey) ?? [];
-	}, [locationSearch, schema, filtersRestorationKey]);
+		return !!urlRules?.length ? urlRules : [];
```
thgreasi, line 126: Is this just v ?
```suggestion
		return urlRules ?? [];
```
After further checking, it seems that `loadRulesFromUrl` always returns an array, so we can just return the result of `loadRulesFromUrl()` directly :)

src/extra/AutoUI/Filters/PersistentFilters.tsx
```diff
@@ -148,11 +144,8 @@ export const PersistentFilters = ({
 	// and we were able to load some from the local storage, then communicate them
```
thgreasi, line 144:
```suggestion
	// and we were able to load some from the page url, then communicate them
```
but the whole comment might need changes.

src/extra/AutoUI/Filters/PersistentFilters.tsx
```diff
@@ -148,11 +144,8 @@ export const PersistentFilters = ({
...
 	// back to the parent component.
 	React.useEffect(() => {
-		updateUrl(storedFilters);
-		if (!filters?.length && storedFilters.length && onFiltersUpdate) {
-			onFiltersUpdate(storedFilters);
-		}
-	}, []);
+		filtersUpdate(storedFilters);
```
thgreasi, line 147: Does this need the `updateUrl` that `filtersUpdate()` triggers?
If yes, then let's plz move this useEffect bellow the definition of `filtersUpdate`.
If not (since `filtersUpdate()` is also called by `<Filters />`'s `onFiltersUpdate()`), I would say let's just call `onFiltersUpdate(filters);`.

thgreasi (commented):
Other than the question on whether the `React.useEffect filtersUpdate(storedFilters)` can be simplified LGTM :+1:  Feel free to self-certify.

myarmolinsky:
> Other than the question on whether the `React.useEffect filtersUpdate(storedFilters)` can be simplified LGTM 👍 Feel free to self-certify.

Thank you. I will hold off on the self-certification until the UI PR is approved as well so that we don't risk a rendition bump before it's done leaving users with *no* filter persistence

src/extra/AutoUI/Filters/PersistentFilters.tsx
myarmolinsky: Nice catch :)

src/extra/AutoUI/Filters/PersistentFilters.tsx
myarmolinsky: It is not needed, good catch. Replaced this call with just `onFiltersUpdate`,

thgreasi (commented):
LGTM :+1:

thgreasi:
Kind reminder to also do the `rawVersion` => `raw_version` change as part of this major if you get the time.
https://github.com/balena-io-modules/rendition/blob/39c1cf8c7dff625579f870cfd65091e7a2007814/src/unstable-temp/DownloadImageModal/models.ts#L53

myarmolinsky:
> Kind reminder to also do the `rawVersion` => `raw_version` change as part of this major if you get the time. See:
> 
> https://github.com/balena-io-modules/rendition/blob/39c1cf8c7dff625579f870cfd65091e7a2007814/src/unstable-temp/DownloadImageModal/models.ts#L53
> 
> 
> See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/efJNf0ALenbim-1KlTunx1YxiOo

Thank you for the reminder, but I had already made the change :)
The only `rawVersion`s still left are parameters such as for the `downloadConfig` prop on the `UnstableTempDownloadImageModalProps` interface. Are we okay with having underscores in parameter names?

myarmolinsky:
@balena-ci retest

myarmolinsky:
@balena-ci I self certify!

PR #1499, opened 2021-12-23
Make notification options accept prefix from alert props

myarmolinsky:
Make notification options accept prefix from alert props

Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/components/Notifications/index.tsx
+++ b/src/components/Notifications/index.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import ReactNotification, { store } from 'react-notifications-component';
 import styled from 'styled-components';
 import { animations } from '../../animations';
-import { Alert } from '../Alert';
+import { Alert, AlertProps } from '../Alert';
 import styles from './defaultStyle';
 
 type CONTAINER =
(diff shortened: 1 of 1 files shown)


src/components/Notifications/index.tsx
```diff
@@ -70,19 +70,13 @@ const getTransformedOptions = (
...
-	children,
+	content,
 	type,
 	onDismiss,
 	id,
-}: NotificationContainerProps) => {
+	...props
+}: NotificationOptions) => {
 	return (
 		<FullWidthContainer
```
thgreasi, line 81: BTW, any reason we didn't inline the `width: 100%;` here and instead we preferred the `styled(Alert)` way?
Imo it's more readable when everything is in one place, we can avoid some precedence edge cases, and such styled system props should perform similarly.

thgreasi (approved):

src/components/Notifications/index.tsx
myarmolinsky: No clue what the reasoning was behind it, this was before my time here 😅

myarmolinsky:
@balena-ci I self certify!

Issue #1465, opened 2021-11-22
Should use `Txt` as the underlying component for rendering text inside `TextWithCopy`


myarmolinsky:
Should also allow accepting `truncate` and other `Txt` props

myarmolinsky:
Resolved by: #1507

PR #1507, opened 2022-01-11
Make Txt base for TextWithCopy component

jmakivic:
Connects-to: #1465  
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/mPP--5c357OWJJkaU2ZdUhxa1B5
Change-type: minor

---
##### Contributor checklist

##### Reviewer Guidelines
---


jmakivic:
Truncate

I added some properties to Txt which make it look like highlighted code. I also added properties which truncate the content it if it gets too long. I wrapped Txt in the <Copy> component. This screenshot features DeviceSummaryCard.tsx.

I applied these changes to  the public IP address and the MAC address. As you can see, the long public IP address gets truncated if there is not enough space. 

This is what the code looks like  on DeviceSummaryCard.tsx . 
```
{device.public_address && (
	<div>
	   <Label>{t('labels.public_ip_address')}</Label>
	   <Copy content={device.public_address} onClick={function noRefCheck(){}} title={device.public_address} show="always" >
	              <Txt style={{width: "80%", whiteSpace: "nowrap", display:"inline-block" }} code truncate>{device.public_address</Txt>
	   </Copy>
	</div>
)}
{device?.mac_address != null && (
	<div>
	   <Label>{t('labels.mac_address')}</Label>
	   <Copy content={device.mac_address}  onClick={function noRefCheck(){}} show="always" title={device.mac_address} >
			<Txt style={{ whiteSpace: "nowrap", display:"inline-block" }} code truncate>{device.mac_address}</Txt>
	   </Copy>
	 </div>
)}
```

src/components/Txt/index.tsx
```diff
@@ -3,7 +3,23 @@ import styled, { css } from 'styled-components';
...
 import { RenditionSystemProps, Theme } from '../../common-types';
-import { monospace } from '../../utils';
+import { px, monospace } from '../../utils';
+
+export const code = (props: ThemedTxtProps) =>
+	props.code
+		? css`
+				font-family: ${(props) => props.theme.monospace};
+				padding: 2px 4px;
+				font-size: 90%;
```
thgreasi, line 13: This get's overrides by the `font-size: 1em;` right below.

src/components/Txt/index.tsx
```diff
@@ -3,7 +3,23 @@ import styled, { css } from 'styled-components';
...
+				color: #c7254e;
+				background-color: #f9f2f4;
+				border-radius: 2px;
+				white-space: normal;
+				word-wrap: break-word;
+				font-size: 1em;
```
thgreasi, line 19: Let's remove this, since it will not have any effect after we remove the `font-size: 90%;` from above.

src/components/Copy/index.tsx
```diff
@@ -14,6 +14,9 @@ const ClipboardContainer = styled(Box)`
 	font-size: 0.875em;
 	line-height: ${(props) =>
 		px(props.theme.lineHeight * props.theme.fontSizes[2])};
+	&:hover {
+		visibility: visible;
+	}
```
thgreasi, line 19: I checked out the respective UI branch and couldn't see any difference w/ or w/o this line.
Can you expand why we needed this?

thgreasi:
Btw it's fine to use `Resolves: #<number>` in Rendition & in the UI in case the PR completely fixes the issue and it can be closed on merge.

src/components/Txt/index.tsx
```diff
@@ -119,6 +134,8 @@ Base.displayName = 'Txt';
 Base.span = Factory('span');
 Base.p = Factory('p');
 
+// defining the stuff for copy, maybe this can be it's own thing in another
```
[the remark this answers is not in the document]
jmakivic: That was a comment which I wrote for myself which I forgot to delete. I removed it.

src/components/Txt/index.tsx
```diff
@@ -107,6 +121,7 @@ export interface InternalTxtProps extends React.HTMLAttributes<HTMLElement> {
 	align?: Align;
 	/** If true, replace the text not contained in the container with three dots */
 	truncate?: boolean;
+	code?: boolean;
```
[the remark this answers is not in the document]
jmakivic: I moved it to be after the italic prop.

src/components/Copy/index.tsx
```diff
@@ -22,7 +25,7 @@ const CopyContainer = styled(Flex)<{
 	${ClipboardContainer} {
 		display: ${(props) => (props.show === 'always' ? 'inline' : 'none')};
 	}
-
+	}
```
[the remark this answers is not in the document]
jmakivic: Do you mean the extra semicolon or the code on line 26? Because the code on line 26 seems valid to me.

package.json
```diff
@@ -136,6 +136,7 @@
     "remark-breaks": "^2.0.1",
     "remark-parse": "^8.0.3",
     "remark-rehype": "^7.0.0",
+    "rendition": "^25.2.0",
```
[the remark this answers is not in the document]
jmakivic: I undid these changes to the package.json.

src/components/Txt/index.tsx
```diff
@@ -93,7 +108,6 @@ export type Align =
 	| 'unset';
 
 export interface InternalTxtProps extends React.HTMLAttributes<HTMLElement> {
-	/** If true, render text in a monospace font */
```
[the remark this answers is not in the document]
jmakivic: I added it back in. Must have gotten misplaced while I was editing.

src/components/Txt/index.tsx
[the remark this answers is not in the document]
jmakivic: This was just a comment I left for myself which I forgot to delete. It's gone.

src/components/Txt/index.tsx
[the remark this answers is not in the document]
jmakivic: I moved it after `italic`.

src/components/Copy/index.tsx
jmakivic: I was trying to replicate the properties of the <Copy> component onto the `copy` prop. It turns out this element isn't necessary for the `copy` prop. I removed it.

src/components/Copy/index.tsx
```diff
@@ -47,6 +46,7 @@ const BaseCopy = ({
 			{...otherProps}
 		>
 			{children}
+
```
JSReds, line 49: we can probably just revert the entire fille unless this is a change made by the linter

src/components/Txt/index.tsx
```diff
@@ -48,24 +60,54 @@ export const align = style({
...
+	return asRendition<React.FunctionComponent<TxtProps>>((props: any) => {
+		return <BaseTxt as={tag} {...props}>
+    	{props.children}
+    	{!!props.copy &&
+        	<Copy content={props.children} show={props.copy} {...(props.copy.show && { showCopyButton: props.copy.show })}/>
+    	}
+		</BaseTxt>;
+		
+	});
+};*/
```
JSReds, line 109: let's remove this if not needed

src/extra/MarkdownEditor/defaultStyle.ts
```diff
@@ -207,16 +207,16 @@ export default css`
 	}
 	.CodeMirror-scroll {
 		overflow: scroll !important;
-		margin-bottom: -30px;
```
JSReds, line 210: Just to make sure, how this file was changed ?

src/components/Txt/index.tsx
```diff
@@ -107,6 +151,10 @@ export interface InternalTxtProps extends React.HTMLAttributes<HTMLElement> {
 	align?: Align;
 	/** If true, replace the text not contained in the container with three dots */
 	truncate?: boolean;
+	/** If true, it adds the copy icon and functionality at the end of a Txt component */
+	copy?: boolean;
+	/**  If true, the copy icon only appears when hovering over the content to be copied */
+	hoverCopy?: boolean;
```
JSReds, line 157: I would honestly keep it as `showCopyButton` maybe we can change it to `showCopyMode`, just in case we want to handle multiple states, and also to keep the same property we have in the Copy component.

```suggestion
	showCopyMode?: 'always' | 'hover';
```

src/components/Txt/index.tsx
```diff
@@ -48,24 +60,54 @@ export const align = style({
...
 	${truncate}
 `;
 
 const Factory = (tag?: string) => {
 	return asRendition<React.FunctionComponent<TxtProps>>((props: any) => {
-		return <BaseTxt as={tag} {...props} />;
+		return props.copy ? (
+			<Copy
+				content={props.children}
+				show={props.hoverCopy ? 'hover' : 'always'}
```
JSReds, line 89: if we keep the same property as suggested in another comment

```suggestion
				show={props.showCopyMode || 'always'}
```

src/components/TextWithCopy/index.tsx
```diff
@@ -12,7 +12,6 @@ import { Txt, TxtProps } from '../Txt';
 const Wrapper = styled(Txt.span)<InternalTextWithCopyProps>`
```
JSReds, line 12: I suppose that once we add the property to the Txt component, this file can be deleted

src/components/Txt/index.tsx
```diff
@@ -48,24 +60,54 @@ export const align = style({
...
+export const Lopy = (props: ThemedTxtProps) =>
+	props.truncate &&
+	css`
+		cursor: pointer;
+	`;
+
+export const LopyWrapper = styled.span`
+	display: inline-block;
+`;
+
```
JSReds, line 72: What is Lopy ? and where are we using them ?

src/components/Txt/index.tsx
```diff
@@ -48,24 +60,54 @@ export const align = style({
...
-
 	${truncate}
 `;
 
 const Factory = (tag?: string) => {
 	return asRendition<React.FunctionComponent<TxtProps>>((props: any) => {
-		return <BaseTxt as={tag} {...props} />;
+		return props.copy ? (
+			<Copy
+				content={props.children}
```
JSReds, line 88: Not sure why TS is not complaining here but you are passing a possible JSX element as a string property.  This is not allowed, or at least this is not correct. You should pass porps.copy that should be a string containing what we want to copy

jmakivic: I understand. I think this is what Matt brought up in the flowdock discussion. I changed content to be props.copy.

src/components/Txt/index.tsx
```diff
@@ -107,6 +151,10 @@ export interface InternalTxtProps extends React.HTMLAttributes<HTMLElement> {
 	align?: Align;
 	/** If true, replace the text not contained in the container with three dots */
 	truncate?: boolean;
+	/** If true, it adds the copy icon and functionality at the end of a Txt component */
+	copy?: boolean;
```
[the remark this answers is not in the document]
jmakivic: Changed it.

src/extra/MarkdownEditor/defaultStyle.ts
jmakivic: I'm not sure. I don't recall changing it.

myarmolinsky: If this was not intentional then please revert it

src/components/Txt/index.tsx
```diff
@@ -52,17 +64,22 @@ const BaseTxt = styled.div<TxtProps>`
...
 	${truncate}
 `;
 
 const Factory = (tag?: string) => {
 	return asRendition<React.FunctionComponent<TxtProps>>((props: any) => {
-		return <BaseTxt as={tag} {...props} />;
+		return props.copy ? (
+			<Copy content={props.copy} show={props.showCopyMode || 'always'}>
+				<BaseTxt as={tag} {...props} />
+			</Copy>
```
myarmolinsky, line 79: Since `Copy` isn't receiving `{...props}`, its styling will be off compared to `BaseTxt`'s. So, I'm not certain but I think `{...props}` should be passed to `Copy` instead of `BaseTxt`. I believe `BaseTxt` will not be happy about no longer receiving `{...props}` so please look into that. If `BaseTxt` requires `props` to function, maybe there is a specific property in `props` that we can destructure and pass to it in order to get it to work?

Giving `Copy` `{...props}` won't be enough to get its styling working properly. I was testing by updating the `Txt.span` containing the `app.total_devices` in the UI's `ApplicationListBox` component so you can use that as your test case. You'll see that while giving `Copy` `{...props}` makes it resize and obey color and other styling, it does not center the text like it is supposed to. Please see if you can find what's causing that and how to fix it.

src/components/Txt/index.tsx
```diff
@@ -107,6 +126,10 @@ export interface InternalTxtProps extends React.HTMLAttributes<HTMLElement> {
 	align?: Align;
 	/** If true, replace the text not contained in the container with three dots */
 	truncate?: boolean;
+	/** Adds copy functionality to Txt and accepts the value to be copied as a string */
+	copy?: string;
+	/**  Setting to 'always' ensures that the copy icon is always visible. 'hover' ensures that the copy icon is only visible on hover. If this prop is not included, the copy icon is set to hover */
```
myarmolinsky, line 131:
```suggestion
	/**  Setting to 'always' ensures that the copy icon is always visible. 'hover' ensures that the copy icon is only visible on hover. By default, this is set to 'hover'. */
```
```suggestion
	/**  Setting to 'always' ensures that the copy icon is always visible. 'hover' ensures that the copy icon is only visible on hover. If this prop is not included, the copy icon is set to hover */
```

src/components/Txt/index.tsx
```diff
@@ -107,6 +126,10 @@ export interface InternalTxtProps extends React.HTMLAttributes<HTMLElement> {
 	align?: Align;
 	/** If true, replace the text not contained in the container with three dots */
 	truncate?: boolean;
+	/** Adds copy functionality to Txt and accepts the value to be copied as a string */
```
myarmolinsky, line 129:
```suggestion
	/** Displays a copy icon which copies the given string to the user's clipboard onClick */
```

src/extra/MarkdownEditor/defaultStyle.ts
myarmolinsky: Please revert the changes to this file then

src/components/Copy/index.tsx
```diff
@@ -22,7 +22,6 @@ const CopyContainer = styled(Flex)<{
 	${ClipboardContainer} {
 		display: ${(props) => (props.show === 'always' ? 'inline' : 'none')};
 	}
-
```
myarmolinsky, line 25: Please revert this file as it is unchanged

src/components/Txt/index.tsx
```diff
@@ -107,6 +155,12 @@ export interface InternalTxtProps extends React.HTMLAttributes<HTMLElement> {
 	align?: Align;
 	/** If true, replace the text not contained in the container with three dots */
 	truncate?: boolean;
+	/** Displays a copy icon which copies the given string to the user's clipboard onClick */
+	copy?: string;
+	/**  Setting to 'always' ensures that the copy icon is always visible. 'hover' ensures that the copy icon is only visible on hover. By default, this is set to 'hover'. */
+	showCopyMode?: 'hover' | 'always';
+	/** Displays Txt as inline-block and vertical aligns items at the bottom. Good for aligning text with icons. */
+	iconalign?: boolean;
```
myarmolinsky, line 163: Why do we need this as a property? The Txt should use the appropriate display by default. If we want the user to be able to change it, we should have a property called `display` (it exists on both BoxProps and FlexProps). Alignment should be a separate property as well, likely and preferably from FlexProps or BoxProps, and it should have a good, sensible default value

src/components/Txt/index.tsx
```diff
@@ -48,21 +63,52 @@ export const align = style({
...
+				<BaseTxt as={tag} {...props} iconalign />
+				<Copy
+					tooltip={{ text: 'Copied!', trigger: 'click' }}
+					onClick={() => copyToClipboard(props.copy)}
+					showCopyButton={props.showCopyMode || 'always'}
+					{...props}
+				>
+					<FontAwesomeIcon icon={faCopy} />
+				</Copy>
+			</div>
```
myarmolinsky, line 108: Why do we need to wrap this in a div? And howcome Copy is wrapping a copy icon instead of the BaseTxt? If the user provides copy, the result should be something like `<Copy><BaseTxt /></Copy>`. Any styling updates that need to be made in order for it to work properly in any contexts (consider the ones in our UI) should be applied to one of those two components.

JSReds: @myarmolinsky IMHO this is the correct structure. Copy as it was before is just a styled component that wrap an icon, as for the style we already discussed that the icon should not have any other than the font size. (ref: https://www.flowdock.com/app/rulemotion/f-rendition/threads/tcqn5E_rvx-y9M66LSJHg57t3Ki)

Also @jmakivic what about:

```
			<div>
				<BaseTxt as={tag} {...props} iconalign />
{ props.copy && (
				<Copy
					tooltip={{ text: 'Copied!', trigger: 'click' }}
					onClick={() => copyToClipboard(props.copy)}
					showCopyButton={props.showCopyMode || 'always'}
          fontSize={props.fontSize}
				>
					<FontAwesomeIcon icon={faCopy} />
				</Copy> )
			}
			</div>
```

src/components/Txt/index.tsx
```diff
@@ -48,21 +63,52 @@ export const align = style({
...
+	${code}
 	${caps}
 	${bold}
 	${italic}
-
 	${truncate}
+	${iconalign}
 `;
-
+// style={{ display: 'inline-block', verticalAlign: 'bottom' }}
```
JSReds, line 94: Remember to remove

src/components/Txt/index.tsx
JSReds: Agree with this. We could probably remove it without any replacement, unless I'm forgetting any strange cases

src/components/Txt/index.tsx
```diff
@@ -48,22 +64,63 @@ export const align = style({
...
+	visibility: ${(props) =>
+		props.showCopyButton === 'always' ? 'visible' : 'hidden'};
+	&:hover {
+		visibility: visible;
+	}
+	display: inline;
+`;
+
+const copy = (props: ThemedTxtProps) =>
+	props.copy &&
```
JSReds, line 78: Let's make this check a boolean

```suggestion
	!!props.cop &&
```

JSReds (commented):
Other than that, LGTM :)

jmakivic:
I changed the `onCopy` prop back to `onCopyClick`. `onCopy` appears to be an existing react property. I saw this error when I added it as a Txt prop:

src/components/Link/index.tsx:12:4 - error TS2345: Argument of type '(props: ThemedTxtProps) => FlattenSimpleInterpolation | null' is not assignable to parameter of type 'Interpolation<ThemedStyledProps<Pick<DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>, "slot" | ... 263 more ... | "referrerPolicy"> & { ...; } & InternalLinkProps, any>>'.
  Type '(props: ThemedTxtProps) => FlattenSimpleInterpolation | null' is not assignable to type 'InterpolationFunction<ThemedStyledProps<Pick<DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>, "slot" | ... 263 more ... | "referrerPolicy"> & { ...; } & InternalLinkProps, any>>'.
    Types of parameters 'props' and 'props' are incompatible.
      Type 'ThemedStyledProps<Pick<DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>, "slot" | ... 263 more ... | "referrerPolicy"> & { ...; } & InternalLinkProps, any>' is not assignable to type 'ThemedTxtProps'.
        Types of property 'onCopy' are incompatible.
          Type '(((event: ClipboardEvent<HTMLAnchorElement>) => void) & ((event: ClipboardEvent<HTMLElement>) => void)) | undefined' is not assignable to type '(() => void) | undefined'.
            Type '((event: ClipboardEvent<HTMLAnchorElement>) => void) & ((event: ClipboardEvent<HTMLElement>) => void)' is not assignable to type '(() => void) | undefined'.
              Type '((event: ClipboardEvent<HTMLAnchorElement>) => void) & ((event: ClipboardEvent<HTMLElement>) => void)' is not assignable to type '() => void'.

This problem was resolved when I renamed `onCopy` to `onCopyClick`.

myarmolinsky (approved):

PR #1508, opened 2022-01-11
Fix AutoUI not filling screen height

myarmolinsky:
Fix AutoUI not filling screen height

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/extra/AutoUI/index.tsx
+++ b/src/extra/AutoUI/index.tsx
@@ -268,10 +268,12 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
 
 	return (
 		<Flex
+			flex={1}
 			flexDirection={Array.isArray(data) ? 'column' : undefined}
 			mt={Array.isArray(data) ? 2 : undefined}
 		>
 			<Spinner
+				flex={1}
 				label={
 					isBusyMessage ??
 					t('loading.resource', {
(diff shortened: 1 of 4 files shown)


src/extra/AutoUI/index.tsx
```diff
@@ -268,10 +268,12 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
 
 	return (
 		<Flex
+			height="100%"
 			flexDirection={Array.isArray(data) ? 'column' : undefined}
```
thgreasi, line 272: @JSReds do you remember the reason that we swapped the direction based on the type?
Given that there is only a single Spinner component inside it, I don't think that does what we expect.
A weird side-effect of it though is the jumpiness of the spinner in the Organizations page, right after the data arrive.
```suggestion
			flexDirection="column"
```

src/extra/AutoUI/index.tsx
```diff
@@ -280,131 +282,134 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
...
+							/>
+						</HeaderGrid>
+					)}
+
+					{lens &&
+						data &&
+						(!Array.isArray(data) ||
+							(Array.isArray(data) && data.length > 0)) && (
+							<lens.data.renderer
+								flex={1}
```
thgreasi, line 393: I think we also need to be wiring this in to the renderers, right?
But I guess that's not a blocker atm.

src/extra/AutoUI/index.tsx
```diff
@@ -268,10 +268,12 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
 
 	return (
 		<Flex
+			height="100%"
 			flexDirection={Array.isArray(data) ? 'column' : undefined}
 			mt={Array.isArray(data) ? 2 : undefined}
 		>
 			<Spinner
+				height="100%"
```
thgreasi, line 276: I think it also works with v right?
```suggestion
				flex={1}
```

src/extra/AutoUI/index.tsx
myarmolinsky: Are you referring to `entity.tsx`'s and `table.tsx`'s renderer props? Given that the two work fine at the moment without the flex, I figured it was unnecessary to do and that we would instead wire it to any lenses that don't work fine without it (i.e. the device map lens)

src/extra/AutoUI/index.tsx
```diff
@@ -268,10 +268,12 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
 
 	return (
 		<Flex
+			height="100%"
```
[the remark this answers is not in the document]
myarmolinsky: Swapped the height for a flex
Keeping this conversation open for the discussion on accepting and destructuring the FlexProps as well as for the PS you left at the bottom

src/extra/AutoUI/index.tsx
thgreasi: I have a feeling we might regret this later. I was going to pass for now, but on the other hand, someone could see that this `flex={1}` is then technically unused and remove it :thinking:

src/extra/AutoUI/index.tsx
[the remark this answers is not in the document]
thgreasi: The reasoning against a hardcoded `flex={1}` is that it would be unexpected in case this was used inside a Flex parent with more children. Eg:
```
<Flex height="100vh">
    <Box flex={1}>Huge graphic here, maybe a big map</Box>
    <AutoUI /> {/* trying to render a short table below the big graphic */}
</Flex>
```
In this case the two components would of course expand to the same height, but it will not be clear why.

thgreasi:
@balena-ci rebase

myarmolinsky:
@balena-ci I self certify!

src/extra/AutoUI/index.tsx
JSReds: @thgreasi this change come from this PR https://github.com/balena-io-modules/rendition/pull/1481/files and I think this was only a mistake made thinking about the collection vs entity. I hadn't noticed this, unfortunately, when I reviewed the PR. In any case, I agree with what you said above, I am quite convinced that this can be removed.

thgreasi: @myarmolinsky can you do this change in a follow-up PR and give it a try?

PR #1509, opened 2022-01-13
Update AutoUI affectedEntries for actions when selected data updates 

myarmolinsky:
Update AutoUI affectedEntries for actions when selected data updates 

Change-type: patch

---
##### Contributor checklist

##### Reviewer Guidelines
---

### DIFF
--- a/src/extra/AutoUI/index.tsx
+++ b/src/extra/AutoUI/index.tsx
@@ -125,13 +125,23 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
 
 	const [filters, setFilters] = React.useState<JSONSchema[]>([]);
 	const [views, setViews] = React.useState<FiltersView[]>([]);
-	const [selected, setSelected] = React.useState<T[]>([]);
+	const [selected, $setSelected] = React.useState<T[]>([]);
 	const [isBusyMessage, setIsBusyMessage] = React.useState<
 		string | undefined
 	>();
 	const [actionData, setActionData] = React.useState<
 		ActionData<T> | undefined
 	>();
+
+	const setSelected = React.useCallback(
+		(items: T[]) => {
+			$setSelected(items);
+			if (!!actionData) {
+				setActionData({ ...actionData, affectedEntries: items });
+			}
+		},
+		[$setSelected, setActionData, actionData],
+	);
 
 	const showFilters = React.useMemo(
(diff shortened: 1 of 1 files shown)


src/extra/AutoUI/index.tsx
```diff
@@ -129,6 +129,12 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
 		ActionData<T> | undefined
 	>();
 
+	React.useEffect(() => {
+		if (actionData) {
```
JSReds, line 133:
```suggestion
		if (!!actionData) {
```

JSReds (commented):
LGTM, let's test it before releasing it. Other than that you can self certify !

thgreasi:
May I ask for a clarification on why we needed to do that in the first place?
We are also calling `setActionData` in the `onActionTriggered`. Should that also be changed?
Afaict both src/extra/AutoUI/Actions/Create.tsx & src/extra/AutoUI/Actions/Update.tsx call onActionTriggered with `affectedEntries: selected`.
Should we instead do this wherever `setSelected()` is called, instead of relying on a `useEffect` and waiting for the extra re-render to finish?

myarmolinsky:
> * May I ask for a clarification on why we needed to do that in the first place?
>   Is this something like "Make AutoUI continue updating the selected item instances passed to AutoUI actions" ?
> * We are also calling `setActionData` in the `onActionTriggered`. Should that also be changed?
>   Afaict both src/extra/AutoUI/Actions/Create.tsx & src/extra/AutoUI/Actions/Update.tsx already call onActionTriggered with `affectedEntries: selected`.
> * Should we instead do this wherever `setSelected()` is called, instead of relying on a `useEffect` and waiting for the extra re-render to finish? (we are only calling `setSelected()` in 3 places atm).

- Yes, it's "Make AutoUI continue updating the selected item instances passed to AutoUI actions"
- They call it, but that means that the triggered action is applied to the selected *at the time of trigger*, it does not continue updating the selected after triggering
- Not sure, that means that the more `setSelected`s we may end up using in the future, the more we have to call `setActionData`. Where do we draw the line? At how many `setSelected`s do we start using the useEffect? It feels like these two are intertwined so I think it makes sense to make the relationship clear via a useEffect with [selected] in the dependency array. If you feel strongly otherwise, I can make the change

myarmolinsky:
@balena-ci I self certify!

PR #1512, opened 2022-01-14
AutoUI: Improve the Lenses renderer prop types

thgreasi:
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

##### Reviewer Guidelines
---


src/extra/AutoUI/Lenses/index.tsx
```diff
@@ -12,13 +12,17 @@ import { IconProp } from '@fortawesome/fontawesome-svg-core';
...
 	name: string;
 	data: {
 		label: string;
 		format: string;
-		renderer: (props: any) => React.ReactElement;
+		renderer: (
+			props:
+				| types.CollectionLensRendererProps<T>
+				| types.EntityLensRendererProps<T>,
+		) => React.ReactElement;
```
myarmolinsky, line 25: I wouldn't pass *either* CollectionLensRendererProps *or* EntityLensRendererProps. I would pass a list of all possible props that might be passed (see the following for all possible props that might be passed: https://github.com/balena-io-modules/rendition/blob/17cc165fa308a714d4a832d57b5f555c6f92ded6/src/extra/AutoUI/index.tsx#L389-L399)
The reason I would not restrict it to *either* Collection *or* Entity is because the user is able to create their own CustomLenses and we do not want to force them to use any props they do not need, so most or all of these props that we pass into `lens.data.renderer` are likely optional and the user should be free to use them all.

src/extra/AutoUI/Lenses/types/index.ts
```diff
@@ -1,2 +1,26 @@
...
+	) => void;
+}
+
+export interface EntityLensRendererProps<T> {
+	data: T;
+	properties: Array<any & { key: string }>;
+	autouiContext: AutoUIContext<T>;
+	model: AutoUIModel<T>;
+	hasUpdateActions: boolean;
+}
```
myarmolinsky, line 26: Most of these props are the same typing-wise. The only differences are `data` and `properties`. If you want we can have a `BaseLensRendererProps` and make CollectionLens and EntityLens extend it with their own variations of `data` and `properties`.

I also would not use the name `CollectionLens` because collection refers to the data being passed, not the way it is being rendered. The lens is a Table lens because it renders a table (using a collection of data). The Entity lens was an iffy one and I am not a fan of keeping it's name as Entity, I just was not sure what else to name it at the time. Perhaps Summary lens would be better?

What I am trying to get at with this comment and my other comment above is that each lens should be named appropriately as per what it renders and it should accept props as necessary rather than being restricted to these two interfaces which were made specifically for the `table` and `entity` lenses. For example, see the two custom lenses we currently have in the UI: Grid lens for ApplicationList.tsx and Map lens for Devices.tsx. The props their renderers accept differ from the default `table` lens props. The Grid lens only needs `filtered`, nothing else. And the Map lens only needs `filtered` and `autouicontext`.

myarmolinsky (changes requested):
I think these changes were made under the impression that the `table` lens renderer props and `entity` lens renderer props are intended to be the props for all AutoUI collections rendered and AutoUI entities rendered (respectively). However, this is not correct. The necessity of these props is exclusive to these specific lenses. Any other lenses (custom lenses the user might make) will not necessarily require the same set of props. We can make a `BaseLensRendererProps` or `LensRendererProps` or `AvailableLensRendererProps` interface which makes all the props known and describes them, but most or all of them should be optional so that the user can pick and choose the ones actually needed for their custom lens. Please see the custom lenses we have in the UI:
- `grid` in ApplicationList.tsx: https://github.com/balena-io/balena-ui/blob/1e2c001b0e076b3b2168f366c616e6c4fa90255f/src/modules/applications-list/ApplicationList.tsx#L473-L510
- `map` in Devices.tsx: https://github.com/balena-io/balena-ui/blob/1e2c001b0e076b3b2168f366c616e6c4fa90255f/src/pages/dashboard/application/Devices.tsx#L284-L320

src/extra/AutoUI/Lenses/index.tsx
thgreasi: This is not an exclusive OR in the typings. It just dictates that the AutoUI component is always guaranteed to be passing those props, depending on how the component is used. That's is a bit less vague for a consumer trying to create a lens. With this they know they what props we are guaranteeing to be setting depending on how the lens is used.

If we made all those props optional, then this will add extra unnecessary checks in the Lens' implementation, checking for each prop whether its defined or not, which with this you can be sure that as far your `fitler` `type` is respected, then you will always be passed the above props.
And it's not that the Lens will have to support all those props, it's just that it will have to use the props interface like:
```
export const table2: LensTemplate = {
	slug: 'table',
	name: 'Default table lens',
	data: {
		label: 'Table',
		format: 'table',
		renderer: (props: CollectionLensRendererProps<number>) => (
			<>{props.data?.length ?? null}</>
		),
		icon: faTable,
		type: '*',
		filter: {
			type: 'array',
			items: {
				type: 'number',
			},
		},
	},
};
```
Moreover for mixed components, you can reason the move in which the Lense is used by just checking one differentiating property and then for the rest of the implementation TS will only give you the correct types & props available in that mode.

thgreasi:
> but most or all of them should be optional so that the user can pick and choose the ones actually needed for their custom lens.

@myarmolinsky even if you declare them as optional, TS will complain if the component you are passing doesn't have them in its props at all. I''m probably missing something, so plz give your proposal a try and we can then compare how the two work.
In both cases consumers will have to just declare our exported props interface, otherwise they will have to replicate it locally b/c of how assignability checks work in TS and how TS will complain for missing/excessive props.

src/extra/AutoUI/Lenses/types/index.ts
thgreasi: > I also would not use the name CollectionLens because collection refers to the data being passed, not the way it is being rendered. The lens is a Table lens because it renders a table (using a collection of data). 

This was exactly why the interface should not be called TableLense. B/c uses can define their own lense which might render a different result. A different way to think this is that this is the interface that all lenses that are supposed to render a collection can use, regardless of whether its a table or whether it's a ApplicationBox DataGrid. The default TableLense is just the backed in default lense that we happen to offer as part of AutoUI.

Now for the UI custom lense that defines only `filtered` in its prop typings, that only works b/c the typings atm for the props are just `any`.

myarmolinsky:
> @myarmolinsky did we end up with which PR about this topic to go?
> Can we get yours or mine in to get better typings right away and we can still change our mind later if we want.

Let's go with yours for now and we can look into further improvements such as `createLens` later

myarmolinsky (approved):
LGTM, we can look into further improvements such as `createLens` after

PR #1531, opened 2022-02-04
Make monaco editor a dev dependency

LucianBuzzo:
Monaco editor is only used by the renderer playground in storybook and
shouldn't be included as a main dependency.

Change-type: patch


LucianBuzzo:
@balena-ci I self-certify!

PR #1544, opened 2022-02-23
Mermaid: Add support for callback functions in graphs

LucianBuzzo:
Fixes https://github.com/balena-io-modules/rendition/issues/1543

This change allows you to pass an object of callback functions that can
be called from click events in mermaid.
These functions have to be bound to the window object to work with
mermaid, which could potentially cause some problems if charts have the
same callback names.
I did try putting the mermaid chart inside an iframe to isolate the
functions, but I couldn't get this working - it seems mermaid then
cannot find the global function to call.

Change-type: minor

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [x] I have regenerated screenshots for any affected components with `npm run generate-snapshots`

##### Reviewer Guidelines
---

### DIFF
--- a/src/extra/Mermaid/Mermaid.stories.tsx
+++ b/src/extra/Mermaid/Mermaid.stories.tsx
@@ -8,6 +8,7 @@ export default {
 } as Meta;
 
 const Template = createTemplate<MermaidProps>(Mermaid);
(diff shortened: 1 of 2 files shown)


LucianBuzzo:
@balena-ci I self-certify!

PR #1548, opened 2022-03-07
Render mermaid with markdown

karaxuna:
Add ability to render mermaid charts with markdown. It's done via opening code block with language=markdown.

Change-type: minor

---

##### Contributor checklist

- [x ] I have regenerated jest snapshots for any affected components with `npm run generate-snapshots`


package.json
```diff
@@ -145,6 +147,7 @@
     "styled-system": "^4.2.4",
     "tslib": "^2.3.1",
     "unified": "^9.2.2",
+    "unist-util-visit": "^2.0.3",
```
LucianBuzzo, line 150: Both these dependencies are going to cause issues because they are esm only. Have you checked to see if this PR will work on balenaCloud dash and/or Jellyfish?
I also think we might want to open up the conversation about converting these projects to support esm...

karaxuna: @LucianBuzzo works well with jellyfish. The problem occurs when running unit tests right? Because node can't read es6 exports. For this we manually add each module to jest's [`transform`](https://github.com/product-os/jellyfish/blob/master/apps/ui/jest.config.js#L12) configuration property. Maybe we should just transform all javascript files?

karaxuna: Node > v13 understands es6 imports if the module has `type: 'module'` in package json (these newly added modules have this flag set), so it should not be a problem. Anyway, I'll test balena dashboard with this version of rendition.

LucianBuzzo (approved):

PR #1549, opened 2022-03-08
Move update actions

JSReds:
Move autoui update actions next to create, move tags action inside update actions, restyle.

See: https://jel.ly.fish/view-user-created-view-21857c98-7dad-4418-9724-202d66d1953d-open-product-call-items/brainstorm-topic-0551085a-9d0a-4b05-9eb7-cc2b14d3066d@1.0.0

Resolves: https://github.com/balena-io/balena-ui/issues/5394
Resolves: https://github.com/balena-io/balena-ui/issues/5395
Resolves: https://github.com/balena-io/balena-ui/issues/5392

Screenshot 2022-03-16 at 18 44 17
Screenshot 2022-03-16 at 18 43 55
Screenshot 2022-03-16 at 18 45 08

---

##### Contributor checklist

##### Reviewer Guidelines

---

### DIFF
--- a/src/components/Filters/ViewsMenu.tsx
+++ b/src/components/Filters/ViewsMenu.tsx
@@ -44,12 +44,6 @@ const ViewsMenu = (props: ViewsMenuProps) => {
 	const hasViews = views.length > 0;
 	const groupedViews = groupBy(views, (item) => item.scope || 'Unscoped');
 
-	let soloRender = false;
-	if (renderMode) {
-		const mode = Array.isArray(renderMode) ? renderMode : [renderMode];
-		soloRender = mode.length === 1 && mode[0] === 'views';
-	}
-
 	const memoizedItems = React.useMemo(
 		() =>
 			!hasViews
@@ -105,13 +99,13 @@ const ViewsMenu = (props: ViewsMenuProps) => {
 
 	return (
 		<DropDownButton
-			ml={soloRender ? 0 : 30}
+			mx={renderMode?.length === 1 ? 0 : 1}
 			disabled={props.disabled}
 			quartenary={!props.dark}
 			light={props.dark}
 			outline
 			joined
-			alignRight={!soloRender}
+			alignRight={renderMode?.length === 1}
 			noListFormat
 			icon={<FontAwesomeIcon icon={faChartPie} />}
 			label="Views"
@@ -131,7 +125,7 @@ export interface ViewsMenuProps {
 	hasMultipleScopes?: boolean;
 	setFilters: (filters: JSONSchema[]) => void;
 	deleteView: (view: FiltersView) => any;
-	renderMode?: FilterRenderMode | FilterRenderMode[];
+	renderMode?: FilterRenderMode[];
 	compact?: boolean[];
 }
 
--- a/src/extra/AutoUI/Actions/Create.tsx
+++ b/src/extra/AutoUI/Actions/Create.tsx
@@ -11,6 +11,7 @@ import { getCreateDisabledReason } from '../utils';
 import { Button } from '../../../components/Button';
 import { faMagic } from '@fortawesome/free-solid-svg-icons/faMagic';
 import { useTranslation } from '../../../hooks/useTranslation';
+import { Box } from '../../../components/Box';
 
 interface CreateProps<T extends AutoUIBaseResource<T>> {
 	model: AutoUIModel<T>;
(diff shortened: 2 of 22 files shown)


src/components/Filters/ViewsMenu.tsx
```diff
@@ -105,7 +105,7 @@ const ViewsMenu = (props: ViewsMenuProps) => {
 
 	return (
 		<DropDownButton
-			ml={soloRender ? 0 : 30}
+			mx={soloRender ? 0 : 1}
```
myarmolinsky, line 108: What's the point of `soloRender`? I see it's based on `renderMode` and `renderMode` is not used for anything except determining `soloRender`. Wouldn't it make more sense to remove this `ml`/`mx` from ViewsMenu and to instead apply it to ViewsMenu via extended FlexProps in the parent component?
i.e.:
```
<>
...
<ViewsMenu mx={renderMode?.length === 1 ? 0 : 1} {...otherRequiredProps} />
...
</>
```

src/extra/AutoUI/utils.ts
```diff
@@ -56,6 +56,8 @@ export const getTagsDisabledReason = <T extends AutoUIBaseResource<T>>(
 		// TODO: Pass the resource name instead.
 		return t('info.edit_tag_no_permissions', { resource: 'item' });
 	}
+
+	return false;
```
myarmolinsky, line 60: We let this slide without a return before? 😱

JSReds: Apparently yes :)

src/extra/AutoUI/Actions/Tags.tsx
```diff
@@ -5,63 +5,77 @@ import {
...
+			items={selected}
+			itemType={autouiContext.resource}
+			titleField={autouiContext.nameField as keyof T}
+			tagField={autouiContext.tagField as keyof T}
+			done={(tagSubmitInfo) => {
+				changeTags(tagSubmitInfo);
+				onDone();
+			}}
+			cancel={() => onDone()}
+		/>
```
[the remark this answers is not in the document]
JSReds: Honestly, I prefer to break things up. This component also handles the `changeTag` function. At the end of the show, there are more or less 70 fewer lines of code to read in a file that is already large (500 lines without that component). I honestly think a 500 line component is already a bit wrong, so I would avoid merging into the component

src/extra/AutoUI/Actions/Tags.tsx
```diff
@@ -5,63 +5,77 @@ import {
...
+			setIsBusyMessage(t(`loading.updating_release_tags`));
+			notifications.addNotification({
+				id: 'change-tags-loading',
+				content: t(`loading.updating_release_tags`),
+			});
+
+			try {
+				await sdk.tags.submit(tags);
+				notifications.addNotification({
+					id: 'change-tags',
```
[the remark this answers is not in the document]
JSReds: In this case not much, in other cases we use them to close a specific notification at a specific time

src/extra/AutoUI/index.tsx
```diff
@@ -282,74 +263,67 @@ export const AutoUI = <T extends AutoUIBaseResource<T>>({
...
+									</Box>
+									<Box alignSelf="flex-start">
+										<Update
+											model={model}
+											selected={selected}
+											autouiContext={autouiContext}
+											hasOngoingAction={false}
+											onActionTriggered={onActionTriggered}
+										/>
+									</Box>
```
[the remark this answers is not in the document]
JSReds: nice catch! But I've changed differently

src/components/Filters/ViewsMenu.tsx
JSReds: Sorry I haven't seen this comment yesterday. I would not expand ViewsMenu props since soloRender is used in two cases, the first is `alignRight` and the second is `mx`, so we should expand it with DropDownProps. I don't like the fact that Filters would contain a declaration that will be used only for ViewsMenu. This part needs a refactor, but I'll not do that in this PR. I've improved it, removing the entire if. I'm also wondering what is the case where we render only views... This button, without filters, is pretty useless IMHO...

src/extra/AutoUI/Actions/Create.tsx
```diff
@@ -42,20 +43,27 @@ export const Create = <T extends AutoUIBaseResource<T>>({
...
-				})
-			}
-			icon={<FontAwesomeIcon icon={faMagic} />}
-			tooltip={typeof disabledReason === 'string' ? disabledReason : undefined}
-			disabled={!!disabledReason}
-			primary
-		>
-			{createActions[0].title}
-		</Button>
+		<Box>
```
myarmolinsky, line 46: Why did we wrap the Button in a Box now?

myarmolinsky (commented):
LGTM 👍

JSReds:
@balena-ci I self-certify!

JSReds:
@balena-ci I self-certify!

PR #1587, opened 2022-05-09
DownloadImageModal: Add the provisioning key expiry date in the advanced options

nitishagar:
Change-Type: minor
Signed-off-by: Nitish Agarwal <1592163+nitishagar@users.noreply.github.com>

HQ: https://jel.ly.fish/e450713f-001c-4470-b571-71724aeb0869
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/vx-rBrpDBqXtkdtMmSU5MmPYSlb
Depends-on: https://github.com/balena-io/balena-sdk/pull/1217

Screenshot:
Screenshot 2022-05-12 at 15-12-13 STAGING balena dashboard Fleets

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,7 @@
     "lodash": "^4.17.21",
     "memoizee": "^0.4.15",
     "mermaid": "^8.13.9",
+    "moment": "^2.29.3",
     "pinejs-client-fetch": "^0.2.2",
     "prismjs": "^1.25.0",
     "prop-types": "^15.7.2",
--- a/src/unstable-temp/DownloadImageModal/FormModel.tsx
+++ b/src/unstable-temp/DownloadImageModal/FormModel.tsx
@@ -13,6 +13,7 @@ import styled from 'styled-components';
 import { DeviceTypeOptions, DeviceTypeOptionsGroup } from './models';
 import { DocsLink } from './OsConfiguration';
 import { useTranslation } from '../../hooks/useTranslation';
+import moment from 'moment';
 
 export const DownloadImageLabel = styled.label`
 	display: flex;
@@ -198,6 +199,42 @@ const FormControl = ({ onModelChange, options, model }: FormControlProps) => {
 								autoComplete="new-password"
 							/>
 						</Box>
+					)}
+
+					{options.type === 'datetime-local' && (
+						<>
+							<input
+								type="hidden"
+								name={options.name}
+								value={
+									!!model[options.name]
+										? moment(model[options.name] as string)
+												.utc()
+												.format()
+										: undefined
+								}
+							/>
+							<Input
+								mb={2}
+								type="datetime-local"
+								value={
+									!!model[options.name]
+										? (model[options.name] as string)
+										: undefined
+								}
+								onChange={(e: React.FormEvent<HTMLInputElement>) => {
+									console.log(e.currentTarget.value);
+									onModelChange({
+										[options.name]: e.currentTarget.value,
+									});
+								}}
+								onKeyPress={(e) => {
+									if (e.key === 'Enter' && !!model[options.name]) {
+										onModelChange({ [options.name]: undefined });
+									}
+								}}
+							/>
+						</>
 					)}
 
 					{(!options.type || options.type === 'text') && (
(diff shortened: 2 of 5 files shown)


src/unstable-temp/DownloadImageModal/FormModel.tsx
```diff
@@ -11,6 +12,8 @@ import { PasswordInput } from './PasswordInput';
 import styled from 'styled-components';
 import { DeviceTypeOptions, DeviceTypeOptionsGroup } from './models';
 
+const DEFAULT_FORMAT = 'yyyy[-]MM[-]DD';
```
myarmolinsky, line 15: `DATE_FORMAT` would probably be more appropriate here. `DEFAULT_FORMAT` does not tell me what it is a format for, and I also do not understand what `DEFAULT` is in reference to considering that there are no other format options available to choose from.

src/unstable-temp/DownloadImageModal/FormModel.tsx
```diff
@@ -173,6 +176,27 @@ const FormControl = ({ onModelChange, options, model }: FormControlProps) => {
...
+					{options.type === 'date' && (
+						<Input
+							mb={2}
+							type="date"
+							value={
+								!!model[options.name]
+									? moment(model[options.name] as string).format(DEFAULT_FORMAT)
+									: undefined
+							}
+							min={moment().utc().format(DEFAULT_FORMAT)}
```
myarmolinsky, line 188: Why do we use UTC? I'm guessing the expiry date is either for the beginning of the day in UTC? If so, we should specify to the user that the expiry date is in UTC (and specify end time such as midnight), i.e. `Will expire at 11:59pm UTC on the selected date` so that there's no confusion when the key expires earlier or later than expected if a user thought it was in their local time

nitishagar: Moved this to `datetime` with local time as discussed here: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/pm6XnWFz9EyRW1QkLQ7OOq8OzBt

nitishagar:
Updated Screenshot:
Screenshot 2022-05-14 at 10-15-55 STAGING balena dashboard Fleets

src/unstable-temp/DownloadImageModal/FormModel.tsx
```diff
@@ -173,6 +176,29 @@ const FormControl = ({ onModelChange, options, model }: FormControlProps) => {
 						</Box>
 					)}
 
+					{options.type === 'datetime-local' && (
+						<Input
+							mb={2}
+							type="datetime-local"
+							value={
```
thgreasi, line 183:
```suggestion
							name={options.name}
							value={
```
W/o this ^ no expiry date is set when downloading an image.

thgreasi (changes requested):

package.json
```diff
@@ -119,6 +119,7 @@
     "lodash": "^4.17.21",
     "memoizee": "^0.4.15",
     "mermaid": "^8.13.9",
+    "moment": "^2.29.3",
```
thgreasi, line 122: It would be nice if we could just rely on `date-fns` which we already use. Ideally the SDK & UI should move away from momentjs to something lighter...

thgreasi (commented):
In my tests the src/components/Renderer/widgets/widget-util.tsx change wasn't required for the provisioning api keys page to work as intended. Are you sure it's required?
In any case, let's move it to a separate PR since it's not related to the download image modal & not block this since other than that LGTM.

Feel free to self-certify after you split the formatTimestamp change.

package.json
nitishagar: Captured it here: https://github.com/balena-io-modules/rendition/issues/1608

src/components/Renderer/widgets/widget-util.tsx
```diff
@@ -82,7 +83,7 @@ export function formatTimestamp(
 	const uiFormat =
 		get(uiSchema, ['ui:options', 'dtFormat']) ||
 		`${DATE_FORMAT}, ${TIME_FORMAT}`;
-	return format(new Date(timestamp), uiFormat);
+	return format(new Date(moment(timestamp).local().toISOString()), uiFormat);
```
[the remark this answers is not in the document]
nitishagar: Moving it out and will test it again.

thgreasi (approved):

PR #1599, opened 2022-06-01
Remove Box from Card component to avoid nesting Box and Flex components

myarmolinsky:
Remove Box from Card component to avoid nesting Box and Flex components

Connects-to: https://github.com/balena-io/balena-ui/pull/5517
Connects-to: https://github.com/balena-io/balena-ui/pull/5500
Change-type: patch

---

##### Contributor checklist

##### Reviewer Guidelines

---


src/components/Card/index.tsx
```diff
@@ -41,25 +40,35 @@ const BaseCard = ({
...
+			>
 				{index > 0 && <Divider />}
 				{row}
-			</Box>
+			</Flex>
 		));
 
 	const Body = children && (
-		<Box fontSize={2} height="100%">
+		<Flex
```
thgreasi, line 56: This sounds like a major change, since in case someone was using `<span>`s or other inline elemenets, they will now show one per line, instead of next to each other.

src/components/Card/index.tsx
```diff
@@ -41,25 +40,35 @@ const BaseCard = ({
...
+			fontSize={2}
+			height="100%"
```
thgreasi, line 58: How about:
```suggestion
			flex="1"
```

src/components/Card/index.tsx
```diff
@@ -41,25 +40,35 @@ const BaseCard = ({
...
+			flexDirection="column"
+			flexWrap="wrap"
+			width="100%"
```
thgreasi, line 61: Should we instead of `width="100%"` elements, apply an `alingItems="stretch"` on the Wrapper on line 68?

myarmolinsky:
@balena-ci I self certify!

myarmolinsky:
@balena-ci retest

myarmolinsky:
@balena-ci retest

myarmolinsky:
@balena-ci republish

PR #1613, opened 2022-06-30
Add optional `badge` property to Sidebar MenuItem

myarmolinsky:
Add optional `badge` property to Sidebar MenuItem

Change-type: minor

---

##### Contributor checklist

##### Reviewer Guidelines

---


src/components/Sidebar/Menu.tsx
```diff
@@ -2,16 +2,28 @@ import React from 'react';
...
 
+const StyledBadge = styled(Badge)<{ color?: string }>`
+	border-radius: 0 1em 1em 0;
+	position: absolute;
+	transform: scale(0.75);
+	left: -5%;
+	bottom: 0;
+
+	${({ color }) => !!color && `background-color: ${color}; color: white`};
+`;
```
JSReds, line 25: Wouldn't be better to add the `color` prop inside the `bedge` component ?

JSReds: Also do we really need the absolute position ?

src/components/Sidebar/Menu.tsx
```diff
@@ -2,16 +2,28 @@ import React from 'react';
...
 import styled from 'styled-components';
+import { Badge, InternalBadgeProps } from '../Badge';
 import { Flex, FlexProps } from '../Flex';
 import { Txt } from '../Txt';
 
 export interface MenuItem {
 	className?: string;
 	href: string;
 	title: string;
+	badge?: Pick<InternalBadgeProps, 'children' | 'shade'> & { color: string };
```
JSReds, line 13: what is the reason of using Pick ? is `className` giving any type error ?

JSReds (commented):
I might be wrong but that seems the case: for next PR's when you see that snapshot generation is including also other changes (in this case probably renovate changes) I would split the commit with a secondary "update snapshots" or something similar

PR #1616, opened 2022-07-04
Fix missing json schema package build error

thgreasi:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/f-rendition/threads/vX8HYnTNYCHgAWMbkVOUFR0oA_e

### DIFF
--- a/src/components/Renderer/index.tsx
+++ b/src/components/Renderer/index.tsx
@@ -18,7 +18,8 @@ import {
 	WidgetMeta,
 } from './widgets';
 import { transformUiSchema, getBestSchemaMatch } from './widgets/widget-util';
-import { Value, JSONSchema, UiSchema, Format, JsonTypes } from './types';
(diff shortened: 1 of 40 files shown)


src/components/Renderer/index.tsx
```diff
@@ -18,7 +18,7 @@ import {
 	WidgetMeta,
 } from './widgets';
 import { transformUiSchema, getBestSchemaMatch } from './widgets/widget-util';
-import { Value, JSONSchema, UiSchema, Format, JsonTypes } from './types';
+import type { Value, JSONSchema, UiSchema, Format, JsonTypes } from './types';
```
JSReds, line 21: apparently, JsonTypes is an object and we use it as a value not as a type

thgreasi: Nice spot :+1:

thgreasi:
@balena-ci I self-certify!

PR #1621, opened 2022-07-14
Fix DownloadImageModal submit button being disabled when switching from a device type that has wifi network capability to one that does not

myarmolinsky:
Fix DownloadImageModal submit button being disabled when switching from a device type that has wifi network capability to one that does not

Change-type: patch

---

##### Contributor checklist

##### Reviewer Guidelines

---


src/unstable-temp/DownloadImageModal/ImageForm.tsx
```diff
@@ -95,12 +95,20 @@ const getDeviceTypeOptions = (t: TFunction, deviceType: DeviceType) => {
...
 	rawVersion: ImageFormProps['rawVersion'],
+	deviceType: DeviceType,
 ) => {
 	if (!rawVersion) {
 		return true;
 	}
 
-	return formModel.network === 'wifi' && !formModel.wifiSsid;
+	if (
+		(deviceType.options?.filter((option) => option.name === 'network') ?? [])
```
thgreasi, line 105: Could we instead try to clear the state properties that are not in the options any more?

myarmolinsky: If we clear the state properties of options that are not there anymore, that means we will also lose the input the user stored for wifi credentials when they go from wifi -> no wifi -> wifi (currently if I enter a network and password, then switch to a DT with no wifi, then switch back to a DT with wifi, the network and password I entered remain). All of this data on the `model` state. So we cannot just clear all state that is not relevant to the current DT because then if the user switches back and forth between DTs, we would be inconveniencing them by making them re-enter all the data they entered priorly. Perhaps we don't want the data persisting anyway though, what do you think about it all? @thgreasi

PR #1622, opened 2022-07-26
Refactor filters

JSReds:
Refactor filters and add refScheme support

Change-type: major
See: https://jel.ly.fish/thread-a72fb692-3732-4e3c-b93e-3c2044780d25

---

##### Contributor checklist

##### Reviewer Guidelines

---


src/components/DataTypes/index.ts
```diff
@@ -8,6 +8,14 @@ import * as objectType from './object';
...
 import * as stringType from './string';
 
+export interface DataTypeEditProps {
+	schema: JSONSchema;
+	value?: any;
+	onUpdate: (value: string | number | boolean) => void;
+	operator: string;
+	slim?: boolean;
+}
+
```
myarmolinsky, line 18: Is this still necessary? I don't see it being used anywhere

src/extra/AutoUI/Filters/PersistentFilters.tsx
```diff
@@ -44,27 +46,24 @@ const isQueryStringFilterRuleset = (
...
 				v: value,
+				r: refScheme,
 			}),
 		);
 	});
 	return qs.stringify(queryStringFilters);
 };
 
-const loadRulesFromUrl = (
+export const loadRulesFromUrl = (
```
myarmolinsky, line 66: Howcome we are exporting this now? We are not importing it anywhere

src/extra/AutoUI/Filters/PersistentFilters.tsx
```diff
@@ -44,27 +46,24 @@ const isQueryStringFilterRuleset = (
...
-		// TODO: The typings need to be fixed in rendition.
-		const flatSchema = flattenSchema(schema);
-		const flattenFilter = flattenSchema(filter);
-		const signatures = decodeFilter(
-			flatSchema,
-			flattenFilter,
-		) as FilterSignature[];
+export const listFilterQuery = (filters: JSONSchema[]) => {
+	const queryStringFilters = filters.map((filter) => {
+		const signatures = getSignatures(filter) as FilterSignature[];
```
myarmolinsky, line 51: Why is this cast necessary? Looks like `FiltersSignature[]` is already the return type of `getSignatures`

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,59 +1,72 @@
...
-) {
-	// Remove all schemas that may have been compiled already
-	ajv.removeSchema(/^.*$/);
-
-	const validators = Array.isArray(filters)
-		? filters.map((s) => ajv.compile(s))
-		: [ajv.compile(filters)];
+export const getSignatures = (filter: JSONSchema): FilterSignature[] => {
+	const signatures = filter.anyOf!.map((f: JSONSchema) => {
+		return f ? parseFilterDescription(f) : null;
```
myarmolinsky, line 35: How does this ternary work if `f` is always `JSONSchema`? What would make it falsy and if it's possible for it to be falsy, shouldn't we add that as a type of `f`?

src/extra/AutoUI/index.tsx
```diff
@@ -478,17 +497,17 @@ export const getColumnsFromSchema = <T extends AutoUIBaseResource<T>>({
 
 			const widgetSchema = { ...val, title: undefined };
 			return {
-				title: val.title || key,
+				title: getSchemaTitle(val, key, refScheme?.[0]),
 				field: key,
 				// This is used for storing columns and views
-				key,
+				key: `${key}_${index}`,
```
myarmolinsky, line 503: Why did we need to make this more unique?

src/components/Filters/spec.tsx
```diff
@@ -1,11 +1,18 @@
...
+	SchemaSieve,
+	Select,
+	Tag,
+	DropDownButton,
+} from '../../';
+import { Summary } from '../../components/Filters/Summary';
 import type { JSONSchema7 } from 'json-schema';
 import { Keyboard } from 'grommet';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+// import { 'svg' } from '@fortawesome/react-fontawesome';
```
myarmolinsky, line 15: comment

src/components/DataTypes/index.ts
JSReds: 🙏

src/extra/AutoUI/Filters/PersistentFilters.tsx
JSReds: Good catch, we re-write it in the UI, but I then decided to export all the persistent filters, so we can remove all that code from the dashboard as well. :pray:

src/extra/AutoUI/Filters/PersistentFilters.tsx
JSReds: nice catch! This was one of the multiple tries that I've done and I forgot to remove it

src/components/Filters/SchemaSieve.ts
JSReds: again my bad, this was another piece of code from another try. Thank you for pointing it out :)

src/extra/AutoUI/index.tsx
JSReds: This is because some properties may come from the same object, as we now use the refScheme as an array. eg:

```
is_for__device_type: {
					title: 'Device type',
					type: 'array',
					description:
						'{"x-ref-scheme": ["name", "is_of__cpu_architecture[0].slug"]}',
					items: {
						type: 'array',
						properties: {
							name: {
								title: 'Device type',
								format: 'device-type',
								type: 'string',
							},
							is_of__cpu_architecture: {
								type: 'array',
								items: {
									type: 'array',
									properties: {
										slug: {
											title: 'Architecture',
											type: 'string',
										},
									},
								},
							},
						},
					},
				},
```

In this case, the key would be for both items `is_for__device_type` which is not something we want for the filterModal select

src/components/Filters/spec.tsx
JSReds: 🙏

src/components/DataTypes/array.tsx
```diff
@@ -190,25 +143,114 @@ export const getFilter = (schema: JSONSchema, value: string) => {
...
+	) {
+		const filter = getModelFilter(
+			schema.items,
+			key,
+			field,
+			operator,
+			value,
+			isFullTextSearch,
+			convertOperator,
+		);
```
myarmolinsky, line 184: The difference between this filter and the one below is just `schema` vs `schema.items` right? Maybe we can have just 1 `const filter` above/outside this if and just determine which to use with the same if?

src/components/DataTypes/array.tsx
```diff
@@ -190,25 +143,114 @@ export const getFilter = (schema: JSONSchema, value: string) => {
...
+							);
+							props.push(filter);
+							return props;
+						},
+						[],
+					),
+				],
+			},
+		};
+	} else if (
```
myarmolinsky, line 171: We don't need this else since we're returning in the if, right?

src/components/DataTypes/array.tsx
```diff
@@ -190,25 +143,114 @@ export const getFilter = (schema: JSONSchema, value: string) => {
...
+		) as JSONSchema['properties'];
+		return property as JSONSchema;
+	}
+	const subSchema = schemaItems || schema;
+	const model = getDataModel(subSchema);
+	return (
+		model && 'editSchema' in model
+			? model.editSchema(subSchema, operator)
+			: subSchema
+	) as JSONSchema;
```
myarmolinsky, line 255: Looks like we only need to cast the `model.editSchema` part, not the entire ternary

src/components/DataTypes/array.tsx
JSReds: this is one of the exceptions I don't like... That's not the same thing as the one we receive from above. this one will return a contain with a filter, the other one return an items, anyof filter

src/components/DataTypes/date-time.tsx
```diff
@@ -30,64 +26,36 @@ interface DateTimeFilter extends JSONSchema {
...
+					refScheme,
+				),
+				type: 'object',
+		  }
+		: ({} as DateTimeFilter);
 
 	const normalizedValue = normalizeDateTime(value);
 
-	if (operator === 'is') {
+	if (operatorSlug === 'is') {
```
myarmolinsky, line 58: Looks like we can just do the following, no?:
```
switch(operatorSlug) {
  case 'is':
  case 'is_before':
  case 'is_after':
    return Object.assign(base, {
                 properties: {
                   [field]: {
	             type: 'string',
		     format: 'date-time',
		     const: normalizedValue,
		  },
		},
		required: [field]
    }
    default:
      return base;
}
```

myarmolinsky: I know this was pre-existing, but since we're refactoring and it appears the simplification can be made
I also assume we can make this simplification can be made in all DataTypes

src/components/DataTypes/enum.tsx
```diff
@@ -1,17 +1,15 @@
...
-		getLabel: (_s: JSONSchema) => 'is',
-	},
-	is_not: {
-		getLabel: (_s: JSONSchema) => 'is not',
-	},
-};
+export const operators = (_s: JSONSchema) => ({
+	is: 'is',
+	is_not: 'is not',
+});
```
myarmolinsky, line 10: Should we make an `operators` enum in a common file and just import the parts of that we need for each `operators` we create per DataType instead of manually re-entering each operator?

i.e. something like this instead of what we have:
```
export const operators = (_s: JSONSchema) => ([OPERATORS['is'], OPERATORS['is_not']].toObject())
```

Forgive the pseudocode but hopefully the above gets the idea across ^

src/components/DataTypes/date-time.tsx
JSReds: I would avoid doing this, since it is only applicable to some DataTypes, others have multiple rules within the if or different rules, which does not allow us to use the switch case and would create a difference between files. I find it easier to read if they all stay consistent

src/components/DataTypes/enum.tsx
JSReds: This was my thought as well, but this is not double because of the Object type. We need all dataTypes to have the same exact method to get operators.  Until we have the object exception for tags, we will have to stick with this solution unfortunatelly

src/components/DataTypes/date-time.tsx
myarmolinsky: But in the case of the DataTypes with multiple rules, we could have multiple different cases in the switch instead of stacking them as above @JSReds 

It stays consistent in the sense that we're using `switch` for it everywhere and we also get the reduce LOC. Honestly I don't think have multiple identical if statements everywhere is clean. If I see separate if statements, I start hunting for the different between them.

src/components/DataTypes/object.tsx
```diff
@@ -210,31 +113,88 @@ function getValueForOperation(
...
+		return schema.title;
+	}
+	const property = schema.properties[Object.keys(value)[0]];
+	if (keySpecificOperators.includes(operator)) {
+		return typeof property === 'object' ? property.title : schema.title;
+	}
+	if (valueSpecificOperators.includes(operator)) {
+		return typeof property === 'object' ? property.title : schema.title;
+	}
+	return schema.title;
```
myarmolinsky, line 131:
```suggestion
	if (typeof value === 'object' && schema.properties) {
	  const property = schema.properties[Object.keys(value)[0]];
	  if ((keySpecificOperators.includes(operator) || valueSpecificOperators.includes(operator)) && typeof property === 'object') {
		  return property.title;
	  }
	}
	return schema.title;
```

src/components/DataTypes/date-time.tsx
myarmolinsky: Looking more I see that date-time is a special case that happens to have all operators returning the same thing, but even in that case I would still at least dump all the conditions into 1 if statement rather than having 3 identical if statements

src/components/DataTypes/oneOf.tsx
```diff
@@ -35,62 +22,28 @@ interface OneOfFilter extends JSONSchema {
...
-		$id: randomString(),
-		title: operator,
-		description: getJsonDescription(
-			title || field,
-			operators[operator].getLabel(schema),
-			getCaption(value, schema) || '',
-		),
-		type: 'object',
-	};
+	const operatorSlug = operator.slug as OperatorSlug;
```
myarmolinsky, line 35: Looks like `OperatorSlug` is always a string. Do we need to be creating the `type` in every file and setting it to the resolvable return type of the `operators` function? I think we can just use `string` as the type everywhere

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,294 @@
...
+						<FilterInputs
+							properties={properties || []}
+							signature={signature}
+							schema={schema}
+							onUpdate={(signature) => addSignature(signature, index)}
+						/>
 						{index > 0 && (
-							<DeleteButton
+							<Button
+								color="rgba(0, 0, 0, 0.4)"
```
myarmolinsky, line 135: Where is the color from? Is it in our styleguide?

src/components/DataTypes/object.tsx
JSReds: nice catch

src/components/DataTypes/oneOf.tsx
JSReds: it is true that operatorSlug turns out to be a string, but by checking the type in my opinion it is clearer that it must be an operators key. string is so generic IMHO. if you think it is not understandable anyway, I can remove it and put only string.

src/components/DataTypes/date-time.tsx
JSReds: > But in the case of the DataTypes with multiple rules, we could have multiple different cases in the switch instead of stacking them as above

how would you use a switch case in this case ?  

```
	if (
		operatorSlug === 'not_contains' &&
		!!schema.items &&
		typeof schema.items === 'object' &&
		'type' in schema.items &&
		!('properties' in schema.items)
	) {
```

Also cases like :

```
		operatorSlug === 'key_matches_re' ||
		operatorSlug === 'value_matches_re'

```

in the case of the object we would add many lines of code, and from the benefits that a switch case gives me it doesn't seem worth it. That said, it is my opinion, I can change the ones that can be changed if we all agree.

cons:

- array without switch case (inconsistent)
- object doubles the lines of code of the `createFilter` function (worth it ?)

src/components/Filters/FilterModal.tsx
JSReds: that was a custom colour added 4 years ago on that button. It can effectively become `tertiary.main` which is similar.

src/components/DataTypes/date-time.tsx
myarmolinsky: For the first case you mention, a switch would indeed be inappropriate
Wouldn't the second case just be this?:
```
switch(operatorSlug) {
  case 'key_matches_re':
  case 'value_matches_re':
    ...
}
```

Anyway I hear your point about the inconsistency of using a switch in some files but not in others (per the first case)
Do you feel like having 1 if statement like this would be inappropriate?:
```
if (['is', 'is_before', 'is_after'].includes(operatorSlug)) {
  ...
}
```

That manages to avoid the inconsistency of a switch in just 1 file while reducing LOC per a repetitive case

src/components/DataTypes/oneOf.tsx
myarmolinsky: I hear your point, you're trying to establish the connection between the `operatorSlug` and the `operators`. To be honest I wouldn't create that extra type which is effectively just a `string` because seeing the extra type (as an outside observing the code) makes me think that we're expecting something other than a type we already have clear access to (i.e. `string | { value: string, id: number } | undefined`). That's what made this bit of code catch my eye was me wondering "it seemed like this just returns strings, what else is it expected to return?" I'd argue it makes things a little more confusing/convoluted

To be clear, if the typing ended up saying _exactly the keys expected in per case_, such as `'is' | 'is_not'` instead of just the generic `string` type, it would be 100% understandable to me and cool with me to have this extra `OperatorSlug` type created. So this would work in my opinion:
```
export const operators = () => ({
	is: 'is',
	is_not: 'is not',
} as const);

const operatorsObj = operators();

type OperatorSlug = keyof typeof operatorsObj;
```

JSReds: that would be a solution for sure, but this would not work for Object.ts case which is 

export const operators = (s: JSONSchema) => {

and operatorsObj cannot be called since `schema` is not defined. 

@thgreasi would you have any other options in mind ?

src/components/DataTypes/date-time.tsx
JSReds: I got lost here on this last piece of information. I probably don't remember the case any more, could you please remind me how this could reduce the lines of code ?
The 3 cases `['is', 'is_before', 'is_after']` are different. We could unify them using ternaries, but since I'm not 100% sure we listed all cases and that all properties in each case are the right one, I would keep them separated for now.

JSReds:
@balena-ci retest

src/components/DataTypes/date-time.tsx
myarmolinsky:
```
if (operatorSlug === 'is') {

      
      


      
            
      
  
        
        

        
      

          

  
    
        
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    

  

  


    

    
        myarmolinsky
  

    

    

    


      on Aug 1

    
  •

  
    
      
        
          edited
          
        
        
    

      
    
    
      
        
  
  

      
    
  


  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      Looks like we can just do the following, no?:
switch(operatorSlug) {
  case 'is':
  case 'is_before':
  case 'is_after':
    return Object.assign(base, {
                 properties: {
                   [field]: {
	             type: 'string',
		     format: 'date-time',
		     const: normalizedValue,
		  },
		},
		required: [field]
    }
  default:
    return base;
}

    
      
    

      
    

    
  
    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    



  
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    

  

  


    

    
        myarmolinsky
  

    

    

    


      on Aug 1

    
  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      I know this was pre-existing, but since we're refactoring and it appears the simplification can be made
I also assume we can make this simplification can be made in all DataTypes
    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    



  
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    Author

  

  


    

    
        JSReds
  

    

    

    


      on Aug 2

    
  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      I would avoid doing this, since it is only applicable to some DataTypes, others have multiple rules within the if or different rules, which does not allow us to use the switch case and would create a difference between files. I find it easier to read if they all stay consistent
    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    



  
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    

  

  


    

    
        myarmolinsky
  

    

    

    


      on Aug 2

    
  •

  
    
      
        
          edited
          
        
        
    

      
    
    
      
        
  
  

      
    
  


  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      But in the case of the DataTypes with multiple rules, we could have multiple different cases in the switch instead of stacking them as above @JSReds
It stays consistent in the sense that we're using switch for it everywhere and we also get the reduced LOC. Honestly I don't think having multiple identical if statements everywhere is clean. If I see separate if statements, I start hunting for the difference between them.
    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    



  
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    

  

  


    

    
        myarmolinsky
  

    

    

    


      on Aug 2

    
  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      Looking more I see that date-time is a special case that happens to have all operators returning the same thing, but even in that case I would still at least dump all the conditions into 1 if statement rather than having 3 identical if statements
    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    





  
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    Author

  

  


    

    
        JSReds
  

    

    

    


      29 days ago

    
  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      
But in the case of the DataTypes with multiple rules, we could have multiple different cases in the switch instead of stacking them as above

how would you use a switch case in this case ?
	if (
		operatorSlug === 'not_contains' &&
		!!schema.items &&
		typeof schema.items === 'object' &&
		'type' in schema.items &&
		!('properties' in schema.items)
	) {

    
      
    

      
    

    
  
Also cases like :
		operatorSlug === 'key_matches_re' ||
		operatorSlug === 'value_matches_re'


    
      
    

      
    

    
  
in the case of the object we would add many lines of code, and from the benefits that a switch case gives me it doesn't seem worth it. That said, it is my opinion, I can change the ones that can be changed if we all agree.
cons:

array without switch case (inconsistent)
object doubles the lines of code of the createFilter function (worth it ?)

    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    



  
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    

  

  


    

    
        myarmolinsky
  

    

    

    


      29 days ago

    
  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      For the first case you mention, a switch would indeed be inappropriate
Wouldn't the second case just be this?:
switch(operatorSlug) {
  case 'key_matches_re':
  case 'value_matches_re':
    ...
}

    
      
    

      
    

    
  
Anyway I hear your point about the inconsistency of using a switch in some files but not in others (per the first case)
Do you feel like having 1 if statement like this would be inappropriate?:
if (['is', 'is_before', 'is_after'].includes(operatorSlug)) {
  ...
}

    
      
    

      
    

    
  
That manages to avoid the inconsistency of a switch in just 1 file while reducing LOC per a repetitive case
    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    



  
        
  
      
          
          
    

      
    
    
        
                👍
        
        
                👎
        
        
                😄
        
        
                🎉
        
        
                😕
        
        
                ❤️
        
        
                🚀
        
        
                👀
        
    

  

      

    
            
    


      
          
            
              
  
  

            
            
              
    

              Sorry, something went wrong.
            
            
             Quote reply
            
          
      
    
  

  
    

      
    Member


    Author

  

  


    

    
        JSReds
  

    

    

    


      21 hours ago

    
  


        
      
  
  
  
  
    
      
    

    

    There was a problem hiding this comment.


  

 
  
  
    
      
    

    

    Choose a reason for hiding this comment

    
      The reason will be displayed to describe this comment to others. Learn more.
    

    
      
        
        Choose a reason
        
        Spam
Abuse
Off Topic
Outdated
Duplicate
Resolved
      
            Hide comment


  


  
    
      I got lost here on this last piece of information. I probably don't remember the case any more, could you please remind me how this could reduce the lines of code ?
The 3 cases ['is', 'is_before', 'is_after'] are different. We could unify them using ternaries, but since I'm not 100% sure we listed all cases and that all properties in each case are the right one, I would keep them separated for now.
    
  
            



  
    
    Refresh and try again.
  




    

        
      
  
    
  
  

  
  
    
    

    Sorry, something went wrong.
  



        
      
  
    
    
      
        All reactions

            
          

    


    




    

      
        
          
            
              
                
              
              
                
                  Reply...
                
              
            
          

          
            
              
              
              
              
              
              
              

              
  
  
      
        
          Write
        
        
          Preview
        
      
    

        
        
  
      
    

  Add a suggestion, <Ctrl+g>


      


  
    
      
    

    
    Add heading text

    
      
    

    
    Add bold text, <Ctrl+b>

    
      
    

    
    Add italic text, <Ctrl+i>
  

  
    
      
    

    
    Add a quote, <Ctrl+Shift+.>

    
      
    

    
    Add code, <Ctrl+e>



    
      
    

    
    Add a link, <Ctrl+k>
  

  
    
      
    

    
    Add a bulleted list, <Ctrl+Shift+8>

    
      
    

    
    Add a numbered list, <Ctrl+Shift+7>

    
      
    

    
    Add a task list, <Ctrl+Shift+l>
  

        

          
            
    

          
          Directly mention a user or team


          
            
    

          
          Reference an issue, pull request, or discussion

          
  
    
    

  

  
    
      Select a reply
      ctrl .
    

    
      
  
  

    

  

Add saved reply


      



  

  
    There was an error creating your PullRequest.
  

  
    
        

      
          
          

      
          
    
    
    
      
        Attach files by dragging & dropping, selecting or pasting them.
      
      
        
  
  

          
            Uploading your files…
          
      
      
        We don’t support that file type.
        
          Try again with a
          GIF, JPEG, JPG, MOV, MP4, PNG, SVG, WEBM, CSV, DOCX, FODG, FODP, FODS, FODT, GZ, LOG, MD, ODF, ODG, ODP, ODS, ODT, PDF, PPTX, TGZ, TXT, XLS, XLSX or ZIP.
        
      
      
        Attaching documents requires write permission to this repository.
        
          Try again with a GIF, JPEG, JPG, MOV, MP4, PNG, SVG, WEBM, CSV, DOCX, FODG, FODP, FODS, FODT, GZ, LOG, MD, ODF, ODG, ODP, ODS, ODT, PDF, PPTX, TGZ, TXT, XLS, XLSX or ZIP.
        
      
      
        We don’t support that file type.
        
          Try again with a GIF, JPEG, JPG, MOV, MP4, PNG, SVG, WEBM, CSV, DOCX, FODG, FODP, FODS, FODT, GZ, LOG, MD, ODF, ODG, ODP, ODS, ODT, PDF, PPTX, TGZ, TXT, XLS, XLSX or ZIP.
        
      
      
      
      
        This file is empty.
        
          Try again with a file that’s not empty.
        
      
      
        This file is hidden.
        
          Try again with another file.
        
      
      
        Something went really wrong, and we can’t process that file.
        
          Try again.
        
      
    
    
      
    

    
    Styling with Markdown is supported
  

    

  
    
    
    
    
    
    
    
    
    
    
  
    Nothing to preview
  


  


  



              
                
                          Add review comment
                      Start a review

                          Add single comment

                

                      Cancel

              
          
        
      
  

    
      
            Resolve conversation






    

      
  
		return Object.assign(base, {
			properties: {
				[field]: {
					type: 'string',
					format: 'date-time',
					const: normalizedValue,
				},
			},
			required: [field],
		});
	}

	if (operatorSlug === 'is_before') {
		return Object.assign(base, {
			properties: {
				[field]: {
					type: 'string',
					format: 'date-time',
					formatMaximum: normalizedValue,
				},
			},
			required: [field],
		});
	}

	if (operatorSlug === 'is_after') {
		return Object.assign(base, {
			properties: {
				[field]: {
					type: 'string',
					format: 'date-time',
					formatMinimum: normalizedValue,
				},
			},
			required: [field],
		});
	}
```
It looks like they have the exact same content in the if statements? @JSReds

myarmolinsky: Argh I see that they're different know 🤦 
They looked the same but they have `const` vs `formatMaximum` vs `formatMinimum` in them 🤦 🤦 🤦 
*sigh* apologies for this huge mess 😞 , the code looks good

src/components/DataTypes/string.tsx
```diff
@@ -57,103 +44,39 @@ interface StringFilter extends JSONSchema {
...
-		$id: randomString(),
-		title: operator,
-		description: getJsonDescription(
-			title || field,
-			operators[operator].getLabel(schema),
-			value,
-		),
-		type: 'object',
-	};
+	const operatorSlug = operator.slug as OperatorSlug;
```
thgreasi, line 56: Do you really need this `as OperatorSlug`?

src/components/DataTypes/array.tsx
```diff
@@ -49,68 +45,44 @@ interface StringFilter extends JSONSchema {
...
+	) {
 		return Object.assign(base, {
 			anyOf: [
 				{
 					properties: {
 						[field]: {
 							type: 'array',
-							not: {
-								...filter,
+							contains: {
```
thgreasi, line 82: Does that mean that if the array is empty then not_contains will always be returning false?
If so, I think that's wrong, right?

src/components/DataTypes/array.tsx
```diff
@@ -49,68 +45,44 @@ interface StringFilter extends JSONSchema {
...
 
-	if (operator === 'not_contains') {
+	const filter = getFilter(field, schema, value, operator, isFullTextSearch);
+	if (
+		operatorSlug === 'not_contains' &&
+		!!schema.items &&
+		typeof schema.items === 'object' &&
+		'type' in schema.items &&
+		!('properties' in schema.items)
+	) {
```
thgreasi, line 75: Does that need a `// TODO: support arrays of objects`?

src/components/DataTypes/array.tsx
```diff
@@ -126,61 +98,24 @@ export const createFilter = (
...
-
-	return (
-		<Textarea
-			onChange={(e: React.FormEvent<HTMLTextAreaElement>) =>
-				onUpdate(e.currentTarget.value)
-			}
-			{...props}
-		/>
-	);
+	return Object.assign(base, {
```
thgreasi, line 101: Should this be inside an
```
	if (
		operatorSlug === 'contains' &&
		!!schema.items &&
		typeof schema.items === 'object' &&
		'type' in schema.items &&
		!('properties' in schema.items)
	) {
```
and at the end just do `return base;` if it's an unknown operator?

src/components/DataTypes/object.tsx
```diff
@@ -21,75 +20,59 @@ const getValueLabel = (schema: JSONSchema) => {
...
+					key_is: `${getKeyLabel(s)} is`,
+					key_matches_re: `${getKeyLabel(s)} matches RegEx`,
+					key_not_matches_re: `${getKeyLabel(s)} does not match RegEx`,
+					value_is: `${getValueLabel(s)} is`,
+					value_contains: `${getValueLabel(s)} contains`,
+					value_not_contains: `${getValueLabel(s)} does not contain`,
+					value_matches_re: `${getValueLabel(s)} matches RegEx`,
+					value_not_matches_re: `${getValueLabel(s)} does not match RegEx`,
+			  }
+			: {}),
```
thgreasi, line 54:
```suggestion
		...(!isKeyValueObj(s)
			? {
					contains: 'contains',
					not_contains: 'does not contain',
			  }
			: (() => {
				const keyLabel = getKeyLabel(s);
				const valueLabel = getValueLabel(s);
				return {
					key_contains: `${keyLabel} contains`,
					key_not_contains: `${keyLabel} does not contain`,
					key_is: `${keyLabel} is`,
					key_matches_re: `${keyLabel} matches RegEx`,
					key_not_matches_re: `${keyLabel} does not match RegEx`,

					value_is: `${valueLabel} is`,
					value_contains: `${valueLabel} contains`,
					value_not_contains: `${valueLabel} does not contain`,
					value_matches_re: `${valueLabel} matches RegEx`,
					value_not_matches_re: `${valueLabel} does not match RegEx`,
				};
			  })()),
```

src/components/DataTypes/array.tsx
```diff
@@ -49,68 +45,44 @@ interface StringFilter extends JSONSchema {
...
 		return Object.assign(base, {
 			anyOf: [
 				{
 					properties: {
 						[field]: {
 							type: 'array',
-							not: {
-								...filter,
+							contains: {
+								not: {
```
thgreasi, line 88:
```suggestion
							not: {
								contains: {
```
See: https://www.jsonschemavalidator.net/s/47xdkAWd

JSReds: with ajv 6.12 contains.not seems to be the right one https://codesandbox.io/s/ajv-playground-forked-c4lk06?file=/src/index.js

src/components/Filters/SchemaSieve.ts
```diff
@@ -1,59 +1,105 @@
...
+			return {};
+		}
+		return (model.createFilter as CreateFilter)(
+			field,
+			operator,
+			value,
+			subSchema,
+			false,
+			refScheme,
+		);
```
thgreasi, line 95: Should we v ?
* Remove the `recursive` argument from `getBase`
* Stop calling getBase from within the DataTypes, in favor of calling it only from here.

```suggestion
		return {
			...getBase(
				// recursive, // drop me
				operator,
				key,
				title,
				field,
				value,
				refScheme,
			),
			...(model.createFilter as CreateFilter)(
				field,
				operator,
				value,
				subSchema,
				false,
				refScheme,
			),
		};
```
And assuming that this will be the only place that `getBase` would be called, we could even inline it here.

src/components/DataTypes/array.tsx
```diff
@@ -49,166 +50,132 @@ interface StringFilter extends JSONSchema {
...
+				const operators = model?.operators(schema);
+				const newOperator = convertOperator(operator, operators);
+				const filter = (model?.createFilter as CreateFilter)(
+					propKey,
+					newOperator,
+					value,
+					subSchema,
+					true,
+				);
+				return filter;
```
thgreasi, line 112: At this point, isn't this just v ?
```suggestion
				return getModelFilter(
					propValue,
					propKey,
					operator,
					value,
					isFullTextSearch,
					convertOperator,
				);
```

src/components/Filters/SchemaSieve.ts
```diff
@@ -112,303 +168,106 @@ const generateFullTextSearchAjvFilter = (
...
+	operatorConverter?: (
+		operator: Operator<OperatorSlugs>,
+		operators: any,
+	) => Operator<OperatorSlugs>,
+): JSONSchema => {
+	let newOperator = operator;
+	const subSchema = isFullTextSearch
+		? { ...schema, oneOf: undefined, enum: undefined }
+		: schema;
+	const model = getDataModel(subSchema);
```
thgreasi, line 260: Should we add an `if (model == null) { return;/throw }` rather than using `?.` and casting the potentially undefined result?

src/components/Filters/SchemaSieve.ts
```diff
@@ -112,303 +168,106 @@ const generateFullTextSearchAjvFilter = (
...
+	isFullTextSearch?: boolean,
+	operatorConverter?: (
+		operator: Operator<OperatorSlugs>,
+		operators: any,
+	) => Operator<OperatorSlugs>,
+): JSONSchema => {
+	let newOperator = operator;
+	const subSchema = isFullTextSearch
+		? { ...schema, oneOf: undefined, enum: undefined }
+		: schema;
```
thgreasi, line 259: Should we try to keep the object a bit cleaner of props?
```suggestion
	let subSchema = schema;
	if (isFullTextSearch) {
		subSchema = {...schema};
		delete subSchema.oneOf;
		delete subSchema.enum;
	}
```

src/components/Filters/FilterDescription.tsx
```diff
@@ -17,12 +17,22 @@ const parseFilter = (filter: JSONSchema | null | undefined) => {
 };
 
 const FilterDescription = ({ filter, ...props }: FilterDescriptionProps) => {
-	const tagProps = filter.anyOf
+	const tagProps = !!filter.anyOf
```
thgreasi, line 20: I guess the `!!` isn't necessary :)
```suggestion
	const tagProps = filter.anyOf
```

src/extra/AutoUI/models/helpers.ts
```diff
@@ -136,12 +147,87 @@ export const autoUIAddToSchema = (
...
+export const getPropertyScheme = (
 	schemaValue: JSONSchema | JSONSchemaDefinition,
 ) => {
 	return typeof schemaValue !== 'boolean' &&
 		!!schemaValue.description &&
 		isJson(schemaValue.description)
-		? JSON.parse(schemaValue.description!)['x-ref-scheme']
+		? JSON.parse(schemaValue.description!)['x-foreign-key-scheme'] ??
+				JSON.parse(schemaValue.description!)['x-ref-scheme']
 		: null;
```
thgreasi, line 158: Let's avoid parsing the JSON so many times:
```suggestion
	if (typeof schemaValue === 'boolean' || !schemaValue.description) {
		return null;
	}
	try {
		const json = JSON.parse(schemaValue.description!);
		return json['x-foreign-key-scheme'] ?? json['x-ref-scheme']
	} catch (err) {
		return null;
	}
```

src/components/Filters/SchemaSieve.spec.ts
```diff
@@ -782,28 +802,34 @@ describe('SchemaSieve', () => {
...
-					},
-					'Entry 5': {
-						foo: 'bar',
-					},
+					// 'Entry 4': {
+					// 	category: null,
+					// },
+					// 'Entry 5': {
+					// 	foo: 'bar',
+					// },
```
thgreasi, line 810: Should we re-enable these and leave a comment explaining that not_contains needs the filtered prop to exist and be a string at the first place?

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
-
-	const setEditField = (field: string, index: number) => {
-		const currentEdit = filters.map((filter, i) =>
-			i === index ? SchemaSieve.getCleanEditModel(schema, field) : filter,
+	const properties: PropertyOption[] | undefined = React.useMemo(() => {
+		if (!schema.properties || typeof schema.properties === 'boolean') {
+			return;
+		}
+		const constructedProperties: PropertyOption[] = [];
+		Object.entries(schema.properties).forEach(
```
thgreasi, line 51: I guess it can become a `.flatMap()`.

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
+	const newSignature = {
+		key: !!properties?.[0].refScheme
+			? `${properties?.[0].field}__${properties?.[0].refScheme}`
+			: properties?.[0].field,
+		title: properties?.[0].title,
+		field: properties?.[0].field,
+		operator: undefined,
+		value: undefined,
+		refScheme: properties?.[0].refScheme,
 	};
```
thgreasi, line 84: Would it make more sense to just `return null;` or show a spinner in case `properties?.[0] == null`?
Is a `newSignature` with undefined props useful?

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
-		);
-		setFilters(currentEdit);
+	const addSignature = (signature: Partial<FilterSignature>, index: number) => {
+		if (!signatures) {
+			return;
+		}
+		if (!signatures.length) {
+			setSignatures([{ ...signature }]);
+		} else if (signatures.length >= index + 1) {
+			signatures[index] = signature;
```
thgreasi, line 97: Should we destructure it here, like right above?
```suggestion
			signatures[index] = {...signature};
```
Or in the other way around, do we really need the destructing in `setSignatures([{ ...signature }]);`?

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
+	};
 
-	const filteredFieldOptions = React.useMemo(() => {
-		if (!searchTerm) {
-			return fieldOptions;
+	const createAndAddFilter = () => {
+		const isInvalid = signatures?.some(
+			(s) => !s.field || !s.operator || !s.value,
+		);
+		if (!!isInvalid) {
```
thgreasi, line 113:
```suggestion
		if (isInvalid) {
```

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
-		const searchTermRegEx = new RegExp(searchTerm, 'i');
-		return fieldOptions.filter((option) => {
-			return (
-				option.title.match(searchTermRegEx) ||
-				option.field.match(searchTermRegEx)
-			);
-		});
-	}, [searchTerm, fieldOptions]);
+		const filter = createFilter(schema, signatures as FilterSignature[]);
+		return !!filter ? addFilter(filter) : onClose();
```
thgreasi, line 117:
```suggestion
		return filter ? addFilter(filter) : onClose();
```

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
+const FilterInputs = ({
+	properties,
+	signature,
+	schema,
+	onUpdate,
+}: FilterInputs) => {
+	const [searchTerm, setSearchTerm] = React.useState('');
+	const debouncedSetSearchTerm = debounce(setSearchTerm, 300);
+	const operators = React.useMemo(
+		() => (!!signature?.field ? getOperators(schema, signature.field) : []),
```
thgreasi, line 181:
```suggestion
		() => (signature?.field ? getOperators(schema, signature.field) : []),
```

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
+		[signature?.field, signature?.refScheme],
```
thgreasi, line 182: Does the `signature?.refScheme` need to be here? B/c I don't see it being used in the useMemo body.
If it does, then let's leave a comment so that we don't forget why.
Does this need to be v ?
```suggestion
		[signature?.field, schema],
```

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
+	const removeSignature = (index: number) => {
+		const newSignatures = signatures.filter((_f, i) => index !== i);
+		setSignatures(newSignatures);
+	};
 
-	const filteredFieldOptions = React.useMemo(() => {
-		if (!searchTerm) {
-			return fieldOptions;
+	const createAndAddFilter = () => {
+		const isInvalid = signatures?.some(
```
thgreasi, line 110:
```suggestion
		const isInvalid = signatures.some(
```

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
-								<FilterInput
-									operator={operator}
-									value={value}
-									schema={schema.properties![field] as JSONSchema}
-									onUpdate={(v: any) => setEditValue(v, index)}
-								/>
-							</Flex>
-						</Flex>
+			{signatures.map((signature, index) => (
+				<Flex flexDirection="column" key={(signature?.field || '') + index}>
```
thgreasi, line 128:
```suggestion
				<Flex flexDirection="column" key={(signature.field || '') + index}>
```

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
+			))}
 			<Button
 				mb={2}
 				mt={4}
 				primary
 				underline
-				onClick={() =>
-					setFilters(filters.concat(SchemaSieve.getCleanEditModel(schema)))
-				}
+				onClick={() => addSignature(newSignature, signatures!.length)}
```
thgreasi, line 157:
```suggestion
				onClick={() => addSignature(newSignature, signatures.length)}
```

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
 			>
 				Add alternative
 			</Button>
 		</Modal>
 	);
 };
+
+interface FilterInputs {
+	properties: PropertyOption[];
+	signature?: Partial<FilterSignature>;
```
thgreasi, line 167: I think that we never pass `undefined` as a value for this
```suggestion
	signature: Partial<FilterSignature>;
```
which means we can drop a few `!` and `?.` from the `FilterInputs`.

src/components/Filters/FilterModal.tsx
```diff
@@ -1,243 +1,296 @@
...
+}
+
+const FilterInput = ({
+	schema,
+	value,
+	operator,
+	onUpdate,
+	refScheme,
+}: FilterInputProps) => {
+	const subSchema = !!refScheme
```
thgreasi, line 273:
```suggestion
	const subSchema = refScheme
```

thgreasi:
Self-certify :clap:

thgreasi (approved):

PR #1624, opened 2022-08-01
Remove "Source Sans Pro"

Page-:
It renders at a smaller size than the other fonts for me and when increasing the font size to match other fonts it still renders with somewhat weird/blurry looking anti-aliasing that the others don't have, and since I don't know the reasoning for using this specific font I decided to remove it since the alternatives seem to be win-win? Let me know if there's a good reason to keep using this specific font/what advantages it has?

### DIFF
--- a/src/components/Renderer/__snapshots__/spec.tsx.snap
+++ b/src/components/Renderer/__snapshots__/spec.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`Renderer component A Heading widget 1`] = `
 .c0 {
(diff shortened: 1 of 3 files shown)


iamsolankiamit:
@Page- we chose Source sans pro as it was used across all our designs. if we want to move away from this font we should maybe ask Jon for alternatives or use system default font sets which looks good on windows/mac/linux/android/iOs. The current defaults are not widely available except for san-serif.

LucianBuzzo (commented):
.

PR #1633, opened 2022-09-20
React 17

thgreasi:
Connects-to: # 
See: <url> 
Depends-on: <url> 

---

##### Contributor checklist

##### Reviewer Guidelines

---


src/components/Notifications/index.tsx
```diff
@@ -54,7 +54,7 @@ const generateNotificationId = () => {
 
 const getTransformedOptions = (
 	options: React.ReactNode | NotificationOptions,
-): NotificationOptions => {
```
myarmolinsky, line 57: Why did we remove this return typing?

thgreasi: @myarmolinsky B/c the inferred return type has the `id` as non-nullable, while NotificationOptions have it as nullable, which made ts complain during an assignment later in the file.

myarmolinsky: @thgreasi Yes I saw that, but `getTransformedOptions` sets some default values. It's possible for them to be overriden if the user explicitly passes undefined for them but since the values are required it makes sense for us to just use `??` to ensure the value is always defined even if the user explicitly passes `undefined`

thgreasi: Ok ok fine....

JSReds: really strange one btw, TS is not considering the `id: undefined` case...  checking TS issues....

thgreasi:
@balena-ci republish

Issue #1608, opened 2022-06-23
Move away from `moment` to `date-fns` 


nitishagar:
> date-fns which we already use. Ideally the SDK & UI should move away from momentjs to something lighter
@thgreasi feedback here: https://github.com/balena-io-modules/rendition/pull/1587

PR #1651, opened 2022-09-23
Remove moment in favor of date-fns

myarmolinsky:
Remove moment in favor of date-fns

Resolves: #1608 
Change-type: patch

---

##### Contributor checklist

##### Reviewer Guidelines

---

### DIFF
--- a/src/unstable-temp/DownloadImageModal/FormModel.tsx
+++ b/src/unstable-temp/DownloadImageModal/FormModel.tsx
@@ -13,7 +13,6 @@ import styled from 'styled-components';
 import { DeviceTypeOptions, DeviceTypeOptionsGroup } from './models';
 import { DocsLink } from './OsConfiguration';
 import { useTranslation } from '../../hooks/useTranslation';
-import moment from 'moment';
 
 export const DownloadImageLabel = styled.label`
 	display: flex;
@@ -208,9 +207,7 @@ const FormControl = ({ onModelChange, options, model }: FormControlProps) => {
 								name={options.name}
 								value={
 									!!model[options.name]
-										? moment(model[options.name] as string)
-												.utc()
-												.format()
+										? new Date(model[options.name] as string).toISOString()
 										: undefined
 								}
 							/>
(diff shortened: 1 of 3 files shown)


myarmolinsky:
@balena-ci I self certify!

src/unstable-temp/DownloadImageModal/FormModel.tsx
```diff
@@ -208,9 +208,7 @@ const FormControl = ({ onModelChange, options, model }: FormControlProps) => {
 								name={options.name}
 								value={
 									!!model[options.name]
-										? moment(model[options.name] as string)
-												.utc()
-												.format()
+										? parseISO(model[options.name] as string).toISOString()
```
iamsolankiamit, line 211: would the date returned in `model[options.name]` will always be an ISO string conforming to ISO 8601 format (`YYYY-MM-DDTHH:mm:ss.sssZ`)? If it does then using `new Date()` or `Date.parse()` should also work.  We wouldn't even need `date-fns`

src/unstable-temp/DownloadImageModal/ImageForm.tsx
```diff
@@ -413,11 +413,9 @@ export const ImageForm = ({
 					onClick={(event: React.MouseEvent) => {
 						if (downloadOptions.provisioningKeyExpiryDate) {
 							// Convert to UTC for /download
-							downloadOptions.provisioningKeyExpiryDate = moment(
+							downloadOptions.provisioningKeyExpiryDate = parseISO(
```
iamsolankiamit, line 416: @myarmolinsky Sorry I didn't mention it here as well. would it also be a valid ISO string?

myarmolinsky: @iamsolankiamit It works but we need to convert it to UTC too

myarmolinsky: Nevermind it works, idk why it appeared to me that it did not at first ...

myarmolinsky: Good catches, thank you 🚀

PR #1660, opened 2022-10-19
Updated engine to 16, installed node-modules, ran tests

zoobot:
Test Suites: 22 passed, 22 total
Tests:       560 passed, 560 total
Snapshots:   308 passed, 308 total
Time:        19.182 s
Ran all test suites.

Change-type: patch


thgreasi:
Previously we have been marking such changes as major (eg in 1ea47cfca8cd6a1676bdeac90760a4153e3552e3), since otherwise `npm i` for users in old node versions would stop.
Since this already works with node16 and flowzone already was testing this against node 12-18, I would personally wait until we have any other major change to do and get the in together.
Let's also change the title to be in present tense like the rest of the commits, like: Drop support for node < 16

thgreasi:
We do that only for majors @zoobot b/c customers have raised concerns at some point about doing majors too often make the packages look like not mature enough or not well though beforehand.

PR #1664, opened 2022-11-01
FilterModal: Allow false and 0 as valid values

thgreasi:
Resolves: #1663
Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #1673, opened 2022-11-15
Add server side pagination

JSReds:
Connects-to: # 
See: <url> 
Depends-on: <url> 

---

##### Contributor checklist

##### Reviewer Guidelines

---

### DIFF
--- a/src/components/Table/TableBase.tsx
+++ b/src/components/Table/TableBase.tsx
@@ -24,6 +24,15 @@ const BaseTableWrapper = styled.div`
 	max-width: 100%;
 	border-bottom: 1px solid ${(props) => props.theme.colors.quartenary.main};
 `;
+export const DEFAULT_ITEMS_PER_PAGE = 50;
+export type Pagination = { itemsPerPage?: number } & (
+	| { serverSide?: undefined }
+	| {
+			serverSide: true;
+			currentPage: number;
+			totalItems: number;
+	  }
+);
 
 interface InternalTableBaseProps {
 	hasCheckbox: boolean;
@@ -170,7 +179,8 @@ export class TableBase<T extends {}> extends React.Component<
 	}
 
 	public componentDidUpdate(prevProps: TableBaseProps<T>) {
-		const { sort, checkedItems, data, itemsPerPage, rowKey } = this.props;
+		const { sort, checkedItems, data, pagination, rowKey } = this.props;
+		const serverSide = pagination?.serverSide;
 		if (sort && !isEqual(prevProps.sort, sort)) {
 			this.setState({
 				sort,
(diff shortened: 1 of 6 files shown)


src/components/Table/TableBase.tsx
```diff
@@ -186,10 +193,12 @@ export class TableBase<T extends {}> extends React.Component<
 			this.setRowSelection(checkedItems);
```
thgreasi, line 193: I think we also need to doublecheck how selection works.
* We need to update `toggleChecked`, since it does include a client side sorting and assumes it has all the data while trying to map the checkbox the user clicked and the item that the row was about.
* We probably also need to check `toggleAllChecked` as well, and maybe in server side more it should be returning `all`, to let the consumer know that they need to query and fetch AAAAAALLLLLLLLLL item IDs from the backend.

thgreasi (commented):
This is GREAT!!!

src/components/Table/TableBase.tsx
```diff
@@ -186,10 +193,12 @@ export class TableBase<T extends {}> extends React.Component<
...
 		}
 
-		const totalItems = data?.length ?? 0;
+		const totalItems = pagination?.totalItems ?? data?.length ?? 0;
+		const currentPage = pagination?.currentPage ?? this.state.page;
 		if (
-			this.state.page !== 0 &&
-			totalItems <= this.state.page * (itemsPerPage ?? 50)
+			currentPage !== 1 &&
```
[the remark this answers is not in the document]
JSReds: my bad sorry, I pushed something I was testing... Removing it

src/components/Table/TableBase.tsx
JSReds: Work in progress. I need to merge this PR first to be able to do something like this in the following PR:

isDisabled & actionFn:
   * isServerSide: false =>  T[]
   * isServerSide: true => 
   * selecting N items: { id: { $in: [] } }
   * select-all: { $and: [{ belongs_to__application: xyz }, searchFilters] }

src/components/Table/TableBase.tsx
```diff
@@ -557,36 +570,37 @@ export class TableBase<T extends {}> extends React.Component<
...
+			pagination,
 		} = this.props;
 
 		const { page, sort } = this.state;
+		const serverSide = pagination?.serverSide;
 		const items = data || [];
-		const totalItems = items.length;
-		const _itemsPerPage = itemsPerPage || 50;
+		const totalItems = serverSide ? pagination?.totalItems : items.length;
+		const itemsPerPage = pagination?.itemsPerPage || DEFAULT_ITEMS_PER_PAGE;
```
thgreasi, line 580: Given that this is a minor & `itemsPerPage` is still in the properties, we should
```suggestion
		const itemsPerPage = pagination?.itemsPerPage ?? this.props.itemsPerPage ?? DEFAULT_ITEMS_PER_PAGE;
```
both here and in `setPage`.
Otherwise we can do a major and remove it completely.
I think atm I'm in favor of the minor so that we can do a single major after the product call on how selection should work.

thgreasi (approved):

PR #1679, opened 2022-12-01
Improve handling of enum and oneOf and fix operators

JSReds:
Change-type: minor

---

##### Contributor checklist

##### Reviewer Guidelines

---

### DIFF
--- a/src/components/DataTypes/boolean.tsx
+++ b/src/components/DataTypes/boolean.tsx
@@ -4,9 +4,9 @@ export const operators = () => ({
 	is: 'is',
 });
 
-export type OperatorSlug = keyof ReturnType<typeof operators>;
+export type OperatorSlug = keyof ReturnType<typeof operators> | 'is_not';
 
-export const createFilter: CreateFilter<OperatorSlug | 'is_not'> = (
+export const createFilter: CreateFilter<OperatorSlug> = (
 	field,
 	operator,
 	value,
--- a/src/components/DataTypes/enum.tsx
+++ b/src/components/DataTypes/enum.tsx
@@ -1,3 +1,5 @@
+import { regexEscape } from '../../utils/index';
+import { FULL_TEXT_SLUG } from '../Filters/SchemaSieve';
 import type { CreateFilter } from './utils';
 
 export const operators = () => ({
@@ -5,7 +7,9 @@ export const operators = () => ({
 	is_not: 'is not',
 });
 
-export type OperatorSlug = keyof ReturnType<typeof operators>;
+export type OperatorSlug =
+	| keyof ReturnType<typeof operators>
+	| typeof FULL_TEXT_SLUG;
 
 export const createFilter: CreateFilter<OperatorSlug> = (
 	field,
@@ -14,7 +18,23 @@ export const createFilter: CreateFilter<OperatorSlug> = (
 ) => {
 	const operatorSlug = operator.slug;
 
-	if (operatorSlug === 'is') {
+	if (operatorSlug === FULL_TEXT_SLUG && typeof value === 'string') {
+		return {
+			properties: {
+				[field]: {
+					type: 'string',
+					// Note: An alternative could be to do: { enum: subSchema.enum.filter(x => x.toLowerCase().includes(value.toLowerCase())) }
+					regexp: {
+						pattern: regexEscape(value),
+						flags: 'i',
+					},
+				},
+			},
+			required: [field],
+		};
+	}
+
+	if (operatorSlug === 'is' || operatorSlug === FULL_TEXT_SLUG) {
 		return {
 			properties: {
 				[field]: {
--- a/src/components/DataTypes/oneOf.tsx
+++ b/src/components/DataTypes/oneOf.tsx
@@ -1,3 +1,6 @@
+import { RequiredProps } from '../../common-types';
+import { FULL_TEXT_SLUG } from '../Filters/SchemaSieve';
+import { JSONSchema } from '../Renderer/types';
 import type { CreateFilter } from './utils';
 
 export const operators = () => ({
@@ -5,16 +8,44 @@ export const operators = () => ({
 	is_not: 'is not',
 });
 
-export type OperatorSlug = keyof ReturnType<typeof operators>;
+export type OperatorSlug =
+	| keyof ReturnType<typeof operators>
+	| typeof FULL_TEXT_SLUG;
 
 export const createFilter: CreateFilter<OperatorSlug> = (
 	field,
 	operator,
 	value,
+	subSchema,
 ) => {
 	const operatorSlug = operator.slug;
+	if (
+		operatorSlug === FULL_TEXT_SLUG &&
+		typeof value === 'string' &&
+		subSchema.oneOf?.every(
+			(o): o is RequiredProps<JSONSchema, 'title' | 'const'> =>
+				typeof o === 'object' &&
+				typeof o.title === 'string' &&
+				typeof o.const === 'string',
+		)
+	) {
+		const transformedEnum = subSchema.oneOf
+			.filter((o) => o.title.toLowerCase().includes(value.toLowerCase()))
+			.map((o) => o.const);
 
-	if (operatorSlug === 'is') {
+		if (transformedEnum.length) {
+			return {
+				properties: {
+					[field]: {
+						enum: transformedEnum,
+					},
+				},
+				required: [field],
+			};
+		}
+	}
+
+	if (operatorSlug === 'is' || operatorSlug === FULL_TEXT_SLUG) {
 		return {
 			properties: {
 				[field]: {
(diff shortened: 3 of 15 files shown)


src/components/DataTypes/oneOf.tsx
```diff
@@ -1,20 +1,49 @@
...
+		return {
+			properties: {
+				[field]:
+					typeof value === 'string'
+						? {
+								enum: subSchema
+									.oneOf!.filter((o) =>
+										o.title!.toLowerCase().includes(value.toLowerCase()),
+									)
+									.map((o) => o.const!),
```
thgreasi, line 39: Add `type RequiredProps<T, K extends keyof T> = T & Required<Pick<T, K>>;` to the common types, and then we should be able to remove all the `!`:
```suggestion
			(o): o is RequiredProps<JSONSchema, 'title' | 'const'> =>
				typeof o === 'object' &&
				typeof o.title === 'string' &&
				typeof o.const === 'string',
		)
	) {
		return {
			properties: {
				[field]:
					typeof value === 'string'
						? {
								enum: subSchema
									.oneOf.filter((o) =>
										o.title.toLowerCase().includes(value.toLowerCase()),
									)
									.map((o) => o.const),
```

src/components/DataTypes/oneOf.tsx
```diff
@@ -1,20 +1,49 @@
...
+						  }
```
thgreasi, line 40: Do we also need to add `required: [field],` here?

src/components/DataTypes/oneOf.tsx
```diff
@@ -1,20 +1,49 @@
...
 	field,
 	operator,
 	value,
+	subSchema,
 ) => {
 	const operatorSlug = operator.slug;
 
-	if (operatorSlug === 'is') {
+	if (
+		operatorSlug === FULL_TEXT_SLUG &&
```
thgreasi, line 22: Maybe this should move the `typeof value === 'string'` check as part of the `if` statement :thinking: ?
B/c otherwise the return is the same as for the `is` operator right below.

src/components/DataTypes/string.tsx
```diff
@@ -46,7 +46,7 @@ export const createFilter: CreateFilter<OperatorSlug> = (
 		};
 	}
 
-	if (operatorSlug === 'contains') {
+	if (operatorSlug === 'contains' || operatorSlug === 'full_text_search') {
```
thgreasi, line 49:
```suggestion
	if (operatorSlug === 'contains' || operatorSlug === FULL_TEXT_SLUG) {
```

src/components/Filters/FilterModal.tsx
```diff
@@ -197,8 +197,11 @@ const FilterInputs = ({
 	const [searchTerm, setSearchTerm] = React.useState('');
 	const debouncedSetSearchTerm = debounce(setSearchTerm, 300);
 	const operators = React.useMemo(
-		() => (signature.field ? getOperators(schema, signature.field) : []),
-		[signature.field, schema],
+		() =>
+			signature.field
+				? getOperators(schema, signature.field, signature.refScheme)
```
thgreasi, line 202: Does this also work or `x-foreign-key-scheme` based properties?

src/components/Filters/SchemaSieve.ts
```diff
@@ -57,11 +55,11 @@ export type EditSchema = (
 interface FullTextFilterItem {
 	key: string;
 	schema: JSONSchema;
-	operator: Operator<string>;
+	// operator: Operator<string>;
 	model: {
 		createFilter: CreateFilter;
 	};
-	searchKey?: string;
+	// searchKey?: string;
```
thgreasi, line 62: comment :police_car:

src/components/Filters/SchemaSieve.ts
```diff
@@ -124,44 +122,18 @@ export const createFullTextSearchFilter = (
...
-						});
-						itemsAccumulator.push({
-							...defaultItem,
-							operator: {
-								label: operators['value_contains' as keyof typeof operators],
-								slug: 'value_contains',
-							},
-							searchKey: findKey(item.properties!, { description: 'value' })!,
-						});
-					}
```
thgreasi, line 164: Did all this disappear b/c it's now handled by each case separately, rather than in a central way?

src/components/Filters/SchemaSieve.ts
```diff
@@ -260,29 +235,15 @@ export const getModelFilter = (
...
 ): JSONSchema => {
-	let newOperator = operator;
-	let subSchema = schema;
-	if (isFullTextSearch) {
-		subSchema = { ...schema };
-		delete subSchema.oneOf;
-		delete subSchema.enum;
-	}
+	const newOperator = operator;
+	const subSchema = schema;
```
thgreasi, line 240:
```suggestion
```
Atm we are just renaming these vars w/o reason. Let's just use `operator` directly and rename the `schema` argument to `subSchema`.

src/components/DataTypes/boolean.tsx
```diff
@@ -1,12 +1,13 @@
...
 
 export const operators = () => ({
 	is: 'is',
 });
 
-export type OperatorSlug = keyof ReturnType<typeof operators>;
+export type OperatorSlugs = OperatorKeys<typeof operators>;
 
-export const createFilter: CreateFilter<OperatorSlug | 'is_not'> = (
+export const createFilter: CreateFilter<OperatorSlugs | 'is_not'> = (
```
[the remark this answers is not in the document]
JSReds: We don't have `is_not`since we want to have a single operator in the operator's dropdown in the filter modal when we filter by a boolean. I'm not sure if I got what you mean but `full text search` cannot be in the operators otherwise we would see it in the dropdown choices. Did I get your point or am I in the wrong direction ?

src/components/DataTypes/enum.tsx
```diff
@@ -1,19 +1,40 @@
...
+									flags: 'i',
+								},
+						  }
+						: { const: value },
+			},
+			required: [field],
+		};
+	}
+
 	if (operatorSlug === 'is') {
```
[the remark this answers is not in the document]
JSReds: yeah make sense, even if I'm not sure if `operatorSlug === FULL_TEXT_SLUG && typeof value !== 'string'` really exist as a case

src/components/DataTypes/oneOf.tsx
JSReds: `RequiredProps` is uresolved. I'm adding it in the common-types.ts but I'll probably add a common.d.ts file in rendition in a secondary PR

src/components/DataTypes/oneOf.tsx
JSReds: yep totally right

src/components/Filters/SchemaSieve.ts
JSReds: Yes sir, I find it much more readable. I would keep it like this if you agree

src/components/Filters/SchemaSieve.ts
JSReds: Definitely a holdover from a past change. Thank you :)

src/components/Filters/FilterModal.tsx
JSReds: yes, because at this moment of the code we created an item for each schema.properties and each x-ref-scheme || x-foreign-key-scheme

src/components/DataTypes/boolean.tsx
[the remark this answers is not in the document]
thgreasi: I probably confused you. I agree on implementing only the operators that makes sense for each data type (ie I don't think contains for Boolean makes 100% sense. I meant:
```
export type OperatorSlug = keyof ReturnType<typeof operators> | 'is_not';

export const createFilter: CreateFilter<OperatorSlug> = (
```
Essentially having OperatorSlug define all the operators that the CreateFilter supports, even if the `operators` map defines just a subset (like in this case, which we do not want to surface an `is_not` operator in the dropdown (actually let's add a comment about this so it's clear to everyone why `operators` only has `is`, assuming that's indeed the reason)).

src/components/DataTypes/oneOf.tsx
thgreasi: We don't have a `common.d.t` in rendition right?
This should be in `src/common-types.ts` just like how `Overwrite` is.
How we do `common.d.ts` in leaf projects, where tsconfig is set to peak global types from there, does not work for libraries. Consumers will probably get errors while trying to compile a library that uses helpers types defined in a `common.d.ts`, unless they "skipLibCheck" or happen to have the same helper types in their own `common.d.ts`.
So, let's keep it in `src/common-types.ts`.

src/components/DataTypes/boolean.tsx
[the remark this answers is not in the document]
JSReds: @thgreasi unfortunately I have seen that there is an underlying problem. Having removed the operator converter from `array.tsx` (change which I honestly love :) ) I find myself having to handle every case in natural types. The case I have in mind is the following:

```
Data = {... numberArray: [1,2,3,4]}

Schema = {
...
  "numberArray": {
     "type": "arrays",
     "items": {
       "type": "number"
     }
  }
}
```

In this case there is no `refScheme`, so we would have array operators (contains, does not contain) to select one of the numbers inside the array, and this would cause an empty filter if we don't handle the contains case in the `number.tsx`

thgreasi (approved):

PR #1688, opened 2023-01-05
Adapt onCheck event for server side pagination

JSReds:
Change-type: minor

Screenshot 2023-01-05 at 16 44 10
Screenshot 2023-01-05 at 16 44 20

---

##### Contributor checklist

##### Reviewer Guidelines

---

### DIFF
--- a/src/components/Table/TableBase.tsx
+++ b/src/components/Table/TableBase.tsx
@@ -139,11 +139,11 @@ const Base = styled.div<InternalTableBaseProps>`
 	}
 `;
 
-type CheckedTypes = 'none' | 'some' | 'all';
+export type CheckedState = 'none' | 'some' | 'all';
 
 interface TableBaseState<T> {
-	allChecked: CheckedTypes;
-	checkedItems: T[];
+	checkedState: CheckedState;
+	checkedItems: T[] | undefined;
 	sort: {
 		reverse: boolean;
 		field: null | keyof T;
@@ -283,13 +283,16 @@ export class TableBase<T extends {}> extends React.Component<
 		this.$getValidatedCheckedItems.clear();
 	}
 
-	public howManyRowsChecked(checkedItems: T[]): CheckedTypes {
-		const { rowKey, data } = this.props;
+	public howManyRowsChecked(checkedItems: T[]): CheckedState {
+		const { rowKey, data, pagination } = this.props;
+
 		if (!rowKey || !data) {
 			return 'none';
 		}
 
-		return data.length === checkedItems.length
+		return (pagination?.serverSide &&
+			checkedItems.length === pagination.totalItems) ||
+			(!pagination?.serverSide && data.length === checkedItems.length)
 			? 'all'
 			: checkedItems.length > 0
 			? 'some'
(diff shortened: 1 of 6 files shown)


src/components/Table/TableBase.tsx
```diff
@@ -377,12 +380,17 @@ export class TableBase<T extends {}> extends React.Component<
...
 		this.setState(newState);
 		if (this.props.onCheck && hasCheckedItemsChanged) {
-			this.props.onCheck(newState.checkedItems);
+			this.props.onCheck(newState.checkedItems, this.state.allChecked);
 		}
 	};
 
+	private toggleCheckedState = () => {
+		const { allChecked } = this.state;
+		return allChecked === 'none' ? 'all' : 'none';
```
thgreasi, line 389: Would it make sense if this was v ?
```suggestion
		return allChecked === 'all' ? 'none' : 'all';
```
This way:
none -> all
some -> all
all -> none

src/components/Table/TableBase.tsx
```diff
@@ -393,12 +401,14 @@ export class TableBase<T extends {}> extends React.Component<
 					: data.slice();
 		}
 		if (onCheck) {
-			onCheck(checkedItems);
+			onCheck(checkedItems, this.state.allChecked);
```
thgreasi, line 404: I think that if we are in server side pagination and `this.toggleCheckedState()` returns `all`, then we shouldn't given checkedItems (in which case we also can skip computing them).

thgreasi: Let me rephrase this, I think that for server side pagination, giving `checkedItems` & `all` at the same time should only happen when `checkedItems.length === pagination.totalItems`, since that's the only case that `checkedItems` will hold really `all` items.

src/components/Table/TableBase.tsx
JSReds: that's not what we have today, should we change it to this version ? @thgreasi

thgreasi: I hadn't realized this. Ignore this, let's preserve the current behavior.

src/components/Table/TableBase.tsx
JSReds: yep good point! If you agree I would just pass it as undefined wrapping the computation in a if statement

thgreasi (approved):

src/components/Table/TableBase.tsx
```diff
@@ -139,11 +139,11 @@ const Base = styled.div<InternalTableBaseProps>`
 	}
 `;
 
-type CheckedTypes = 'none' | 'some' | 'all';
+export type CheckedTypes = 'none' | 'some' | 'all';
```
thgreasi, line 142: How about CheckedState or something?

src/components/Table/TableBase.tsx
```diff
@@ -592,6 +614,10 @@ export class TableBase<T extends {}> extends React.Component<
 		const checkedRowIdentifiers = this.getCheckedRowIdentifiers();
 		const highlightedRowIdentifiers = this.getHighlightedRowIdentifiers();
 		const disabledRowIdentifiers = this.getDisabledRowIdentifiers();
+		const selectedItems =
```
thgreasi, line 617:
```suggestion
		const selectedItemCount =
```

PR #1698, opened 2023-03-17
Table: Restrict the tagField to be a string key of the entity type

thgreasi:
Prepares us for TS 5.0 which deprecates
keyofStringsOnly.

Change-type: patch


thgreasi:
I self-certify!

